PR #8, opened 2016-12-08
Class indexing

dantleech:


lib/Console/Command/IndexCommand.php
```diff
@@ -0,0 +1,81 @@
...
+        $progress = new ProgressBar($output, count($this->finder));
+        $progress->start();
+
+        $index = $this->indexer->__invoke($index, $this->finder, function ($file) use ($progress, $output) {
+            $progress->advance();
+        });
+
+        $progress->finish();
+
+        file_put_contents($indexPath, "<?php return <<<'EOT'" . PHP_EOL . serialize($index) . PHP_EOL . 'EOT' . PHP_EOL . ';');
```
dantleech: Should be handled by index class

lib/Index/Index.php
```diff
@@ -0,0 +1,66 @@
...
+    public function getPaths() 
+    {
+        return $this->paths;
+    }
+
+    public function getTimestamp() 
+    {
+        return $this->timestamp;
+    }
+}
```
dantleech: trestths

PR #11, opened 2017-01-07
Add missing letters

greg0ire:


dantleech:
Cheers, it might be worth saying that this library is waiting for some PRs in BetterReflection before it can work effectively!

greg0ire:
Noted… I tried it very briefly yesterday evening, and it worked quite well. What about the test suite? Why is it broken? Blocked by PRs in BetterReflection too? I had a quick look, it seems that testing with "invalid" code (by that I mean the character you used to represent the cursor) was possible before and no longer is.
Also, I noticed the minimum-stability for Composer packages was set to "dev"… is that (still) necessary?

dantleech:
Yeah, we need to disable error throwing in the `php-parser` lib for tests to pass: https://github.com/Roave/BetterReflection/pull/232. Also performance is a real issue, but could be improved dramatically: https://github.com/Roave/BetterReflection/pull/239

PR #21, opened 2017-06-23
fix hashbang

bendavies:
I have php 7.1 installed, and calling `./bin/phpactor` resulted in
```
Parse error: parse error, expecting `';'' or `'{'' in /....phpactor/vendor/dtl/class-to-file/lib/CompositeTransformer.php on line 16
```
which confused me for a while!

Fix this by using `#!/usr/bin/env php` that will use whatever php executable appears first in the user's `$PATH`

### DIFF
--- a/bin/phpactor
+++ b/bin/phpactor
@@ -1,3 +1,3 @@
-#!/usr/bin/php
+#!/usr/bin/env php
 <?php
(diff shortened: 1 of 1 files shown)


dantleech:
Ha, I was scratching my head over why the travis build was failing :)

PR #35, opened 2017-07-02
Use worse reflection

dantleech:
This PR uses the updated type-inferrer, which in turn uses `worse reflection` to infer method and property types.

### DIFF
--- a/lib/Application/ClassSearch/ClassSearch.php
+++ b/lib/Application/ClassSearch/ClassSearch.php
@@ -8,7 +8,14 @@ use DTL\ClassFileConverter\Domain\FilePath;
 
 final class ClassSearch
 {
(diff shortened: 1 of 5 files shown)


lib/Application/ClassSearch/ClassSearch.php
```diff
@@ -33,7 +40,7 @@ public function classSearch(string $name)
             $candidates = $this->fileToClass->fileToClassCandidates(FilePath::fromString((string) $file->path()));
 
             if (false === $candidates->noneFound()) {
-                $result['class_name'] = (string) $candidates->best()->name();
+                $result['class_name'] = $candidates->best();
```
dantleech: nein

PR #112, opened 2017-08-08
README: fix transform anchor

TomasVotruba:


dantleech:
:+1: thanks

TomasVotruba:
You're welcome 👍

TomasVotruba:
Great tool btw 👏 

 I'm working on similar one based in nikic/php-parser

dantleech:
nice, I preferred the tolerant parser in the end as you can traverse to parent nodes which made things much simpler :)

TomasVotruba:
@dantleech That is definitelly a useful feature! Could you link me to that code and method on Github?

TomasVotruba:
I found similar section in nikic/php-parser: https://github.com/nikic/PHP-Parser/blob/master/doc/5_FAQ.markdown#how-can-the-parent-of-a-node-be-obtained

I'd like to inspire, where do you use parent nodes in your code?

dantleech:
Tolerant Parser is [here](https://github.com/Microsoft/tolerant-php-parser), example of getting an ancestor [here](https://github.com/phpactor/worse-reflection/blob/master/lib/Reflection/Inference/FrameBuilder.php#L53).

Also hads lots of helper functions such as `$node->getResolvedNamespace()` to return the fully qualified name of, f.e., a class.

TomasVotruba:
@dantleech Thanks. Link goes to empty line. Where exactly that should go?

> Also hads lots of helper functions such as $node->getResolvedNamespace() to return the fully qualified name of, f.e., a class.

Nice. I'm have a prejudice to Microsoft products, because in my experience it fails in time.
On the other hand, I have good experience with nikic in many other projects.

dantleech:
4 lines above it. I am also skeptical about that company, but the Tolerant parser is a better fit for this project at least.

PR #153, opened 2017-08-31
Parse cwd path until find .git repository

cmizzi:
I'm currently using this plugin with `vim` but my configuration handles a `autochdir`. So, when I opened a file, the `cwd` changed. This feature implements a simple way to parse the sub-directory and find the related `.git` directory (base root). It allows to use the base `.git` directory and load easily the `vendor/autoload.php` file

### DIFF
--- a/lib/Container/CoreExtension.php
+++ b/lib/Container/CoreExtension.php
@@ -80,7 +80,7 @@ class CoreExtension implements ExtensionInterface
         }
 
         // Return base CWD where .git directory is present
-        if (!is_dir($path . '/' . '.git') and $path !== '/') {
+        if (!is_dir(sprintf('%s/.git', $path)) && $path !== '/') {
             return $this->getBaseCwd(dirname($path));
         }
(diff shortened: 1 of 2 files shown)


lib/Container/CoreExtension.php
```diff
@@ -74,6 +74,19 @@ public function load(Container $container)
         $this->registerApplicationServices($container);
     }
 
+    private function getBaseCwd($path = null) {
+        if (is_null($path)) {
+            $path = getcwd();
+        }
+
+        // Return base CWD where .git directory is present
+        if (!is_dir($path . '/' . '.git') and $path !== '/') {
```
dantleech: Could use `sprintf('%s/.git')` and use `&&` instead of `and`

cmizzi: Done !

dantleech:
Thanks :) 

It would be nice to have a test for this, I created #154

PR #159, opened 2017-09-01
.git in submodules are not dirs but normal files

docteurklein:
This allows to detect the cwd even when working in submodules


dantleech:
:+1: thanks :)

PR #160, opened 2017-09-01
Adds vim and neovim paths to default `code_transform.template_paths`

cmizzi:
By default, only the current project is checked. Now, it there's no `.phpactor/phpactor/template` directory, fallback on `$HOME/.vim/phpactor/template` and `$HOME/.nvim/phpactor/template` directories to load custom transformation templates

### DIFF
--- a/lib/Config/ConfigLoader.php
+++ b/lib/Config/ConfigLoader.php
@@ -41,7 +41,9 @@ class ConfigLoader
     public function configDirs(): array
     {
         $configDirs = $this->xdg->getConfigDirs();
-        array_unshift($configDirs, Path::join(getcwd(), '/.phpactor'));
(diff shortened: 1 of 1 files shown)


dantleech:
You can add these paths to the global configuration: https://github.com/phpactor/phpactor#configuration

I don't think we should be adding references to VIM in the PHP code.

cmizzi:
Yes, you're not really wrong. I thing the documentation about overriding must be more explained :)

dantleech:
Yes indeed, though the functionality is still quite new and may change - but then again, it might be years before it does change.

I think restructuring the README file to have a table-of-contents and then giving a better explanation of the configuration system might help? Created https://github.com/phpactor/phpactor/issues/166

Closing this now as I guess its a non-issue, feel free to re-open otherwise.

PR #161, opened 2017-09-01
Prevents `cwd` to be took from vendor directory

cmizzi:
If we go to definition, the file has potentiel chance to be placed into the `vendor` directory. But, Composer doesn't install dependencies into the dependency project folder. So, we can't next go to definition within the dependency project. This patch prevents for getting `cwd` within the `vendor` directory, always parent one.

```
./project
./project/vendor
./project/vendor/toto/Alias.php
./project/vendor/phpactor/Definition.php -> refers to ./project/vendor/toto/Alias.php.

Cannot be found because the cwd is in the project/vendor/phpactor directory. That's
wrong because there's no vendor directory within
```

### DIFF
--- a/lib/Container/CoreExtension.php
+++ b/lib/Container/CoreExtension.php
@@ -80,7 +80,7 @@ class CoreExtension implements ExtensionInterface
         }
 
         // Return base CWD where .git directory is present
-        if (!file_exists(sprintf('%s/.git', $path)) && $path !== '/') {
+        if (!file_exists(sprintf('%s/.git', $path)) && $path !== '/' || strpos($path, 'vendor')) {
             return $this->getBaseCwd(dirname($path));
         }
 


lib/Container/CoreExtension.php
```diff
@@ -80,7 +80,7 @@ private function getBaseCwd($path = null) {
         }
 
         // Return base CWD where .git directory is present
-        if (!file_exists(sprintf('%s/.git', $path)) && $path !== '/') {
+        if (!file_exists(sprintf('%s/.git', $path)) && $path !== '/' || strpos($path, 'vendor')) {
```
dantleech: vendor is configurable in `composer.yml`, it is not guaranteed to be `vendor`.

cmizzi: Yes, but you supposed that is `vendor` because you hardcoded the `autoload` attributes (https://github.com/phpactor/phpactor/blob/master/lib/Container/CoreExtension.php#L59)

dantleech: Thats configurable :)

cmizzi: Is there a usercase that overrides the `autoload` without overriding `cwd` ?

cmizzi: In this part of configuration, we supposed the user using default configuration. If the `composer.json` file is overridden, that means the user must also define an override file, no ?

dantleech: Yeah, you can actually configure mutliple autoloaders (e.g. some applications have their own package management).

Also, some people change the name of `vendor` in `composer.json`

cmizzi: Are you checking this entry somewhere ? Or the user must override the default configuration to performs `autoloader` and `cwd` ?

dantleech: > Are you checking this entry somewhere ?

Not sure what you mean. If the `vendor/autoload.php` (of whatever is configured) is not found, an exception will be thrown.

cmizzi: So, for a default configuration, what did you mean about vendor directory could change ?

dantleech:
```

dantleech: See https://getcomposer.org/doc/06-config.md#vendor-dir.

I think potentially a better solution to this issue is to always pass the `cwd` as an option to Phpactor from the plugin, so that it is _always_ the path in which VIM was loaded.

Would that make sense?

cmizzi:
It will not work with autochdir option from Vim. Traversing directories is the good solution but we have to take care about composer configuration, IMHO

dantleech:
> It will not work with autochdir option from Vim

I don't see why not? When VIM is booted you are in the project root? or? It is at this point that we will store the CWD, so it will remain constant for the lifetime of VIM regardless of any chdirs that happen.

dantleech:
Closing in favor of #194

PR #172, opened 2017-09-03
Context action

dantleech:
This PR introduces a framework for contextual actions.


dantleech:
It is probably better to adopt the [RPC](https://github.com/phpactor/phpactor/wiki/WIP-Phpactor-RPC-Protocol) approach here, and reuse some of this (especially the Goto definitions).

Issue #177, opened 2017-09-08
Object of class RecursiveIteratorIterator could not be converted to string


panders23:
Hi,

i'm currently facing the following problem with HEAD:

```
  [Symfony\Component\Debug\Exception\ContextErrorException]                                          
  Catchable Fatal Error: Object of class RecursiveIteratorIterator could not be converted to string
```

command to reproduce the issue:

```
~/phpactor/bin/phpactor class:search HttpClient
```

Same error happens in vim with `:call phpactor#ClassExpand` 

I'm using the setup as described here: https://github.com/phpactor/phpactor/blob/master/plugin/README.md

dantleech:
1. Is this a regression - was it previously working?
2. Can you should the stack trace? (add `--verbose` to the command).
3. Are you able to reproduce this on an open-source code base? (e.g. Symfony Standard Edfition).

dantleech:
Seems to be reproduced now by Travis with some version of 7.1: https://travis-ci.org/phpactor/phpactor/jobs/273672984

dantleech:
Possibly related to this in `7.1.9`? https://bugs.php.net/bug.php?id=74977

dantleech:
PHP Bug: https://bugs.php.net/bug.php?id=75155&edit=2

Implemented temproary workaround be re-implementing `ArrayIterator`: https://github.com/phpactor/source-code-filesystem/blob/master/lib/Iterator/AppendIterator.php

panders23:
You're right, i'm using php 7.1.9 on arch. Your workaround fixed my issue, thank you.

PR #183, opened 2017-09-09
binary cleanup

Taluu:
In order to be able to install phpactor as a dependency through composer (through adding the correct repositories in the composer json user file), the `autoload.php` file can be not located in a `vendor` directory at the root of the phpactor project.

It can be deeper, so I ported the code from [PHPUnit](https://github.com/sebastianbergmann/phpunit/blob/master/phpunit#L30-L36) to check some other files.

I also removed the version check for php 7.0, as the version constraint in composer should be enough.


bin/phpactor.php
```diff
@@ -1,38 +1,37 @@
...
 
 $application = new Application();
 $output = new ConsoleOutput();
 
-try { 
+try {
     $application->initialize();
     $application->run(null, $output);
-} catch (\Exception $e) {
+} catch (Exception $e) {
```
Taluu: `\` not needed here as we're on a namespace-less file, so `Exception` is already mapping to `\Exception`. Can revert if needed.

bin/phpactor.php
```diff
@@ -1,38 +1,37 @@
 <?php
+use Phpactor\Console\Application;
+
+use Symfony\Component\Debug\Debug;
 
 use Symfony\Component\Console\Input\ArgvInput;
+use Symfony\Component\Console\Output\ConsoleOutput;
```
Taluu: Moved the use declaration [because PSR](http://www.php-fig.org/psr/psr-2/#namespace-and-use-declarations), unless you are not sticking to them (I can revert)

Taluu:
Something else that is kind of bothering me ; why the require on the `phpactor` file to `phpactor.php` ? If it's for the coloration, the shebang in the `phpactor` file is enough... I can merge the two files then if you want.

bin/phpactor.php
```diff
@@ -1,38 +1,37 @@
...
-$autoloadFile = __DIR__ . '/../vendor/autoload.php';
-if (!file_exists($autoloadFile)) {
+foreach ([__DIR__ . '/../../autoload.php', __DIR__ . '/../vendor/autoload.php', __DIR__ . '/vendor/autoload.php'] as $file) {
+    if (file_exists($file)) {
+        define('PHPACTOR_COMPOSER_INSTALL', $file);
+        break;
+    }
+}
+
+if (!defined('PHPACTOR_COMPOSER_INSTALL')) {
```
dantleech: This would be fine as a variable no? Why a constant?

bin/phpactor.php
```diff
@@ -1,38 +1,37 @@
...
+use Symfony\Component\Debug\Debug;
 
 use Symfony\Component\Console\Input\ArgvInput;
+use Symfony\Component\Console\Output\ConsoleOutput;
 
-$autoloadFile = __DIR__ . '/../vendor/autoload.php';
-if (!file_exists($autoloadFile)) {
+foreach ([__DIR__ . '/../../autoload.php', __DIR__ . '/../vendor/autoload.php', __DIR__ . '/vendor/autoload.php'] as $file) {
+    if (file_exists($file)) {
+        define('PHPACTOR_COMPOSER_INSTALL', $file);
```
dantleech: This can be defined as a variable no? Why a constant?

bin/phpactor.php
```diff
@@ -1,38 +1,37 @@
 <?php
+use Phpactor\Console\Application;
+
+use Symfony\Component\Debug\Debug;
 
 use Symfony\Component\Console\Input\ArgvInput;
```
dantleech: No empty spaces between use statements and one blank line after `<?php` (I know Phpactor does this sometimes, that's a bug!)

bin/phpactor.php
```diff
@@ -1,38 +1,37 @@
...
+use Symfony\Component\Console\Output\ConsoleOutput;
 
-$autoloadFile = __DIR__ . '/../vendor/autoload.php';
-if (!file_exists($autoloadFile)) {
+foreach ([__DIR__ . '/../../autoload.php', __DIR__ . '/../vendor/autoload.php', __DIR__ . '/vendor/autoload.php'] as $file) {
```
dantleech: Please put each file path on a new line:

```php
foreach ([
    __DIR__ . '/../foo ...',
    __DIR__ . '/../bar ...',
    // ..
]) {
    // ...
}
```

dantleech:
> Something else that is kind of bothering me

There is no reason, other than the weak one for syntax highlighting by default. Feel free to move `phpactor.php` to `phpactor`.

dantleech:
Test failure looks like the issue reported in https://github.com/phpactor/phpactor/issues/177 and unrelated to this PR.

bin/phpactor.php
Taluu: No that's on me, I do have tendencies to split uses by groups. Reverting.

greg0ire:
Using `-vvv` here might help: https://github.com/phpactor/phpactor/blob/9616ae1acd9321e5a4dc73fd3ea2e074767ec538/tests/Integration/SystemTestCase.php#L89

dantleech:
> Using -vvv here might help: 

Yes and no -- verbose mode also logs a whole load of debug information to the console and will make tests fail...

bin/phpactor.php
Taluu: As I ported the autoload solver from phpunit, and it was using a constant, I thought it would be OK to use one here. But I redefined it as a variable which is ok too.

greg0ire:
You mean you assert the output is empty? Or is it output parsing the output of the commands? Because if it is, debug information like this should, and might already got to the badly named stderr, and will probably be ignored by the thing that parses it, right?

dantleech:
> You mean you assert the output is empty?

Piping to `stderr` would, I think, be better... and very easy indeed: https://github.com/phpactor/phpactor/blob/master/lib/Console/Application.php#L43

greg0ire:
Let's do this! #185

greg0ire:
> bin/phpactor 100755 → 100644 

You need to `chmod +x bin/phpactor`

Taluu:
Yep just saw. Changed.

Taluu:
Rebased and now it's green o/ @dantleech

bin/phpactor.php
```diff
@@ -1,38 +0,0 @@
...
-}
-
-require_once($autoloadFile);
-
-$minVersion = '7.0.0';
-
-if (version_compare(PHP_VERSION, $minVersion) < 0) {
-    echo 'Phpactor requires at least PHP ' . $minVersion;
-    exit(255);
-}
```
dantleech: I think this version check is missing from the new version?

Taluu: > I also removed the version check for php 7.0, as the version constraint in composer should be enough.

But then I can put it back, but IMO it is not really worth it

dantleech: Yes please. Composer won't stop Phpactor being executed with PHP < 7 (e.g. when you change your default version of PHP, or when executed from a PHAR).

Taluu: Done

bin/phpactor
```diff
@@ -1,3 +1,46 @@
...
+        'Phpactor dependencies not installed. Run `composer install` (https://getcomposer.org) in "%s"' . PHP_EOL,
+        realpath(__DIR__ . '/..')
+    );
+    exit(255);
+}
+
+$minVersion = '7.0.0';
+
+if (version_compare(PHP_VERSION, $minVersion) < 0) {
+    echo "Phpactor requires at least PHP {$minVersion}";
```
dantleech: `sprintf` would be more consistent here.

dantleech:
Thanks, I will fix the CS in master (I never use curly-encapsulation in strings)

dantleech:
See https://github.com/phpactor/phpactor/commit/bf02b703c74ec8de59eaeda81105f900fd9a76c8

Taluu:
I wasn't until PHP 7.0, string encapsulation got a lot better (even better than sprint and concat) : https://blog.blackfire.io/php-7-performance-improvements-encapsed-strings-optimization.html

dantleech:
Good to know, I prefer to use `sprintf` (or concat for simple cases) for now as it's consistent (won't comment on readability, as I am biased).

PR #185, opened 2017-09-09
Use stderr for logs

greg0ire:
Logs are not supposed to be taken into account by commands we might pipe to.


dantleech:
:+1: have added `--verbose` to the test case.

PR #194, opened 2017-09-11
Pass initial directory to Phpactor on every request.

### LINKED ISSUE #190: [VIM] CWD to an option passed to passed to Phpactor.
The "CWD" should be that in which VIM was invoked initially. It should be passed to Phpactor on each request.

dantleech:
Instead of traversing backwards until we find something that _resembles_ a project directory, we assume that the directory in which VIM was initially invoked is the project root.

For every request to Phpactor, the plugin will also send the initial working directory, this means you can `chdir` as much as you want, but you must _start_ in the project root.

@cmizzi @docteurklein what do you think?

Fixes #190


plugin/phpactor.vim
```diff
@@ -464,7 +466,8 @@ endfunction
 " !DEPRECATED! Will be removed when everything is ported to RPC
 ""
 function! phpactor#ExecStdIn(cmd, stdin)
-    let cmd = 'php ' . s:phpactorbinpath . ' ' . a:cmd
+    call confirm(s:phpactorInitialCwd)
```
dantleech: Note that this and the above method will shortly be removed completely.

tests/Integration/Console/ApplicationTest.php
```diff
@@ -28,16 +29,23 @@ public function testConfig()
...
 
         chdir($this->workspaceDir());
+        $output = new BufferedOutput();
         $application = new Application();
-        $application->initialize();
+        $application->setAutoExit(false);
+        $application->setCatchExceptions(false);
+        $application->run(new ArrayInput([
+            'command' => 'class:reflect',
+            'name' => 'asd',
```
dantleech: This test case could be refactored.

docteurklein:
great idea! looks good to me.

cmizzi:
Hum. I'm using Vi with autochdir. The project root will not be the same between each opened file :/

I don't think that a good idea because we have to start Vi within the project root. When you have installed fzf, you can easily edit files with Vi without changing the current cwd...

dantleech:
@cmizzi not sure I understand your issue. The only prerequisite for this to work is that you initially open VIM in the project root. As long as you don't exit VIM and start it again somewhere else, it will be fine.

You or VIM can change the CWD as many times as you like, as long as you don't close VIM Phpactor will still use the working directory in which you started VIM.

cmizzi:
So, to work on many projects at the same time, I've got to have many opened instances... it's not perfect IMHO. The basic idea to find the project root is good, but must be optimised, no need to know the current cwd, it could be easily guessed

dantleech:
> So, to work on many projects at the same time, I've got to have many opened instances

Unrelated, but I think thats a benefit of using VIM and Tmux, I often have > 20 instances of VIM open at any one time.

> it could be easily guessed

I think that's the problem, it's not so easy and there were edge cases (like it traversing to the root of the FS.

However, you can easily modify the global `s:phpactorInitialCwd` in your VIM config (you could implement the directory traversal thing in VIM).

PR #198, opened 2017-09-14
Refactor references handler

dantleech:
Support methods and accept offset and source code.


lib/Rpc/Handler/ReferencesHandler.php
```diff
@@ -0,0 +1,127 @@
...
+
+    private function classReferences(SymbolInformation $symbolInformation)
+    {
+        $classType = (string) $symbolInformation->type();
+
+        $references = $this->classReferences->findReferences($this->defaultFilesystem, $classType);
+        return $references['references'];
+    }
+
+    private function methodReferences(SymbolInformation $symbolInformation)
```
dantleech: the "member references" class can be easily refactored to handle all class members.

tests/Unit/Rpc/Handler/ReferencesHandlerTest.php
```diff
@@ -0,0 +1,205 @@
...
+
+        $second = array_shift($actions);
+        $this->assertEquals([
+            'file_references' => [
+                [
+                    'file' => 'barfoo',
+                    'references' => [
+                        [
+                            'start' => 10,
+                            'end' => 20,
```
dantleech: Need to refactor the "Application" layer to return object responses. ...

PR #209, opened 2017-09-23
FEATURE: Respect composer configured vendor-dir

DanielSiepmann:
It's possible to configure a different folder for vendor, then 'vendor'.
This way autoload.php will reside in this folder. We should check this
option to not force user to configure the path in two places.

### DIFF
--- a/lib/Container/CoreExtension.php
+++ b/lib/Container/CoreExtension.php
@@ -50,12 +50,23 @@ class CoreExtension implements ExtensionInterface
 
     public function getDefaultConfig()
     {
-        return [
+        $config = [
             'autoload' => 'vendor/autoload.php',
             'cwd' => getcwd(),
             'console_dumper_default' => 'indented',
             'cache_dir' => __DIR__ . '/../../cache',
         ];
+
+        $composerConfig = json_decode(
+            file_get_contents(getcwd() . '/composer.json'),
+            true
+        );
+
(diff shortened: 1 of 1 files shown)


DanielSiepmann:
Looks like we need to provide a `composer.json` as asset for tests, or adjust a test, as its broken due to my `file_get_contants`.

Also this solution is just hacked into the current code base. There should be a nicer way.

lib/Container/CoreExtension.php
```diff
@@ -50,12 +50,23 @@ class CoreExtension implements ExtensionInterface
...
         ];
+
+        $composerConfig = json_decode(
+            file_get_contents(getcwd() . '/composer.json'),
+            true
+        );
+
+        if (isset($composerConfig['config']['vendor-dir'])) {
+            $config['autoload'] = $composerConfig['config']['vendor-dir'] . '/autoload.php';
+        }
```
dantleech: What if we create an `Phpactoe\Config\AutoloadGuesser` class and set the above `autoload` parameter to `null` ?

The `AutoloadGuesser` can be registered as a service and used [here](https://github.com/phpactor/phpactor/pull/209/files#diff-890d76bedc8f8763259ce12c500a3b04R212) if `autoload` is null.

DanielSiepmann:
Was already implemented with e5ec0b5152f30e884327b9187984e4438bfc7e9f .

PR #211, opened 2017-09-25
changes scoped vim variables to global

cmizzi:
In order to make thing like following works : 

```vim
" Update PHPActor cwd each time a new buffer is accessed
function! UpdatePHPActorPath()
	" Change working dir to the current file
	cd %:p:h

	" Set 'gitdir' to be the folder containing .git
	let gitdir = system("git rev-parse --show-toplevel")

	" See if the command output starts with 'fatal' (if it does, not in a git repo)
	if empty(matchstr(gitdir, '^fatal:.*'))
		let g:phpactorInitialCwd = gitdir
	endif
endfunction

autocmd BufEnter *.php call UpdatePHPActorPath()
```

### DIFF
--- a/plugin/phpactor.vim
+++ b/plugin/phpactor.vim
@@ -6,16 +6,16 @@
 " | _|      |__|  |__| | _|    /__/     \__\ \______|    |__|      \______/  | _| `._____|
 "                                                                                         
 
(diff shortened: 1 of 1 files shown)


dantleech:
Makes sense I think, cheers :+1:

PR #229, opened 2017-10-10
adds debug mode

cmizzi:
When using PHPActor globally with `ncm-phpactor`, the completion will be tried each time `<C-o>` is pressed or when `ncm-phpactor` triggers will matches. But, if we're on a non-conform repository (no `Composer` or no `git`), an error will be thrown on each triggered events. This patch implements a debug mode that will throw error if true, or do nothing otherwise.

### DIFF
--- a/plugin/README.md
+++ b/plugin/README.md
@@ -73,6 +73,16 @@ nnoremap <silent><Leader>d :call phpactor#OffsetTypeInfo()<CR>
 **NOTE**: The above mappings are probably sub-optimal, feel free to find a
 something that works for you.
 
+Debug mode
+----------
+
+By default, no error will be thrown in your interface. If you want enable debug
+mode, feel free to dynamically call the following command
+
+```vim
+:let g:phpactorDebug = 1
+```
+
 Omni-completion
 ---------------
 
--- a/plugin/phpactor.vim
+++ b/plugin/phpactor.vim
@@ -10,6 +10,7 @@ let g:phpactorpath = expand('<sfile>:p:h') . '/..'
 let g:phpactorbinpath = g:phpactorpath. '/bin/phpactor'
 let g:phpactorPhpBin = 'php'
 let g:phpactorInitialCwd = getcwd()
+let g:phpactorDebug = 0
 
 """""""""""""""""
 " Update Phpactor
@@ -261,7 +262,11 @@ function! phpactor#rpc(action, arguments)
             endif
         endfor
     else
-        echo "Phpactor returned an error: " . result
+        " Only show error if debug is true
+        if (g:phpactorDebug == 1)
+            echo "Phpactor returned an error: " . result
+        endif
+
         return
     endif
 endfunction


dantleech:
Hmm, I would prefer a different solution, I would much rather see feedback when an error is encountered then not.

Would be better to detect the case of a non-conforming repository and disable phpactor.

dantleech:
> The detection must be done on the server side and return empty array on each response IMO.

Yeah, I guess that would make sense. Will have a look....

cmizzi:
Or, this kind of code instead of changing core server ?

```vim
function! phpactor#rpc(action, arguments)
    " Remove any existing output in the message window
    execute ':redraw'

    " Assert that the project handle a composer.json file and a .git directory
    if (!filereadable(g:phpactorInitialCwd . '/composer.json') || !isdirectory(g:phpactorInitialCwd . '/.git'))
        return
    endif

    let request = {"actions": [ { "action": a:action, "parameters": a:arguments } ] }

    let cmd = g:phpactorPhpBin . ' ' . g:phpactorbinpath . ' rpc --working-dir=' . g:phpactorInitialCwd
    let result = system(cmd, json_encode(request))

    if (v:shell_error == 0)
        let result = json_decode(result)

        for action in result['actions']
            let actionName = action['action']
            let parameters = action['parameters']

            let result = phpactor#_rpc_dispatch(actionName, parameters)

            if !empty(result)
                return result
            endif
        endfor
    else
		echo "Phpactor returned an error: " . result
        return
    endif
endfunction
```

dantleech:
I quickly made a PR to which should hopefully enable us to support projects which do not use composer or git:

https://github.com/phpactor/worse-reflection/pull/16

Basically it will use a normal file search to find classes. It will only work when the file is named after the class (so `Foobar.php` and `Acme\Foobar` is fine, but `Barbar.php` and `Acme\Foobar` is not).

Would that help in your specific situation?

Otherwise I think just returning an empty array as you suggested (and logging warnings) should happen anyway.

cmizzi:
I'm just thinking about global includes, etc. that doesn't really need Composer. I work on really old projects that use PHP configuration include path that link two projects (before PSR-4). So, why the fallback adapter could not following this kind of format ?

- Try looking for a class using Git or ack or whatever (FooBar class in a Toto.class.php should be found)
- If the file is not found, look into each include path directories and repeat process

But, the fallback adapter should use a cache system in order to kind in mind relative dependencies (Composer has it own)

dantleech:
> that use PHP configuration include path

yeah, good point, that would be great.

> But, the fallback adapter should use a cache system

It will

--

I guess with "old" PHP classes, they would also have `Very_Long_ClassNames` and no namespaces, so the adapter would have to be aware of that too.

dantleech:
```

Einenlum:
Is this PR still up to date, @dantleech ?

dantleech:
I guess as it's so old it should be closed. Not sure to what extent this is still and issue - the fallback locator should at least partially solve the issue.

PR #235, opened 2017-10-16
Remove deleted method from documentation

greg0ire:
It was deleted in dc97958db8bffc604a50f16e1f053c11d32a8a3d


dantleech:
Was more of a debug command - would be better to implement a navigable skeleton of the class, or show a well presented summary in the preview window.

PR #240, opened 2017-10-19
Variant docs

### LINKED ISSUE #239: Variant not taken into account
I'm trying to get final classes when calling `ClassNew()`

Issues found:

1. the README contradicts itself, `code_transform.class_new.variants` seems to work, but `new_class_variants` is not recognized
2. Under `code_transform.class_new.variants`, I believe it should not be a numeric index, otherwise you get `0` in the vim list, not very descriptive
3. When I use the following config, and select `final`, I don't get a final class, and I don't understand why there should be a for loop:
```
./templates/final.php.twig
namespace {{ prototype.namespace  }};

{% for class in prototype.classes %}
final {{ class.name  }}
{

}
{% endfor %}


./phpactor.yml
code_transform.class_new.variants:
    final: final
```

I tried renaming `final` to `SourceCode` like in the docs, to no avail.
I also tried writing `final: typo`, and did not get an exception.

greg0ire:
Closes #239

### DIFF
--- a/README.md
+++ b/README.md
@@ -417,8 +417,8 @@ $ phpactor class:new tests/Registry/GeneratorTest.php --variant=test
 Variants are registered in `.phpactor.yml`:
 
 ```yaml
-new_class_variants:
-    phpunit_test: phpunit_test
+code_transform.class_new.variants:
+    "Phpunit test": phpunit_test
 ```
 
 In order to create the above variant we need to create a template locally in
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -55,9 +55,9 @@ reflector_stub_directory:/home/daniel/www/phpactor/phpactor/lib/Container/../../
 cache_dir:/home/daniel/www/phpactor/phpactor/lib/Container/../../cache
 code_transform.indentation: "    "
 code_transform.class_new.variants:
-  - exception
-  - symfony_command
-  - phpunit_test
+  exception: exception
+  'Symfony Command': symfony_command
+  'Phpunit test': phpunit_test
```
dantleech: Using double quotes above

greg0ire:
Fixed

dantleech:
:+1: cheers :)

PR #243, opened 2017-10-20
Generate in new class

dantleech:
If refactoring happens in a new file, open other file and replace the text in that file.


lib/Console/Command/RpcCommand.php
```diff
@@ -37,6 +37,7 @@ public function execute(InputInterface $input, OutputInterface $output)
         try {
             $response = $this->processRequest($request);
         } catch (\Exception $e) {
+            file_put_contents('dump', $e->getTraceAsString());
```
dantleech: Remove, but would be good to have a way to debug failed RPC commands.

PR #250, opened 2017-10-27
Logging

dantleech:
Introduce a way to log RPC calls and write to a log file.

The goal is to enable RPC calls to be debugged when they fail.

Problems:

- This will write an insane amount of data when logging is enabled.

TODO

- [ ] Invetigate way to automatically log / dump when an exception is encountered during an RPC call.


lib/Application/Complete.php
```diff
@@ -45,7 +45,7 @@ public function complete(string $source, int $offset): array
             return [];
         }
 
-        $classReflection = $this->reflector->reflectClassLike(ClassName::fromString((string) $type));
+        $classReflection = $this->reflector->reflectClassLike(ClassName::fromString((string) $type), $source);
```
dantleech: revert

lib/Console/Command/RpcCommand.php
```diff
@@ -68,4 +63,15 @@ private function stdin()
 
         return $in;
     }
+
+    private function response(array $request): Response
+    {
+        try {
+            return $this->processRequest($request);
+        } catch (\Exception $e) {
+            return Response::fromActions([
```
dantleech: This is redundant, as a decorator now does this.

PR #268, opened 2017-11-16
Overload method

dantleech:
Overload method coming with a load of questionable refactoring in code-builder.

TODO:

- [x] VIM: Alphabet/confirm choice is inappropriate, better to use standard numeric choice
- [x] Use statements should be included


tests/Unit/Rpc/Handler/OverloadMethodHandlerTest.php
```diff
@@ -0,0 +1,91 @@
...
+
+    /**
+     * @var ObjectProphecy
+     */
+    private $overloadMethod;
+
+    public function setUp()
+    {
+        $this->reflector = Reflector::create(
+            new StringSourceLocator(SourceCode::fromString('<?php class ParentClass { public function foobar() {} }'))
```
dantleech: Because `reflectClassesIn` does not mean that classes can be located in the given source.

lib/Rpc/Handler/OverloadMethodHandler.php
```diff
@@ -0,0 +1,125 @@
...
+                'Currently will only overload methods in files with one class'
+            );
+        }
+
+        $class = $arguments[self::PARAM_CLASS_NAME] ? $classes->get($arguments[self::PARAM_CLASS_NAME]) : $classes->first();
+
+        /** @var ReflectionClass $parentClass */
+        $parentClass = $class->parent();
+
+        if (null === $parentClass) {
```
dantleech: Extract all this stuff into a method.

lib/Container/config/menu.json
```diff
@@ -98,6 +98,13 @@
                 "source_path": "%path%"
             }
         },
+        "overload_method": {
+            "action": "overload_method",
+            "parameters": {
+                "path": "%current_path%",
+                "source": "%source%"
```
dantleech: This shoud really be an offset - we should be able to determine the outer class from an offset.

lib/Container/CodeTransformExtension.php
```diff
@@ -147,12 +150,13 @@ private function registerTransformers(Container $container)
             );
         }, [ 'code_transform.transformer' => [ 'name' => 'implement_contracts' ]]);
 
-        $container->register('code_transform.transformer.add_missing_properties', function (Container $container) {
+        $container->register('code_transform.refactor.add_missing_properties', function (Container $container) {
```
dantleech: Revert

tests/Unit/Rpc/Handler/OverloadMethodHandlerTest.php
```diff
@@ -0,0 +1,91 @@
...
+class ChildClass extends ParentClass
+{
+}
+EOT
+        ;
+
+        $this->overloadMethod->overloadMethod($source, 'ChildClass', 'foobar')->willReturn(TransformSourceCode::fromString('hello'));
+
+        $action = $this->handle('overload_method', [
+            'class_name' => 'ChildClass',
```
dantleech: This test, and the other handler tests are a complete pain to maintain, should create a better environment for testing them.

tests/Unit/Console/ApplicationTest.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+namespace Phpactor\Tests\Unit\Console;
+
+use PHPUnit\Framework\TestCase;
+
```
dantleech: Remove

PR #284, opened 2017-12-06
Linter

dantleech:
TODO:

- [ ] Add regression test
- [ ] Add awesome CLI output as with reference commands.
- [ ] Finish the WR Lint feature.

This PR will eventually introduce the `lint` command:

image

Worse reflection already recognizes when something isn't right (out of necessity in a number of cases), but until now it did not associate that knowledge with anything in particular.

Although there are already great projects for static analysis, they will not necessarily tell you why Phpactor isn't completing something.

The Lint command will enable tools such as Ale or Syntastic to determine problems which f.e. prevent completion, and on a debugging level, enable much better feedback for improving Phpactor.


dantleech:
Abandoning this for now -- replacing with showing errors on omni-complete #289

PR #289, opened 2017-12-10
Error or omni

dantleech:
Show error on omni-complete if there were issues with completion (e.g. class not found, variable not defined, or whatever comes back from WorseReflection).

Also updates to new WR refactorings.

showerror

### DIFF
--- a/lib/Application/Complete.php
+++ b/lib/Application/Complete.php
@@ -2,16 +2,19 @@
 
 namespace Phpactor\Application;
 
-use Phpactor\WorseReflection\Reflector;
-use Phpactor\WorseReflection\Core\SourceCode;
-use Phpactor\WorseReflection\Core\Offset;
 use Phpactor\Application\Helper\FilesystemHelper;
+use Phpactor\WorseReflection\Core\ClassName;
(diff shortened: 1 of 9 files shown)


lib/Application/Complete.php
```diff
@@ -39,44 +42,14 @@ public function complete(string $source, int $offset): array
...
-        if ($classReflection instanceof ReflectionClass) {
-            foreach ($classReflection->properties() as $property) {
-                $suggestions[] = [
-                    'type' => 'm',
-                    'name' => $property->name(),
-                    'info' => $this->getPropertyInfo($property),
-                ];
-            }
-        }
+        $issues = [];
```
dantleech: dd

plugin/phpactor.vim
```diff
@@ -52,6 +55,11 @@ function! phpactor#Complete(findstart, base)
         endfor
     endif
 
+    if !empty(issues)
+        echoe join(issues, ', ')
+        return -3
```
dantleech: It might be somehow possible that the issue is non-blocking and we still have suggestions, should probably just continue.

PR #299, opened 2018-01-01
Support working with no git nor composer

dantleech:
This PR implements support for "brute force" class location / name conversion.

- By default we fallback to simple directory traversal for locating files (where composer / git would be much more optimal).
- `SimpleFileToClass` converts file names to class names by scanning (not parsing) the file for the namespace / class name (avoids composer dep).
- `SimpleClassToFile` converts class to file by globbing the current CWD to find the class short name, and then uses the above method to match the class name.

This is much less efficient, and some functionality will not work (e.g. generating a class - because we can't infer the class name). But it's a good starting point, it is still assumed the composer will be used eventually.h


lib/Container/CoreExtension.php
```diff
@@ -334,7 +340,7 @@ private function registerApplicationServices(Container $container)
             return new ClassCopy(
                 $container->get('application.helper.class_file_normalizer'),
                 $container->get('class_mover.class_mover'),
-                $container->get('source_code_filesystem.git')
+                $container->get('source_code_filesystem.registry')->get('git')
```
dantleech: Should really pass in the registry and FS string here.

lib/Container/SourceCodeFilesystemExtension.php
```diff
@@ -36,7 +39,13 @@ public function load(Container $container)
         $container->register('source_code_filesystem.composer', function (Container $container) {
             $providers = [];
             $cwd = FilePath::fromString($container->getParameter('cwd'));
-            foreach ($container->get('composer.class_loaders') as $classLoader) {
+            $classLoaders = $container->get('composer.class_loaders');
+
+            if (!$classLoaders) {
+                throw new NotSupported('No composer class loaders found/configured');
```
dantleech: Boundry violation

lib/Container/SourceCodeFilesystemExtension.php
```diff
@@ -45,10 +54,18 @@ public function load(Container $container)
         $container->register('source_code_filesystem.registry', function (Container $container) {
             $filesystems = [];
             foreach ($container->getServiceIdsForTag('source_code_filesystem.filesystem') as $serviceId => $attributes) {
-                $filesystems[$attributes['name']] = $container->get($serviceId);
+
+                try {
+                    $filesystems[$attributes['name']] = $container->get($serviceId);
+                } catch (NotSupported $exception)
+                {
```
dantleech: Log a warning

dantleech:
This is now working. Tested on a large project (17,000 PHP files, 106,000 files altogether) and it took around 6 seconds to locate a class.

On a fresh Symfony project it takes around 0.25s.

PR #307, opened 2018-01-13
Cache clear

dantleech:
Command / Handler / VIM function for clearing the cache.

Currently the only cached information is the phpstorm stub map used for locating standard PHP classes.


plugin/phpactor.vim
```diff
@@ -222,6 +222,11 @@ function! phpactor#Navigate()
     call phpactor#rpc("navigate", { "source_path": currentPath })
 endfunction
 
+function! phpactor#CacheClear()
+    let currentPath = expand('%')
```
dantleech: rm

PR #312, opened 2018-01-20
Extract method

dantleech:
Introduces extract method refactoring.

Still experimental.

- [x] Method name is suggested as `null` (string) by default
- [x] Case where method already exists
- [x] It asks for method name repeatedly?
- [x] Plugin: handle case where no selection has been made
- [ ] Transform: Handle case where variables are descoped before use (e.g. assigned to temp var).
- [x] Return value?

- [x] Indentation is funky
- [x] Tuple returns are not captured when using (normal) visual selection ??

recording

### DIFF
--- a/doc/vim-plugin.md
+++ b/doc/vim-plugin.md
@@ -1,5 +1,5 @@
----
 currentMenu: vim-plugin
+---
 ---
 Phpactor VIM Plugin
 ===================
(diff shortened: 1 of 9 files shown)


plugin/phpactor.vim
```diff
@@ -65,6 +65,21 @@ endfunc
 """"""""""""""""""""""""
 " Expand a use statement
 """"""""""""""""""""""""
+function! phpactor#ExtractMethod()
+    let [lineStart, columnStart] = getpos("'<")[1:2]
+    let [lineEnd, columnEnd] = getpos("'>")[1:2]
+    let offsetStart = line2byte(lineStart) + columnStart -2
+
+    if (columnEnd > 100000)
```
dantleech: In block selection mode VIM just returns a stupid big number for columnEnd ...

doc/vim-plugin.md
```diff
@@ -75,6 +75,9 @@ map <Leader>tt :call phpactor#Transform()<CR>
 
 " Generate a new class (replacing the current file)
 map <Leader>cc :call phpactor#ClassNew()<CR>
+
+" Extract method from selection
+map <Leader>em :call phpactor#ExtractMethod()<CR>
```
dantleech: Might be worth adding a selection context menu in the future

PR #314, opened 2018-01-24
Spell "class" properly

greg0ire:


dantleech:
Looks like build is broken because I use an encrypted secret, will have to look into that.

Issue #214, opened 2017-09-27
Completion: Try and be a bit more clever with visibility.


dantleech:
Currently all members are listed for completion, regardless of their visiblity.

- If not in a class. All members must be public.
- If in a class:
    - All members not of that class must be public.
    - All methods of the actual class are visible.
    - Protected and public methods of parent classes are visible.

phux:
> Protected and public methods of parent classes are visible.
This is already implemented, right?
I've tested it with `lib/Console/Command/ClassInflectCommand.php`:
- I see the parent's public & protected methods like `ignoreValidationErrors()` and `interact()`
- I don't see the parent's private methods like `validateName`

dantleech:
No, I clearly see the private methods. The auto complete  logic is [very simple](https://github.com/phpactor/phpactor/blob/master/lib/Application/Complete.php#L52). The idea is to break it out into a separate component to be more aware of context and provide completions for new classes, use statements, functions etc.

phux:
Weird! Here is what I get (latest master):

![2017-11-13-092311_956x1055_scrot](https://user-images.githubusercontent.com/852372/32715986-913b2dfa-c854-11e7-9aaa-900df5fe9f4e.png)

As you can see, only the private properties and methods of the ClassInflectCommand are shown, nothing private from its parent.

dantleech:
Indeed, very strange. Are you using the `neo-complete` plugin (for neovim) or the omni-complete? Maybe you have some other plugin which is somehow filtering? (I would say maybe another plugin is doing the completion by the abbreviations look like Phpactor).

I have:

![untitled](https://user-images.githubusercontent.com/530801/32726851-bb623cec-c87a-11e7-8962-8622ffed97db.png)

In anycase, Phpactor certainly isn't doing it... yet :)

phux:
I am using omnicomplete via `<c-x><c-o>` with `:set omnifunc=phpactor#Complete`.

phux:
Indeed phpactor does it, or better the `worse-reflection` package.

`worse-reflection/lib/Bridge/TolerantParser/Reflection/ReflectionClass.php` Line 147

dantleech:
Completetion is not a "bit more" clever. closing. :)

Issue #109, opened 2017-08-07
Auto-complete library


dantleech:
Create a new library which focuses on auto-complete to replace basic functionality included in `phpactor/phpactor`:

- Member access completion as now.
- `new` class instantation completion (using class search)
- local variable name completion.
- method auto-completion (override)
- extends, implements, trait autocompletion (by "class" name only initially)

PR #316, opened 2018-01-24
Completion: hide private/protected methods/properties

phux:
Just a naive quick fix for hiding private/protected methods & properties, as you mentioned rewriting a new completion engine in #109 and #214

### DIFF
--- a/lib/Application/Complete.php
+++ b/lib/Application/Complete.php
@@ -160,13 +160,13 @@ class Complete
             return $symbolInformation->withIssue(sprintf('Could not find class "%s"', (string) $type));
         }
 
-        $needsPublicAccess = !in_array($symbolInformation->symbol()->name(), ['this', 'self']);
+        $publicOnly = !in_array($symbolInformation->symbol()->name(), ['this', 'self'], true);
         /** @var $method ReflectionMethod */
         foreach ($classReflection->methods() as $method) {
             if ($method->name() === '__construct') {
                 continue;
             }
-            if ($needsPublicAccess && !$method->visibility()->isPublic()) {
+            if ($publicOnly && false === $method->visibility()->isPublic()) {
                 continue;
             }
             $info = $this->getMethodInfo($method);
@@ -179,7 +179,7 @@ class Complete
 
         if ($classReflection instanceof ReflectionClass) {
             foreach ($classReflection->properties() as $property) {
-                if ($needsPublicAccess && !$property->visibility()->isPublic()) {
+                if ($publicOnly && false === $property->visibility()->isPublic()) {
                     continue;
                 }
                 $suggestions[] = [
--- a/lib/Application/Complete.php
+++ b/lib/Application/Complete.php
@@ -160,9 +160,13 @@ class Complete
             return $symbolInformation->withIssue(sprintf('Could not find class "%s"', (string) $type));
         }
 
+        $needsPublicAccess = !in_array($symbolInformation->symbol()->name(), ['this', 'self']);
         /** @var $method ReflectionMethod */
         foreach ($classReflection->methods() as $method) {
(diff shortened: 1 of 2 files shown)


dantleech:
Nice! Will have a look this evening. Build is failing because of an encrypted secret in travis, but otherwise tests pass.

lib/Application/Complete.php
```diff
@@ -160,11 +160,15 @@ private function populateSuggestions(SymbolContext $symbolInformation, Type $typ
             return $symbolInformation->withIssue(sprintf('Could not find class "%s"', (string) $type));
         }
 
+        $needsPublicAccess = !in_array($symbolInformation->symbol()->name(), ['this', 'self']);
```
dantleech: Rename to `$publicOnly`? also `false === in_array` rather than `!`

lib/Application/Complete.php
```diff
@@ -160,11 +160,15 @@ private function populateSuggestions(SymbolContext $symbolInformation, Type $typ
...
         /** @var $method ReflectionMethod */
         foreach ($classReflection->methods() as $method) {
             if ($method->name() === '__construct') {
                 continue;
             }
+            if ($needsPublicAccess && !$method->visibility()->isPublic()) {
```
dantleech: `false === $method->visibility()->isPublic()`

lib/Application/Complete.php
```diff
@@ -175,6 +179,9 @@ private function populateSuggestions(SymbolContext $symbolInformation, Type $typ
 
         if ($classReflection instanceof ReflectionClass) {
             foreach ($classReflection->properties() as $property) {
+                if ($needsPublicAccess && !$property->visibility()->isPublic()) {
```
dantleech: same

lib/Application/Complete.php
```diff
@@ -160,11 +160,15 @@ private function populateSuggestions(SymbolContext $symbolInformation, Type $typ
             return $symbolInformation->withIssue(sprintf('Could not find class "%s"', (string) $type));
         }
 
+        $needsPublicAccess = !in_array($symbolInformation->symbol()->name(), ['this', 'self']);
         /** @var $method ReflectionMethod */
         foreach ($classReflection->methods() as $method) {
```
dantleech: I think the next step (after this PR) would be to add `$classReflection->methods()->visibleTo($symbolInformation->containerType())`

lib/Application/Complete.php
greg0ire: `in_array` only returns booleans though :P

greg0ire: I'd be more worried about not setting its 3rd argument to `true`

phux: I agree with greg0ire.

@dantleech if you want or it's the coding style of this project I will change it accordingly.

dantleech: > in_array only returns booleans though :P

Strict comparators and boolean keywords are used mostly in the code base - I find it easier to read, but its subjective. Main thing is consistency.

dantleech:
Thanks!

lib/Application/Complete.php
dantleech: > if you want or it's the coding style of this project I will change it accordingly.

Nah, its fine. Not sure what my algorithm is for the two styles anyway.

Taluu:
Didn't check, but what about parent's protected methods ? And what about static ? :}

Otherwise, just what I've been looking for and complaining about just before updating phpactor :P

dantleech:
Yep, its a naieve implementation - if the variable is `self` or `this` then it will show all visibilities, otherwise only public.

Taluu:
And what if I have `$self`, will it show privates for this one too ?

dantleech:
`self::` and `$this->`

PR #341, opened 2018-03-20
Use relative link

greg0ire:
Pretty sure localhost is not the intended host for this.


dantleech:
Cool. Looks like Travis is failing for reasons.

PR #345, opened 2018-03-27
Add documentation for phpactor vim plugin

joereynolds:
Previously there wasn't any documentation.
I've basically copy-pasta'd the documentation and formatted it.

Now you can do `:help phpactor`.

:)

### DIFF
--- a/doc/phpactor.txt
+++ b/doc/phpactor.txt
@@ -1,0 +1,123 @@
+*phpactor* *phpactor.txt* PHP introspection tools
+
+                          PHPACTOR REFERENCE MANUAL~
+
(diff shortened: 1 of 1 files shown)


dantleech:
Awesome :) Would be even better if we could generate this documentation, maybe using `pandoc`

joereynolds:
Probably doable but I imagine they would diverge at some point. Also I imagine most people wanting to read the help for it would do so in vim? For the plugin at least

dantleech:
Merging now, it would be good to automate this in the future. Or indeed, not document the VIM functionality on the website and get the user to use VIM `:help` instead

PR #351, opened 2018-03-30
Class import

dantleech:
- [x] Handle class importing in Phpactor and not in the VIM plugin
- [x] Will ask for an Alias if the class is already imported

TODO:

- [x] <strike>Currently the code updater makes unnecessary changes (e.g. reformatting method declarations), this should be fixed before merging this as it's a very frequent operation.</strike> this was already fixed.

### DIFF
--- a/lib/Rpc/Handler/ImportClassHandler.php
+++ b/lib/Rpc/Handler/ImportClassHandler.php
@@ -1,0 +1,138 @@
+<?php
+
+namespace Phpactor\Rpc\Handler;
+
+use Phpactor\Rpc\Handler\AbstractHandler;
+use Phpactor\CodeTransform\Domain\Refactor\ImportClass;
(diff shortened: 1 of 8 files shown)


lib/Rpc/Handler/ImportClassHandler.php
```diff
@@ -0,0 +1,138 @@
...
+        ];
+    }
+
+    public function handle(array $arguments)
+    {
+        if (null === $arguments[self::PARAM_QUALIFIED_NAME])    {
+            $suggestions = $this->suggestions($arguments[self::PARAM_NAME]);
+            $suggestions = array_map(function (array $suggestion) {
+                return $suggestion['class'];
+            }, $suggestions);
```
dantleech: Move this into the method

PR #355, opened 2018-03-31
Replay and paths

dantleech:
Adds a `--replay` option to the RPC command, allowing commands executed in f.e. VIM to be replayed in the CLI.

### DIFF
--- a/lib/Config/Paths.php
+++ b/lib/Config/Paths.php
@@ -1,0 +1,70 @@
+<?php
+
+namespace Phpactor\Config;
+
+use Webmozart\PathUtil\Path;
+use XdgBaseDir\Xdg;
+
+class Paths
+{
+    /**
+     * @var Xdg
+     */
+    private $xdg;
+
+    /**
+     * @var string
+     */
+    private $cwd;
+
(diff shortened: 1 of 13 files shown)


lib/Config/Paths.php
```diff
@@ -0,0 +1,70 @@
...
+            return Path::join($configPath, '/phpactor.yml');
+        }, $this->configPaths());
+
+        $localConfigPath = Path::join($this->cwd, '.phpactor.yml');
+        $paths[] = $localConfigPath;
+
+        return $paths;
+    }
+
+    public function configPaths(string $suffix = ''): array
```
dantleech: `null`

lib/Console/Command/ConfigDumpCommand.php
```diff
@@ -22,20 +23,20 @@ class ConfigDumpCommand extends Command
...
+     * @var Paths
      */
-    private $configLoader;
+    private $paths;
 
     public function __construct(
         array $config,
         DumperRegistry $registry,
-        ConfigLoader $configLoader
+        Paths $paths  = null
```
dantleech: Not null

PR #356, opened 2018-03-31
Updated WR for support for Foreach

### LINKED ISSUE #292: Support for "generic collections types"
Don't know if it's implemented (if so you can close...), but supporting "generic-like" types would be awesome :

```php
<?php
use Doctrine\Common\Collections\Collection;

/** @var Collection<Foo> | Foo[] */
$foos = new ArrayCollection(...);
```

WDYT ? Even if it's not really something recognized by php, it is I think used enough to offer a support for that (there is even a RFC in php to allow that : https://wiki.php.net/rfc/generics and https://wiki.php.net/rfc/generic-arrays)

dantleech:
Support for inferring foreach item types. Closes #292 

foreach

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -354,12 +354,12 @@
             "source": {
                 "type": "git",
                 "url": "https://github.com/phpactor/code-transform.git",
-                "reference": "a163601c241921e6d55bcbc8c3fb84d3e38ce431"
+                "reference": "03d4c89fc92553724bf3c67ba2be9687dc81a9aa"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/phpactor/code-transform/zipball/a163601c241921e6d55bcbc8c3fb84d3e38ce431",
-                "reference": "a163601c241921e6d55bcbc8c3fb84d3e38ce431",
+                "url": "https://api.github.com/repos/phpactor/code-transform/zipball/03d4c89fc92553724bf3c67ba2be9687dc81a9aa",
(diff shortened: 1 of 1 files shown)


dantleech:
Note this only works with `@{var|method|return} Foobar[]` annotations.

Not sure what the best practice is for collections ? @greg0ire @Taluu 

Is there a way to declare this on the collection class itself?

Taluu:
Collection types are usually generic types. At least, the `type[]` notation shows that it's iterable, which is generic enough.

But `Type<Subtype>` (e.g, working with doctrine's collections) can also be used, which can be used as an iteable (through `foreach ...`), or as an object (`$collection->count()`, and so on).

But as least this would be able to cover most cases, so 👍

dantleech:
it seems `@var Collection|User[] $users` is what people recommend on google (which should also work now).

But would really like something like:

```
/**
 * @collectionOf Foobar
 */
class Collection
{
}
```
To avoid the client type-hint completely ...

Taluu:
AFAIK, you don't create a collection type class to set the type (your `@CollectionOf`). So `@var Collection|Foo[]` should work, but generics would be more common I think (`@var Collection<Foo>`. It's even supported by the phpdoc psr : https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md#abnf

dantleech:
Just merged #360 which allows `/** @var Generator<NamespacedNameInterface> $nodeCollection */`

PR #367, opened 2018-04-01
Implement completion library

### LINKED ISSUE #109: Auto-complete library

dantleech:
Will fix #109 

Outsource the completion to a [separate library](https://github.com/phpactor/completion) and provide:

- [x] Existing member completion
- [x] Local variable completion
- [ ] Class name completion (`use`, `new`, `extends` etc)
- [ ] Method autocompletion? (overriding, but then the override refactoring does this better...)
- [ ] ...

Completion will be extensible and configurable.

And eventually:

- [ ] Function name completion (when we can search them)

### DIFF
--- a/lib/Container/CompletionExtension.php
+++ b/lib/Container/CompletionExtension.php
@@ -1,0 +1,36 @@
+<?php
+
+namespace Phpactor\Container;
+
+use PhpBench\DependencyInjection\ExtensionInterface;
+use PhpBench\DependencyInjection\Container;
+use Phpactor\Completion\Completor;
+use Phpactor\Completion\Completor\ClassMemberCompletor;
+use Phpactor\Completion\Completor\LocalVariableCompletor;
+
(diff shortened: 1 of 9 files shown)


lib/Container/CompletionExtension.php
```diff
@@ -0,0 +1,36 @@
...
+     * {@inheritDoc}
+     */
+    public function load(Container $container)
+    {
+        $container->register('completion.completor', function (Container $container) {
+
+            $reflector = $container->get('reflection.reflector');
+            $completors = [
+                new ClassMemberCompletor($reflector),
+                new LocalVariableCompletor($reflector),
```
dantleech: Make these taggable...

PR #386, opened 2018-04-07
Use HTTPS for documentation site

Garbee:
Changing the baseUrl to `https` allows access to the doc site using an authenticated connection.

Before this change, the docs site will throw mixed content errors and not render properly.


dantleech:
Great, thanks!

PR #389, opened 2018-04-08
Container and Extension Refactoring

dantleech:
This PR aims to refactor the current container / extension system.

The aims are:

- [x] Provide better configuration (better validation, nested configs etc.)
- [x] Provide a stable extension point
- [x] Separate the mixed classes in `Application`, `Rpc` etc. into dedicated extensions.

## Changes

The current system is based on a Phpbench package which provides a _configurable_ container with tags. But it's not ideal, as third-party extensions shouldn't have to depend on a PHPBench package, and the configuration is badly coupled to the container implementation.

This PR introduces a custom container, with a decoupled extension system. Each extension can register services with the container, and define a configuration _schema_:

```php
use Phpactor\Extension\Extension;
use Phpactor\Extension\ContainerBuilder;
use Phpactor\Extension\Container;
use Phpactor\Extension\Schema;

class PathFinderExtension implements Extension
{
    const PATH_FINDER_DESTINATIONS = 'navigator.destinations';

    /**
     * {@inheritDoc}
     */
    public function load(ContainerBuilder $container)
    {
        $container->register('path_finder.path_finder', function (Container $container) {
            return PathFinder::fromDestinations($container->getParameter(self::PATH_FINDER_DESTINATIONS));
        });
    }

    /**
     * {@inheritDoc}
     */
    public function configure(Schema $schema)
    {
        $schema->setDefaults([
            self::PATH_FINDER_DESTINATIONS => [],
        ]);
    }
}
```

The container API has remained almost exactly the same, the PHPBench extension interface's `getDefaultConfig` method has been replaced with `configure`


Taluu:
Instead of making your own container, you should have used an existing one. I get that you don't want 3rd party extensions to depend on a 3rd party container, but IMO this is a false argument... At least, let them depend on the `ContainerInterface` and then let the people choose the container of their own choice.

But once again, IMO, I think using yet another custom container is not really a good idea when there are so many good containers (symfony, php-di, ...).

dantleech:
> At least, let them depend on the ContainerInterface and then let the people choose the container of their own choice.

We _are_ implementing the PSR container, but this doesn't really help as you as Phpator will still want you to use a concrete implementation.

> when there are so many good containers (symfony, php-di, ...).

These containers need to be compiled to be performant (not an insurmountable problem, but it complicates things for this application) . They also have a lot of moving parts and unnecessary (in this application) features.

> Instead of making your own container

The container is pretty much as simple as can be (you can make it in 10 minutes) and because of this it is technically the [fastest container](https://github.com/phpbench/benchmarks-containers#time) out there.

Existing lightweight containers do not have tags, and if any do exist they are not mainstream enough to depend on.

As a more minor point it also means I don't have to spend a day rewriting every line of container configuration.

Saying all that, this is not yet a public extension point, so can still be changed.

PR #410, opened 2018-04-14
Update vim plugin markdown docs with description and link to deoplete integration.

kristijanhusak:


kristijanhusak:
@dantleech not sure if the content and structure is ok, but i didn't want to put too much stuff in here, because most of the vim users already know how to do the installation.
If something needs to be changed, let me know, or fix it after it's merged.

dantleech:
I think that's fine for now, I have a `Vundle` example at the top, would be good to be consistent (maybe switch to using `vim plug` if that's the best choice).

PR #436, opened 2018-04-21
Optimise replace private / protected members

### LINKED ISSUE #349: [CM] Limit change scope to local class heirarchy when renaming private members.
Currently the whole project tree is considered when renaming local private / protected members. We should limit the scope to the current class (private) and the class hierarchy (protected).

Also fix the file reloading in the VIM plugin.

dantleech:
- [x] When a known protected / private class member is given, only traverse the possible related files,
- [x] Do not ask for the filesystem scope (default to either composer or full FS depending on env.).
- [x] Issue a reload request for the current file in the RPC response.
- [x] Replace references in current file (not just those on the disk) for clkass members.
- [x] Same for class references

**NOTE**: The application code for this is quite frightning, should extract it see #440 

Fixes #349

### DIFF
--- a/lib/Extension/ClassMover/Application/Finder/FileFinder.php
+++ b/lib/Extension/ClassMover/Application/Finder/FileFinder.php
@@ -1,0 +1,85 @@
+<?php
+
+namespace Phpactor\Extension\ClassMover\Application\Finder;
+
+use Phpactor\Filesystem\Domain\Filesystem;
+use Phpactor\WorseReflection\Core\Reflection\ReflectionClass;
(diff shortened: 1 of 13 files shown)


lib/Extension/ClassMover/Application/Finder/FileFinder.php
```diff
@@ -0,0 +1,100 @@
...
+            throw new RuntimeException(sprintf(
+                'Source has no path associated with it'
+            ));
+        }
+        
+        if ($private) {
+            return FileList::fromFilePaths($filePaths);
+        }
+        
+        $context = $reflection;
```
dantleech: `$reflection->parent()`, should probably extract method

PR #444, opened 2018-04-23
Introducing Tests for the VIM Plugin

dantleech:
Introduces [Vader](https://github.com/junegunn/vader.vim) for VIM testing, adds to Travis CI :crossed_fingers: 

- [x] Only run the VIM tests in one build.

Bonus points:

- [ ] Test on both VIM 8 and Neovim


plugin/tests/insert_use.vader
```diff
@@ -9,74 +9,13 @@ Given php (source with no namespace):
   }
 Do (put the cursor over an existing class name and add use):
   /Phpactor\<Enter>
-  :call PhactUseAdd()\<Enter>
+  :call phpactor#UseAdd()\<Enter>
 
 Expect php (the use statement to be inserted):
   <?php
-  use Phpactor\Phpactor;
```
dantleech: This test has been here for ages, and never worked before. All of the test cases are now handled in PHP.

PR #460, opened 2018-04-29
Standalone usage

talbergs:
I hoped to use phpactor as standalone binary that just sits in my PATH. 

Once I do `composer global require phpactor/phpactor dev-develop`, then I can call that binary from anywhere. But when I do, I get this message:

<a href="https://cl.ly/1k040J382N2s" target="_blank"></a>

I noticed the path here tries to reach within phpactor/vendor folder that is now levels up (at standalone usage).

This may not be the best way to fix this issue, but it works.

I have to point out this problem, as I'd like to access the binary from global path so it's easily reached/wrapped by my vim commands, or any other tools.

### DIFF
--- a/bin/phpactor
+++ b/bin/phpactor
@@ -13,7 +13,7 @@ foreach ([
 ] as $file) {
     if (file_exists($file)) {
         $autoloadFile = $file;
-        define('__VENDOR_DIR__', realpath(basename($autoloadFile)));
+        define('__VENDOR_DIR__', realpath(dirname($autoloadFile)));
         break;
     }
 }
--- a/bin/phpactor
+++ b/bin/phpactor
@@ -13,7 +13,7 @@ foreach ([
 ] as $file) {
     if (file_exists($file)) {
         $autoloadFile = $file;
-        define('VENDOR_DIR', realpath(basename($autoloadFile)));
+        define('__VENDOR_DIR__', realpath(basename($autoloadFile)));
         break;
     }
 }
--- a/bin/phpactor
+++ b/bin/phpactor
@@ -13,6 +13,7 @@ foreach ([
 ] as $file) {
     if (file_exists($file)) {
         $autoloadFile = $file;
+        define('VENDOR_DIR', realpath(basename($autoloadFile)));
         break;
     }
 }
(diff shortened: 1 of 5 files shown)


talbergs:
Another push for "paths does not work", I suspect this will happen in a lot of places. Can we come up with global fix for this standalone usage somehow? Of course we can check for paths thru-out package, and make a note of technical debt. Sure it's messy and extra expense.

But this now at least yield some results! 🌅 
```sh
>>> echo '{"action":"complete","parameters":{"source":"<?php func_g","offset":12}}' | phpactor rpc
>>> {"action":"return","parameters":{"value":{"suggestions":[{"type":"f","name":"func_get_arg","info":"func_get_arg(int $arg_num): <unknown>"},{"type":"f","name":"func_get_args","info":"func_get_args(): array"}],"issues":[]}}}
```

prior this commit there were the same path errors..

any how, it seems to work only when output is piped (as in the snippet above), else this happens https://cl.ly/3G3L323u1h3s is that the desired behaviour?

talbergs:
> would it be useful to accept it also as an argument?

No, not necessarily. But the lack of RPC documentation may be yet another issue (if the protocol is not going to drastically change it's shape any soon).

At `phpactor:complete` description says it's deprecated.

Ok, then I'll be happy to push a global helper method implementation into this PR.

talbergs:
It's good to merge now.

Please note, one convention is being introduced: a constant named `__VENDOR_DIR__` that is set at binary invocation. It will be the path to either global or local vendor directory, so use that as prefix to reach for files (that are not auto loaded) from other packages.

> I suspect this will happen in a lot of places.
>> Yeah, we should probably define a helper method to get the vendor dir, I think.

I or somebody could update other paths later, once needed, as they encounter the "path" problem.
I choose not to go for a global helper as that would be minor performance hit (better to determine path once)

TestCase assumes the local vendor directory is used.

dantleech:
I created an alernative that does not introduce a global state dependency: https://github.com/phpactor/phpactor/pull/462 , the whole thing bootstrapping phase is a bit of a mess to be honest.

talbergs:
I see you decided to go down that rabbit hole of refactoring, good luck (honestly)!
Looks like WIP for now, yes? https://cl.ly/3C1i2I0B1U37, I tested that PR - same error.

dantleech:
Yes, should be better now - I think I might bite the bullet and just port the whole thing to the Symfony DI Container and Configuration framework.

Issue #405, opened 2018-04-13
Use of Symfoy DI Container?


dantleech:
In `develop` everything has been split into extensions using a custom DI container / extension system - as this was the most simple way to achieve the end-goal.

Using the Symfony DI container was considered, but initially rejected as it introduces complications (i.e. needs to be compiled to be efficient) and would be a significant amount of work to port the existing configurations as XML (using plain-old-PHP for SF DI Config was horrible last time I checked).

Using the SF DI Container would have the following benefits:

- Package level visibility (?): Only expose extension's public services.
- Auto-wiring: Reducing development time, easier refactoring.
- Developer familiarity.
- Stable extension system.
- Make use of some existing SF infrastructure maybe? `debug:config` etc.

Additionally we would probably also adopt the Symfony Configuration component.

It would require that we create a per-project / instance container cache (and an option to disable the cache I suppose).

PR #462, opened 2018-04-30
Use actual vendor dir

dantleech:
This is another approach to #460 which avoids the global state dependency.

To be honest, the whole bootstrapping process needs to be refactored and probably migrated to use the Symfony DI container and Configuration, as with #405

### DIFF
--- a/lib/Container/PhpactorContainer.php
+++ b/lib/Container/PhpactorContainer.php
@@ -58,7 +58,7 @@ class PhpactorContainer implements Container, ContainerBuilder
      */
     public function has($id)
     {
-        return isset($this->factories[$id]);
+        return array_key_exists($id, $this->factories);
     }
 
     public function getServiceIdsForTag(string $tag): array
--- a/lib/Container/PhpactorContainer.php
+++ b/lib/Container/PhpactorContainer.php
@@ -85,7 +85,7 @@ class PhpactorContainer implements Container, ContainerBuilder
 
     public function getParameter(string $name)
     {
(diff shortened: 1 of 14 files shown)


lib/Container/PhpactorContainer.php
```diff
@@ -85,7 +85,7 @@ public function register(string $serviceId, Closure $factory, array $tags = [])
 
     public function getParameter(string $name)
     {
-        if (!isset($this->parameters[$name])) {
+        if (!array_key_exists($name, $this->parameters)) {
```
dantleech: Add a test for this

lib/Phpactor.php
```diff
@@ -53,12 +53,16 @@ public static function boot(InputInterface $input): PhpactorContainer
...
-        $container->register('config.paths', function () {
-            return new Paths();
+        $paths = new Paths();
+        $container->register('config.paths', function () use ($paths) {
+            return $paths;
         });
 
+        if (!isset($config['cache_dir'])) {
+            $config['cache_dir'] = $paths->userCacheDir();
+        }
```
dantleech: Argh.

talbergs:
That's awesome, can we merge it?
Thou one minor thing I noticed - https://cl.ly/2H3U1n0b2x1H
Of course I do not need this configurable, just noticed...

dantleech:
> Of course I do not need this configurable,

Yeah, it's an immutable parameter, but I wanted to keep it in the "config" so that it can be inspected. Ideally should be handled better, but will hopefully change all of this at some point.

PR #472, opened 2018-05-06
Parameter completion

dantleech:
Parameter completion:

untitled

- [x] Suggests local variables which are valid for the current parameter position
- [x] Shows parameter type and index
- [x] Preserves other completions, but has a higher priority
- [x] Handle case where containing class for call cannot be determined (for whatever reason)
- [x] Add issue if parameter index exceeds parameter arity.
- [x] Function parameter completion
- [x] Support triggering on `(` bracket.
- [x] Support after `,`
- [x] Support static mehtod calls....
- [x] Completion candidates include variables after the offset.
- [x] Annoying bug:

```php
    public function doRun(InputInterface $input, OutputInterface $output)
    {
        $this->doRun($<>
        $this->initialize($input);
```

```
string(27) "$                              
        $this->initialize"                 
{"action":"return","parameters":{"value":{"suggestions":[],"issues":["Could not reflect function \/ method"]}}}%                                                              
```

- [x] :-1: truncating the source means that any structural definitions after the offset are not found.

### DIFF
--- a/lib/Extension/Completion/CompletionExtension.php
+++ b/lib/Extension/Completion/CompletionExtension.php
@@ -9,6 +9,7 @@ use Phpactor\Completion\Bridge\WorseReflection\Formatter\VariableFormatter;
 use Phpactor\Completion\Bridge\TolerantParser\ChainTolerantCompletor;
 use Phpactor\Completion\Bridge\TolerantParser\WorseReflection\WorseClassMemberCompletor;
 use Phpactor\Completion\Bridge\TolerantParser\WorseReflection\WorseBuiltInFunctionCompletor;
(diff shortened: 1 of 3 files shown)


lib/Extension/Completion/CompletionExtension.php
```diff
@@ -65,6 +66,13 @@ private function registerCompletion(ContainerBuilder $container)
                 $container->get('reflection.tolerant_parser')
             );
         }, [ 'completion.completor' => []]);
+
+        $container->register('completion.completor.param_function', function (Container $container) {
```
dantleech: s/param_function/param

PR #480, opened 2018-05-22
[Vim Plugin] Goto reference in a modified file causes warning fixes #477

### LINKED ISSUE #477: g(o)to_definition not working on unsaved file.
```
Error detected while processing function phpactor#ContextMenu[1]..phpactor#rpc[15]..phpactor#_rpc_dispatch[116]..phpactor#rpc[15]..phpactor#_rpc_dispatch[54]..phpactor#_switchToBufferOrEdit:                                                                                   
line    8:
E37: No write since last change (add ! to override)
Error detected while processing function phpactor#ContextMenu[1]..phpactor#rpc[15]..phpactor#_rpc_dispatch[116]..phpactor#rpc[15]..phpactor#_rpc_dispatch:
line   55:
E37: No write since last change (add ! to override)
Press ENTER or type command to continue
```

This happens when you edit a file but do not save it, then open the context menu and attempt to g(o)to_definition.
As soon as you save the file the error goes away.

Reproduced on
NVIM v0.2.2
Ubuntu 18.04
phpactor installed via vim-plug

dantleech:
- [x] Fixes #477 :crossed_fingers: 
- [x] Adds regression to test for goto definition, because why not.

### DIFF
--- a/plugin/phpactor.vim
+++ b/plugin/phpactor.vim
@@ -376,7 +376,6 @@ function! phpactor#_rpc_dispatch(actionName, parameters)
     " >> open_file
     if a:actionName == "open_file"
         call phpactor#_switchToBufferOrEdit(a:parameters['path'])
-        exec ":edit"
 
         if (a:parameters['offset'])
             exec ":goto " .  (a:parameters['offset'] + 1)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+## develop
+
+Bug fixes:
+
+  - [Vim Plugin] Goto reference in a modified file causes warning #477.
+
 ## 2018-05-20 0.5.0
 
 Features:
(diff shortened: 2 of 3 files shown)


plugin/phpactor.vim
```diff
@@ -376,7 +376,6 @@ function! phpactor#_rpc_dispatch(actionName, parameters)
     " >> open_file
     if a:actionName == "open_file"
         call phpactor#_switchToBufferOrEdit(a:parameters['path'])
-        exec ":edit"
```
dantleech: This was here for a reason probably - for cases where the file _has_ been modified on disk and should be reloaded.

dantleech: cases where this might happen:

- Replace references in current file
- Generate method in new file
- ??

Works on all of the above

dantleech: .. and if files _have_ been changed on disk, VIM should warn about this.

PR #492, opened 2018-06-02
Transformer fix namespace

### LINKED ISSUE #474: Mapping to update namespace?
In Atom I used to have a package that would let me update namespace.

For example, I could use a `class` snippet to generate a blank class, ie)

```php
<?php

class MyModel
{
    // ...
```

Then I would save the file into a PSR-4 autoloaded folder, and then hotkey to `Update Namespace`, and it would try to intelligently insert the php namespace based on the composer autoloading, ie)

```php
<?php

namespace App\Models;

class MyModel
{
    // ...
```

Similarily, if you manually moved a file into a new PSR-4 folder, and ran `Update Namespace`, it would just refactor the namespace to be correct again.

I love how you've implemented the `use` import functionality.  Think it's possible to implement a similar `update namespace` mapping?

dantleech:
Transformer to update the class name and namespace.

Fixes #474


doc/refactorings.md
```diff
@@ -293,6 +295,68 @@ class Post
 }
 ```
 
+Fix Namespace or Class Name
+---------------------------
+
+Update the files namespace (and/or class name) based on the composer
```
dantleech: s/the/a

doc/refactorings.md
```diff
@@ -293,6 +295,68 @@ class Post
...
+configuration.
+
+- **Command**: `$ phpactor class:transform path/to/class.php --transform=complete_constructor`
+- **VIM plugin**: _Class context menu > Complete Constructor_.
+- **VIM function**:`:call phpactor#Transform()`
```
dantleech: This doc is incomplete

plugin/phpactor.vim
```diff
@@ -146,7 +146,7 @@ function! phpactor#OffsetTypeInfo()
 endfunction
 
 function! phpactor#Transform()
-    let currentPath = expand('%')
+    let currentPath = expand('%:p')
```
dantleech: Seems odd that this is necessary

PR #506, opened 2018-06-20
Allow transform command to act on globs and changed behavior

dantleech:
- Transform command accepts globs
- Writes to files (unless in dry-run mode)
- Shows diffs if requested

This allows, for example, recursively fixing namespaces after updating a project namespace:

```
$ phpactor code:transform "lib/**/*.php" --transform=fix_namespace_class_name
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ Changelog
 Features:
 
   - [CodeTransforn] Extract expression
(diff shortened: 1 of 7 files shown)


cmizzi:
It looks like to be `class:transform` instead of `code:transform` :)

dantleech:
Ah yeah, typo in the PR description, it's `class:transform` everywhere else :)

PR #511, opened 2018-06-27
Fixed extracting expressions on single elements, fixes #501

### LINKED ISSUE #501: Extract expression
Given:

```php
        $files = new ArrayFileIterator([
            <>new SplFileInfo('path/one'),
            new SplFileInfo('path/two'),
        ]);
```

When invoking extract expression on a single offset:

```php
        $foobar = new SplFileInfo('path/one'),
            new SplFileInfo('path/two'),;
        $files = new ArrayFileIterator([
            $foobar
        ]);

```

The expression is determined to be the contents of the array statement instead of the individual element.

dantleech:
Fixed extract expression bug detailed in #501 and downgraded composer to the 1.x branch (as the 2.x branch now requires PHP 7.2).

Also, the `composer update` seems to have brought in `constant completion` feature from the completion library which was apparently not available in develop yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ Improvements:
   - [Completion] Support namespaced functions, fixes #473
   - [Docs] Added section on completion
 
+Misc
+
+  - Downgraded composer to 1.x as 2.x-dev now requires PHP 7.2
+
 ## 2018-06-16 0.6.0
 
 Features:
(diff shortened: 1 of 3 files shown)


dantleech:
Failure seems to be a problem with a newer version of VIM on Ubuntu Trusty, going to merge this now as cannot reproduce this locally at the moment.

Issue #512, opened 2018-06-28
PHPActor fails with logging disabled


Miljar:
I updated phpactor to v0.1 and now it stopped working. I have `logging.enabled: false` in my local `.phpactor.yml`.

When I change that to `true` however, everything works as it's supposed to work.
Switching back to `false` introduces the errors again.

![screen shot 2018-06-28 at 11 52 17](https://user-images.githubusercontent.com/472386/42027340-c438363a-7ac9-11e8-8e37-d7a33250ffd0.png)

dantleech:
Thanks for reporting, this looks like it may be related to me downgrading Monolog in `develop` yesterday.

This actually caused the VIM tests to fail with the same error above, but I assumed it was a problem with the version of VIM Travis was using as I could not reproduce it locally.

I'll try and reproduce this and fix ASAP, in the mean time you could switch to `master` branch.

As a side note, Phpactor has no versioning, it's either `master` or `develop`.

dantleech:
https://github.com/phpactor/phpactor/issues/512 should do it, continuing there.

Should really find a way to automatically update that number :)

weirdan:
> Should really find a way to automatically update that number :)

https://packagist.org/packages/ocramius/package-versions, or, for legacy php versions, https://packagist.org/packages/muglug/package-versions-56

dantleech:
Should be ok now, will have a look at package-versions thanks @weirdan

PR #514, opened 2018-06-28
Fetch phpactor's version from composer local package registry

weirdan:
Sample output:
image

Refs https://github.com/phpactor/phpactor/issues/512#issuecomment-401067095


dantleech:
Awesome! I'll add an entry to the changelog

PR #520, opened 2018-07-02
Soft implementation of macros

dantleech:
This PR integrates the macro work in `code-transform` and just adjusts the existing RPC handlers to call `__invoke` instead of the previous methods.


lib/Extension/CodeTransform/CodeTransformExtension.php
```diff
@@ -11,14 +11,14 @@
 use Phpactor\CodeTransform\Adapter\Native\GenerateNew\ClassGenerator;
 use Phpactor\CodeTransform\Adapter\TolerantParser\ClassToFile\Transformer\ClassNameFixerTransformer;
 use Phpactor\CodeTransform\Adapter\TolerantParser\Refactor\TolerantExtractExpression;
```
dantleech: This is a macro too...

dantleech:
Benching this for now.

PR #530, opened 2018-07-14
[VIM] Auto import class

dantleech:
This is an POC PR for automatically importing completed classes.

recording

The mechanism works fine, but:

- The cursor jumps because of the replaced text in the current file, as seen in the screenshot above this is not good.

It might also be more explicit to add a confirmation step?

In anycase, this could be merged as a disabled-by-default feature.


plugin/phpactor.vim
```diff
@@ -9,6 +9,9 @@
 let g:phpactorpath = expand('<sfile>:p:h') . '/..'
 let g:phpactorbinpath = g:phpactorpath. '/bin/phpactor'
 let g:phpactorInitialCwd = getcwd()
+let g:_phpactor_completion_meta = {}
+
+autocmd CompleteDone *.php call phpactor#_completeImportClass(v:completed_item)
```
dantleech: This should be an option

plugin/phpactor.vim
```diff
@@ -9,6 +9,9 @@
 let g:phpactorpath = expand('<sfile>:p:h') . '/..'
 let g:phpactorbinpath = g:phpactorpath. '/bin/phpactor'
 let g:phpactorInitialCwd = getcwd()
+let g:_phpactor_completion_meta = {}
```
dantleech: Nice selection of coding styles here.

plugin/phpactor.vim
```diff
@@ -70,9 +73,12 @@ function! phpactor#Complete(findstart, base)
     let issues = result['issues']
 
     let completions = []
+    let g:_phpactor_completion_meta = {}
+
     if !empty(suggestions)
         for suggestion in suggestions
             call add(completions, { 'word': suggestion['name'], 'menu': suggestion['info'], 'kind': suggestion['type']})
+            let g:_phpactor_completion_meta[suggestion['name']] = suggestion
```
dantleech: The `name` key is non-unique.

lib/Extension/Rpc/Diff/Differ.php
```diff
@@ -0,0 +1,61 @@
...
+                    break;
+
+                // Removed
+                case 2:
+                    $edits[] = [
+                        'start' => [
+                            'line' => $lineNumber,
+                            'character' => 0,
+                        ],
+                        'length' => strlen($token),
```
dantleech: this is characters, use `mb_strlen` (and add multi-bute test)

lib/Extension/Rpc/Diff/Differ.php
```diff
@@ -0,0 +1,61 @@
...
+     * @var BergmannDiffer
+     */
+    private $differ;
+
+    public function __construct(BergmannDiffer $differ = null)
+    {
+        $this->differ = $differ ?: new BergmannDiffer();
+    }
+
+    public function chunkDiff(string $original, string $new)
```
dantleech: This class is about inferring text edits, do some renaming.

lib/Extension/Rpc/Response/UpdateFileSourceResponse.php
```diff
@@ -0,0 +1,71 @@
...
+     * @var Differ
+     */
+    private $differ;
+
+    private function __construct(string $path, string $originalSource, string $replacementSource)
+    {
+        $this->replacementSource = $replacementSource;
+        $this->path = $path;
+        $this->originalSource = $originalSource;
+        $this->differ = new Differ();
```
dantleech: Tidy up

lib/Extension/Rpc/Response/UpdateFileSourceResponse.php
```diff
@@ -0,0 +1,71 @@
...
+    {
+        return 'update_file_source';
+    }
+
+    public function parameters(): array
+    {
+        return [
+            'path' => $this->path,
+            'source' => $this->replacementSource,
+            'edits' => $this->differ->chunkDiff($this->originalSource, $this->replacementSource),
```
dantleech: Use `oldSource` / `newSource` instead of `origina/replacement`

plugin/phpactor.vim
```diff
@@ -469,6 +500,51 @@ function! phpactor#_rpc_dispatch(actionName, parameters)
...
+                else
+                    " insert characters after the current line
+                    let appendLines = split(edit['text'], "\n")
+                    call append(line('.'), appendLines)
+                    let newLines = newLines + len(appendLines)
+                endif
+
+                if postCursorPosition[1] > start['line']
+                    let postCursorPosition[1] = postCursorPosition[1] + newLines
+                endif
```
dantleech: Extract this to a method

plugin/phpactor.vim
```diff
@@ -469,6 +500,51 @@ function! phpactor#_rpc_dispatch(actionName, parameters)
...
+                    call append(line('.'), '')
+                    let newLines = 1
+                else
+                    " insert characters after the current line
+                    let appendLines = split(edit['text'], "\n")
+                    call append(line('.'), appendLines)
+                    let newLines = newLines + len(appendLines)
+                endif
+
+                if postCursorPosition[1] > start['line']
```
dantleech: Not sure this is exactly correct

plugin/phpactor.vim
```diff
@@ -469,6 +500,51 @@ function! phpactor#_rpc_dispatch(actionName, parameters)
         return
     endif
 
+    " >> update file source
+    "
+    " NOTE: This method currently works on a line-by-line basis as currently
+    "       supported by Phpactor.
```
dantleech: Expand on this explanation - the Diff implementation we use (from PHPUnit) only works on a line-by-line basis, so we are always removing and adding lines, this means that same-line edits might leave the cursor at the wrong position.

plugin/phpactor.vim
```diff
@@ -469,6 +500,51 @@ function! phpactor#_rpc_dispatch(actionName, parameters)
...
+        if !empty(a:parameters['edits'])
+            for edit in a:parameters['edits']
+
+                let newLines = 0
+
+                " start = { line: 1234, character: 0 }
+                let start = edit['start']
+
+                " move the cursor into the start position
+                call setpos('.', [ 0, start['line'], start['character'] + 1 ])
```
dantleech: `character` is actually not used in theory, should always be 0

plugin/phpactor.vim
```diff
@@ -469,6 +500,51 @@ function! phpactor#_rpc_dispatch(actionName, parameters)
...
     endif
 
+    " >> update file source
+    "
+    " NOTE: This method currently works on a line-by-line basis as currently
+    "       supported by Phpactor.
+    "
+    if a:actionName == "update_file_source"
+        let postCursorPosition = getpos('.')
+        if !empty(a:parameters['edits'])
```
dantleech: we need to switch to or open the buffer here.

plugin/phpactor.vim
dantleech: Fixed

lib/Extension/Rpc/Response/ReplaceFileSourceResponse.php
```diff
@@ -16,15 +17,15 @@ class ReplaceFileSourceResponse implements Response
...
 
-    private function __construct(string $path, string $replacementSource)
+    private function __construct(string $path, string $originalSource, string $replacementSource)
     {
         $this->replacementSource = $replacementSource;
         $this->path = $path;
     }
 
-    public static function fromPathAndSource(string $path, string $replacementSource)
+    public static function fromPathOldAndNewSource(string $path, string $originalSource, string $replacementSource)
```
dantleech: No need for original source here

lib/Extension/Rpc/Response/UpdateFileSourceResponse.php
```diff
@@ -0,0 +1,73 @@
...
+        return [
+            'path' => $this->path,
+            'source' => $this->replacementSource,
+            'edits' => $this->textEditBuilder->calculateTextEdits($this->originalSource, $this->replacementSource),
+        ];
+    }
+
+    public function replacementSource(): string
+    {
+        return $this->replacementSource;
```
dantleech: s/replacement/new

lib/RandomClass.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+namespace Phpactor;
+
+class RandomClass
+{
+}
```
dantleech: Rm

plugin/phpactor.vim
```diff
@@ -22,6 +23,14 @@ if !exists('g:phpactorOmniError')
     let g:phpactorOmniError = v:false
 endif
 
+if !exists('g:phpactorAutoClassImport')
+    let g:phpactorAutoClassImport = v:true
+endif
+
+if g:phpactorAutoClassImport == v:true
+    autocmd CompleteDone *.php call phpactor#_completeImportClass(v:completed_item)
+endif
```
dantleech: Add documentation

plugin/phpactor.vim
```diff
@@ -70,9 +79,18 @@ function! phpactor#Complete(findstart, base)
...
+    let g:_phpactorCompletionMeta = {}
+
     if !empty(suggestions)
         for suggestion in suggestions
-            call add(completions, { 'word': suggestion['name'], 'menu': suggestion['info'], 'kind': suggestion['type']})
+            let completion = { 
+                        \ 'word': suggestion['name'], 
+                        \ 'menu': suggestion['short_description'],
+                        \ 'kind': suggestion['type'],
+                        \ 'dup': 1
```
dantleech: Could add this as a bug fix - it allows duplicate name suggestions (e.g. Symfony `Request` and Phpactor `Request`)

plugin/phpactor.vim
```diff
@@ -160,9 +208,17 @@ function! phpactor#OffsetTypeInfo()
...
+
     let currentPath = expand('%:p')
-    call phpactor#rpc("transform", { "path": currentPath, "source": phpactor#_source() })
+    let args = { "path": currentPath, "source": phpactor#_source() }
+
+    if transform != ''
+        let args.transform = transform
+    endif
+
+    call phpactor#rpc("transform", args)
```
dantleech: ADd changelog entry for regression test here

plugin/phpactor.vim
```diff
@@ -242,6 +303,54 @@ function! phpactor#_selectionEnd()
...
+        else
+            " insert characters after the current line
+            let appendLines = split(edit['text'], "\n")
+            call append(start['line'], appendLines)
+            let newLines = newLines + len(appendLines)
+        endif
+
+        if postCursorPosition[1] > start['line']
+            let postCursorPosition[1] = postCursorPosition[1] + newLines
+        endif
```
dantleech: This seems to work fine - but couldn't figure out how to do this "properly" in VIM script, doesn't matter for the naieve implementation currently supported.

PR #555, opened 2018-08-08
Added hint to help determine cause of bad diagnostic

wjzijderveld:
It didn't recognize composer on my (freshly check out) project.

After some digging I found it actually tried to find `vendor/autoload.php` instead of `composer.json` or something, and I hadn't yet run composer install :hankey: 

With this extra message it might help people understand better why detection might fail. Although maybe it should even be more specific about `vendor/autoload.php`?


dantleech:
Heh, I think this message is OK, I guess it's something of an edge case and Phpactor should work despite not finding composer (and be pretty fast as there is no vendor director to scan :)

Issue #558, opened 2018-08-11
[vim plugin] Add an ability to determine a namespace from the file name


voronkovich:
This feature would be very useful in the snippets (Ultisnips and etc.). For example:
```
<?php

namespace `!v phpactor#getFileNamespace()`;

class Foo
{
}
```

PR #559, opened 2018-08-11
Determine namespace from file path

voronkovich:
Fix https://github.com/phpactor/phpactor/issues/558


dantleech:
This looks perfect thanks.

Would it be possible to add a [unit test case](https://github.com/phpactor/phpactor/blob/develop/tests/Unit/Extension/SourceCodeFilesystem/Rpc/ClassSearchHandlerTest.php) for the handler and bonus points if you can add a simple [VIM regression test](https://github.com/phpactor/phpactor/blob/develop/plugin/tests/rpc_file_references.vader)?

voronkovich:
@dantleech, I don't know what to test in Vim. I've added unit and system tests.

lib/Extension/ClassToFile/Application/FileInfo.php
```diff
@@ -6,7 +6,7 @@
 use Phpactor\ClassFileConverter\Domain\FilePath;
 use Phpactor\ClassFileConverter\Domain\FileToClass;
 
-final class FileInfo
+class FileInfo
```
voronkovich: Prophecy doesn't know how to test final classes.

tests/System/Extension/ClassToFile/Rpc/FileInfoTest.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+namespace Phpactor\Tests\System\Extension\ClassToFile\Rpc;
+
+use Phpactor\Tests\System\SystemTestCase;
+
+class FileInfoTest extends SystemTestCase
```
dantleech: This one is not necessary I think, at least there is no precedent for RPC system tests at the moment, so reluctant to duplicate effort here.

The test I suggested was for the VIM plugin, which would have been an E2E test. These are few and are kept in `plugin/tests/*`, this one could look like:

```
# plugin/tests/get_namespace.vader
Execute:
    execute "edit src/Phpactor.php"
    namespace = call phpactor#GetNamespace()
    AssertEqual "Phpactor", namespace
```

Like I said, you would need the [vader](https://github.com/junegunn/vader.vim) plugin to execute this (with `:Vader` from within VIM with that file open).

But, if you don't want to do this, I can put it in later. No worries.

tests/Unit/Extension/ClassToFile/Rpc/FileInfoHandlerTest.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+namespace Phpactor\Tests\Unit\Extension\ClassToFile\Rpc;
+
+use Phpactor\Extension\ClassToFile\Application\FileInfo;
+use Phpactor\Extension\ClassToFile\Rpc\FileInfoHandler;
+use Phpactor\Extension\Rpc\Response\ReturnResponse;
+use PHPUnit\Framework\TestCase;
+
+class FileInfoHandlerTest extends TestCase
```
dantleech: This shoud extend `HandlerTestCase`, and would look something like:

```php
class FileInfoHandlerTest extends HandlerTestCase
{
    private $fileInfo;

    public function setUp()
    {
        $this->fileInfo = $this->prophesize(FileInfo::class);
    }

    protected function createHandler(): Handler
    {
        return new FileInfoHandler($this->fileInfo->reveal());
    }

    public function testReturnsAResponseWithAFileInfo()
    {
        $result = [
            'class' => 'App\Controller\BlogController',
            'class_name' => 'BlogController',
            'class_namespace' => 'App\Controller',
        ];

        $this->fileInfo->infoForFile($path)->willReturn($result);

        $response = $this->handle('file_info', [
            'path' => self::EXAMPLE_PATH
        ]);

        $this->assertInstanceOf(ReturnResponse::class, $response);
        $this->assertEquals($result, $response->parameters()['value']);
    }
}


```

tests/System/Extension/ClassToFile/Rpc/FileInfoTest.php
```diff
@@ -0,0 +1,40 @@
...
+        $response = json_decode($process->getOutput(), true);
+
+        $this->assertSuccess($process);
+        $this->assertEquals([
+            'action' => 'return',
+            'parameters' => [
+                'value' => [
+                    'class' => 'Animals\Badger\Carnivorous',
+                    'class_name' => 'Carnivorous',
+                    'class_namespace' => 'Animals\Badger',
```
dantleech: Would it be worth adding `GetClassFullName` and `GetClassShortName` VIM functions while we are here?

voronkovich:
@dantleech, I've fixed all issues you pointed out. Also I added the `GetClassFullName` method as you suggested. Such method could be helpful when it's needed to detect an alternate file (tests, form type and etc.). To get just a class name it's better to use `expand('%:t:r')`.

dantleech:
Great, thanks for your efforts :) I'll add this to the CHANGELOG file.

PR #587, opened 2018-09-01
Force reload file

### LINKED ISSUE #582: [develop] Regression rename method (in open file), file is modified on disk

dantleech:
Fixes #582


dantleech:
@kermorgant note that the `force_reload` flag should probably be applied to the emacs plugin

Issue #476, opened 2018-05-17
Composer and Git not detected


Tiriel:
Hi there,

Sorry to bother you, but I got some wee problem.
As said in the title, neither Git nor Composer are detected by phpactor (which is damn weird since I had to use both to install phpactor), thus preventing me from using most of the features I like.

The stack:
* OS: Ubuntu 18.04
* Installation type: as Vim plugin, with symlink to /usr/local/bin
* Vim version: 8
* `:call phpactor#Status()`
```
Support                                                                                                                                                                                                                                       
-------                                                                                                                                                                                                                                       
[✔] XDebug is disabled. XDebug has a negative effect on performance.                                                                                                                                                                          
[✘] Composer not found - some functionality will not be available (e.g. class creation) and class location will fallback to scanning the filesystem - this can be slow.                                                                       
[✘] Git not detected. Some operations which would have been better scoped to your project repository will now include vendor paths.                                                                                                           
Config files                                                                                                                                                                                                                                  
------------                                                                                                                                                                                                                                  
[✘] /etc/xdg/phpactor/phpactor.yml                                                                                                                                                                                                            
[✘] /usr/share/upstart/xdg/phpactor/phpactor.yml                                                                                                                                                                                              
[✘] /etc/xdg/xdg-gnome/phpactor/phpactor.yml                                                                                                                                                                                                  
[✘] /home/[anonymized]/.config/phpactor/phpactor.yml                                                                                                                                                                                             
[✘] /home/[anonymized]/Dev/php/[anonymized]/.phpactor.yml  
```

I've looked and looked again, but I can't find any doc on how to tell phpactor where to find Git and Composer.

Do I need to add a config file to tell phpactor where to find the bins? I didn't find any doc on the subject either.

Thanks!

kermorgant:
Hi,

Emacs user here so only thing I can say is I had this issue when invoking phpactor from outside of project root directory.

weirdan:
AFAIK it's not about the binaries. It's about phpactor being unable to find *your project's* .git folder and composer.json

Tiriel:
Thanks for the answers!

That's strange, because I had CDed to my project's root directory. Although I might have done that from *inside* vim. I'll check this asap (it occurs only on my work computer). Thanks again!

dantleech:
The Phpactor VIM plugin assumes the direcrory in which you invoked VIM is your proect root, so it won't matter if you change directories within VIM.

[...]

PR #596, opened 2018-09-12
fix issues on updating cwd

Mte90:
Actually if the cwd is changed inside vim, phpactor doesn't care of that.
This fix the issue.

### DIFF
--- a/plugin/phpactor.vim
+++ b/plugin/phpactor.vim
@@ -374,7 +374,7 @@ function! phpactor#rpc(action, arguments)
 
     let request = { "action": a:action, "parameters": a:arguments }
 
-    let cmd = g:phpactorPhpBin . ' ' . g:phpactorbinpath . ' rpc --working-dir=' . g:phpactorInitialCwd
+    let cmd = g:phpactorPhpBin . ' ' . g:phpactorbinpath . ' rpc --working-dir=' . getcwd()
     let result = system(cmd, json_encode(request))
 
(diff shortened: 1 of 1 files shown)


dantleech:
This is by design. The current assumption (and I appreciate this isn't what everybody does) is that you open VIM for each project. The change you are reverting here means that Phpactor will not _stop_ working if you change directory to, for example, a submodule or a vendor directory in your project (see [here](https://github.com/phpactor/phpactor/issues/476#issuecomment-390026247) and [here](https://github.com/phpactor/phpactor/issues/476#issuecomment-390026247)

The solution would be to automatically detect the root directory of the project, but this is not as easy as it sounds (lots of edge cases), but could be implemented.

dantleech:
But anyway, thanks for the PR but closing this now as this just reverts a fix for other issues. Feel free to create an issue for detecting the project root as I don't think there is one currently.

Mte90:
maybe a solution can be create a command that update the working directory, I am using startify so I use Vim as an IDE to start the project that I need.

dantleech:
yeah I don't think there is anything stopping you updating the `g:phpactorInitialCwd` variable in VIM.

In `plugin/phpactor.vim` it is set as follows when the plugin is loaded:

```
let g:phpactorInitialCwd = getcwd()
```

dantleech:
Detecting the project root is certainly a solvable problem though (lots of people have requested it), would be good to implement it in Phpactor.

PR #609, opened 2018-10-07
Completion generators

dantleech:
The completion `Completor` instances now return generators of `Suggestion` instances. This means it can stream results to clients rather than blocking until all completions can be resolved.

Unfortunately there was no simple way to maintain the list of issues encountred when building the completion list, so this feature has been removed. We could re-introduce the feature if it seems that it is missed, but linters should do this job anyway.

### DIFF
--- a/lib/Test.php
+++ b/lib/Test.php
@@ -1,6 +1,0 @@
-<?php
-
-namespace Phpactor;
-
-
-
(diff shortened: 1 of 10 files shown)


lib/Test.php
```diff
@@ -0,0 +1,6 @@
+<?php
+
+namespace Phpactor;
+
+
+
```
dantleech: rm

PR #610, opened 2018-10-07
[server] publish diagnostics after completion

dantleech:
- disables the reflection cache when running the language server
- publishes diagnostics after providing completion suggestions

### DIFF
--- a/lib/Extension/LanguageServer/LanguageServerExtension.php
+++ b/lib/Extension/LanguageServer/LanguageServerExtension.php
@@ -7,8 +7,11 @@ use Phpactor\Container\ContainerBuilder;
 use Phpactor\Container\Extension;
 use Phpactor\Extension\LanguageServer\Command\StartCommand;
 use Phpactor\Extension\Completion\LanguageServer\CompletionHandler;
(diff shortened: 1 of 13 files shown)


lib/Extension/LanguageServer/LanguageServerExtension.php
```diff
@@ -49,5 +60,9 @@ public function load(ContainerBuilder $container)
         $container->register('language_server.session_manager', function (Container $container) {
             return new Manager();
         });
+
+        $container->register('language_server.extension.code', function (Container $container) {
```
dantleech: s/code/core

lib/Phpactor.php
```diff
@@ -35,6 +35,7 @@ public static function boot(InputInterface $input, string $vendorDir): PhpactorC
         if ($input->hasParameterOption([ '--working-dir', '-d' ])) {
             $config[CoreExtension::WORKING_DIRECTORY] = $cwd = $input->getParameterOption([ '--working-dir', '-d' ]);
         }
+        $config['command'] = $input->getFirstArgument();
```
dantleech: use constnat

PR #619, opened 2018-10-26
Hover

dantleech:
Introduces the "hover" functionality which shows some simple information about the symbol under the cursor.

recording

... also uses external container package and updates composer.


lib/Extension/Completion/Rpc/HoverHandler.php
```diff
@@ -0,0 +1,140 @@
...
+                return $this->renderMember($symbolContext);
+            case Symbol::CLASS_:
+                return $this->renderClass($symbolContext->type());
+            case Symbol::FUNCTION:
+                return $this->renderFunction($symbolContext);
+            case Symbol::VARIABLE:
+                return $this->renderVariable($symbolContext);
+        }
+
+        return null;
```
dantleech: add message here

Issue #632, opened 2018-11-07
"The command argument does not exist" unless $pwd = dirname($phpactor_bin)


weirdan:
phpactor fails when it's run from some random folder (and works when it's run from its own root project folder), with two distinct errors:

# No service with ID "monolog.logger" exists

Happens when phpactor is run without arguments.

Output:
```console
$ phpactor/bin/phpactor -vvv
Phpactor dev-develop@d71f2cbb6a5c679d501e9b6e000672afe150f293

Usage:
  command [options] [arguments]

Options:
  -h, --help                     Display this help message
  -q, --quiet                    Do not output any message
  -V, --version                  Display this application version
      --ansi                     Force ANSI output
      --no-ansi                  Disable ANSI output
  -n, --no-interaction           Do not ask any interactive question
  -d, --working-dir=WORKING-DIR  Working directory
  -v|vv|vvv, --verbose           Increase the verbosity of messages: 1 for normal output, 2 for more verbose output and 3 for debug


In PhpactorContainer.php line 68:
                                              
  [RuntimeException]                          
  No service with ID "monolog.logger" exists  
                                              

Exception trace:
 Phpactor\Container\PhpactorContainer->get() at /home/weirdan/.config/nvim/bundle/phpactor/vendor/phpactor/source-code-filesystem-extension/lib/SourceCodeFilesystemExtension.php:50
 Phpactor\Extension\SourceCodeFilesystem\SourceCodeFilesystemExtension->Phpactor\Extension\SourceCodeFilesystem\{closure}() at /home/weirdan/.config/nvim/bundle/phpactor/vendor/phpactor/container/lib/PhpactorContainer.php:74
 Phpactor\Container\PhpactorContainer->get() at /home/weirdan/.config/nvim/bundle/phpactor/lib/Extension/Core/CoreExtension.php:125
 Phpactor\Extension\Core\CoreExtension->Phpactor\Extension\Core\{closure}() at /home/weirdan/.config/nvim/bundle/phpactor/vendor/phpactor/container/lib/PhpactorContainer.php:74
 Phpactor\Container\PhpactorContainer->get() at /home/weirdan/.config/nvim/bundle/phpactor/lib/Extension/Core/CoreExtension.php:78
 Phpactor\Extension\Core\CoreExtension->Phpactor\Extension\Core\{closure}() at /home/weirdan/.config/nvim/bundle/phpactor/vendor/phpactor/container/lib/PhpactorContainer.php:74
[...]

PR #633, opened 2018-11-07
Don't fail error handling on unconfigured input

weirdan:
Prevents 'The "command" argument does not exist' error (see #632) when exception happened before input has been configured.


dantleech:
Cheers :)

Issue #507, opened 2018-06-20
[VIM] Unknown function: phpactor#Status()


YaroslavMolchan:
Hello, I try to install phpactor as vim plugin using this instructions: https://phpactor.github.io/phpactor/vim-plugin.html

I also installed it with vim-plug, composer install works good, but when I try to check status or other function got same error that function not found.

![image](https://user-images.githubusercontent.com/9335727/41655369-5315dc94-7495-11e8-9491-59ee576d57e4.png)

I used gVIM on Ubuntu 18.04, my VIM version:

> VIM - Vi IMproved 8.0 (2016 Sep 12, compiled Apr 10 2018 21:31:58)
> Included patches: 1-1453

dantleech:
When using VIM Plug you will need to open a PHP file before Phpactor is registered.

dantleech:
@YaroslavMolchan can you confirm?

YaroslavMolchan:
@dantleech yes, that was a problem, sorry for late reply, forgot about it.

PR #635, opened 2018-11-11
Add troubleshooting to vim-plugin doc

dshoreman:
I've been bitten by #507 on more than one occasion: setting up Phpactor yields the dreaded **E117: Unknown function: phpactor#Status** as I'm attempting to do it from my vim config and not a PHP file.

This adds some text to the doc to hopefully workaround the confusion for new users or those like me with the memory of a goldfish.

### DIFF
--- a/doc/vim-plugin.md
+++ b/doc/vim-plugin.md
@@ -73,8 +73,15 @@ auto-completion can still work (sometimes slowly depending on project size).
 
 **E117: Unknown function: phpactor#Status**:
 
-When using vim-plug, you will need to open a PHP file before Phpactor is registered.
-For best results, `cd` to a project that uses Git and Composer before starting Vim.
+Vim-plug and most other package managers for Vim will lazy-load Phpactor when it's
+needed, i.e. when opening a PHP file. If you get this error, open a PHP file and
+run the command again.
+
+**Composer not found** or **Git not detected**:
+
+The Git and Composer checks are referring to the current "workspace" (i.e. where you
+started Vim from). If you've already setup Git and Composer for your project, ensure
+you are starting Vim from the project directory to enable detection.
 
 Updating
 --------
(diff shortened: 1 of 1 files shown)


doc/vim-plugin.md
```diff
@@ -69,6 +69,13 @@ Config files
 Phpactor works best with Composer - but much functionality including
 auto-completion can still work (sometimes slowly depending on project size).
 
+### Troubleshooting
+
+**E117: Unknown function: phpactor#Status**:
+
+When using vim-plug, you will need to open a PHP file before Phpactor is registered.
+For best results, `cd` to a project that uses Git and Composer before starting Vim.
```
dantleech: I guess this is two peices of advice? i.e. one for the error, and the other is for "best results".

What about:

```
**E117: Unknown function: phpactor#Status**:

The `vim-plug` (and similar) package managers will lazily load plugins only when they are relevant (i.e. when you open a PHP file). if you receive the above error, then you should open a PHP file in VIM before issuing the command.
```

Not sure if we already have a hint about starting VIM in the project root?

dantleech: Or maybe it should be mentioned before `Now issue the following command ...` above? As we are assuming already that the user is using `vim-plug`?

dshoreman: @dantleech Thanks for the feedback.

I've reworded the error explanation as you suggested. I'll be out for the rest of the day but let me know your thoughts and I can make any changes (especially where composer/git are concerned) later on.

#### Composer/git bits

I've moved this to its own subheader to keep it separate from the actual 'E117' error. I think it could still be useful though, so rather than remove it entirely I've reworded to be less of a "tip" and more of an explanation of *why* it couldn't be detected.

While the readme and Standalone pages say *"Phpactor works better with Git and Composer"*, I couldn't find anywhere that specifically mentions starting from the project directory.

dantleech:
Thanks!

PR #638, opened 2018-11-12
include symfony debug as dependency

dantleech:
It was noticed that Phpactor didn't run with `composer install --no-dev` as the main `bin/phpactor` file enables the symfony Debug error handler.

Having this enabled is useful in development (ensuring that errors are always presented as exceptions), and presents no security risk or performance penalty in the context of the tool that I'm aware of.

So adding as a regular dependency for now.


Taluu:
Instead you could maybe add a `if (class_exists(Debug::class))` ?

dantleech:
Including it in #639 anyway

PR #654, opened 2018-12-01
Lazy handlers

dantleech:
Halfs the baseline latency for RPC calls.


bin/phpactor
```diff
@@ -34,10 +34,6 @@ if (version_compare(PHP_VERSION, $minVersion) < 0) {
 
 require_once $autoloadFile;
 
-if (class_exists(Debug::class)) {
-    Debug::enable();
-}
-
```
dantleech: this actually causes a performance issue.

PR #668, opened 2018-12-12
Define application root path according to location of vendor folder

kermorgant:
Application path is defined differently when phpactor is installed through composer (as a required package)

### DIFF
--- a/lib/Phpactor.php
+++ b/lib/Phpactor.php
@@ -51,7 +51,7 @@ class Phpactor
 
         $config = $loader->load();
         $config[CoreExtension::COMMAND] = $input->getFirstArgument();
-        $config[FilePathResolverExtension::PARAM_APPLICATION_ROOT] = realpath(__DIR__ . '/..');
+        $config[FilePathResolverExtension::PARAM_APPLICATION_ROOT] = self::resolveApplicationRoot();
         $config = self::configureExtensionManager($config, $vendorDir);
 
         if ($input->hasParameterOption([ '--working-dir', '-d' ])) {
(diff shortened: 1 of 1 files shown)


lib/Phpactor.php
```diff
@@ -51,7 +51,13 @@ public static function boot(InputInterface $input, string $vendorDir): PhpactorC
 
         $config = $loader->load();
         $config[CoreExtension::COMMAND] = $input->getFirstArgument();
-        $config[FilePathResolverExtension::PARAM_APPLICATION_ROOT] = realpath(__DIR__ . '/..');
+
+        if (is_dir(realpath(__DIR__ . '../vendor'))) {
+            $config[FilePathResolverExtension::PARAM_APPLICATION_ROOT] = realpath(__DIR__ . '/..');
+        } else {
+            $config[FilePathResolverExtension::PARAM_APPLICATION_ROOT] = realpath(__DIR__ . '/../../../..');
+        }
```
dantleech: can we extract to a method and remove the `else`?

maybe even something like the following?

```
private function resolveApplicationRoot() 
{
   $paths = [ __DIR__ . '/../vendor', __DIR__ .'/../../../..' ];
   foreach ($paths as $path) { if (is_dir($path)) { return $path; } }
   throw new RuntimeException(sprintf('Could not resolve application root, tried "%s"', implode('", "', $paths)));
   }
}
```

kermorgant: sure ! (I kept the call to `realpath`)

dantleech:
Thanks :+1:

PR #727, opened 2019-01-29
Update Slack invite link

dshoreman:
The Slack invite link in README had expired. Updates to use one that redirects which should avoid future expiry.


dantleech:
I didn't realise it would expire, thanks :)

PR #729, opened 2019-01-30
Allow case insensitive search in vim autocompletion

rushebshah:
Introduces global variable g:phpactorCompletionIgnoreCase (0 by default),
which if set to 1, allows case insensitive search for completion.


dantleech:
I'd think we could maket this `1` by default?

dantleech:
I'd think we could maket this `1` by default?

rushebshah:
@dantleech sounds good, I've updated the default to be `1`.

PR #734, opened 2019-02-03
updated composer.lock

dantleech:
```
dantleech/what-changed: 12 updated
  ...

  jetbrains/phpstorm-stubs f66ca5ff..d1ecf642

    [2019-01-16 15:08:44] 7b58b73c tvlooy Correct session_gc signature
    [2019-01-28 10:26:32] 071ccb67 <???> More expectedArguments for standard functions/methods
    [2019-01-28 10:30:29] a53c084b <???> Use singe quotes for consistency
    [2019-01-28 10:34:21] 8948c521 <???> expectedArguments for HTML encodings
    [2019-01-28 13:35:48] 34ae6d56 wbars WI-45131 Introduce registerArgumentsSet and argumentsSet to reuse expected arguments between function refere...
    [2019-01-28 13:36:34] 09303a57 wbars Expected arguments for ini_* functions
    [2019-01-28 13:37:24] 035672ec wbars cleanup
    [2019-01-28 13:39:15] d30aa8e3 wbars More sane logic
    [2019-01-28 14:05:35] ed07a902 wbars WI-45131 Check registerArgumentsSet in tests as well
    [2019-01-28 14:10:39] 7df9650b wbars Use single quotes
    [2019-01-28 14:48:10] 582f6965 wbars expectedArguments for get_cfg_var
    [2019-01-28 14:48:55] be0dae10 wbars Update PHPDoc
    [2019-01-30 10:39:22] 5f17166b pronskiy Add expected args for \json_* functions
    [2019-01-30 14:46:21] c507ea29 Girgias Fixed array_push and array_unshift stub as of PHP 7.3

  phpactor/code-builder 4269d94e..63c33352

    [2019-01-28 08:58:31] 9fe8fe21 dshoreman Copy from Class builder/prototypes
    [2019-01-28 09:02:15] 2d9213d3 dshoreman Update class names, drop extends and implements
    [2019-01-28 10:52:26] 6474ca43 dshoreman Add trait handling to source builder
    [2019-01-28 10:53:06] 84c78885 dshoreman Allow PropertyBuilder to take TraitBuilder  This also inadvertently enables passing an InterfaceBuilder....
    [2019-01-28 11:40:41] 2cfc43c1 dshoreman Add Traits Collection
    [2019-01-28 14:05:13] 7f474ee6 dshoreman Duplicate necessary parts of ClassUpdater
    [2019-01-28 14:06:11] 190c1130 dshoreman Use Trait Prototype, Declaration and members
[...]

Issue #726, opened 2019-01-29
Add Missing Properties transformation not working as expected in Traits


dshoreman:
## Summary
When I `:call phpactor#Transform("add_missing_properties")` (or hit `a` on the Transform prompt) while in a Trait, it adds a new class. When performing the same action from a class, it works as expected.

### Expected Behaviour
Phpactor adds a new private property to the trait.

### Actual Behaviour
It creates an entirely new class matching the name of the trait and adds the property there instead.

## Example
![phpactor-traitbug](https://user-images.githubusercontent.com/1521802/51891814-25f00980-2398-11e9-92ae-11e2c7459e7a.gif)

dshoreman:
After a few hours yesterday digging around, I was able to track the issue down to the code-transform repo and in turn found code-builder was lacking Trait support.

I've been able to fix locally so I'll open PRs on the respective repositories soon. While I've added some tests, it's still a work in progress and could likely use some feedback on potential refactoring and extra tests.

With everything split between at least two repos, hopefully this issue will help to keep everything tied together in one place.

Side note: I tried to join the Slack linked at the bottom of the README before opening this, but the invite link has expired. Is it still in use?

dantleech:
Great, feel free to ping me on Slack :)

PR #735, opened 2019-02-04
Update to include latest code-transform

dshoreman:
Last night's `composer.lock` commit (#734) brought `code-builder` up to date but missed the PR merge in `code-transform`. Without that, #726 is still an issue so this pulls in the latest changes to resolve it.


dantleech:
Ah, must have made the release before composer updated, thanks!

PR #741, opened 2019-02-24
Add PHP open tag in template example

Einenlum:
Hi Dan (and other contributors): thanks for this awesome tool :pray: 

This PR just fixes a thing in the [templates documentation](https://phpactor.github.io/phpactor/templates.html):
The example could let the user think that phpactor will automatically prepend `<?php\n` to the file, but it is not.


dantleech:
Thanks, your welcome :)

PR #769, opened 2019-05-30
Adds fzf support for list input

elythyr:
There is the PR.

After checking there is already a fallback to the standard vim `inputlist()`.
It's in the `autoload/phpactor/input/list.vim` file.
It's possible to define a strategy for the inputlist with the global variable `phpactor#input#list#strategy`.
The value must be a valid function inside the namespace `phpactor#input#list`.
The corresponding function will be used.

If the global variable does not exist, then the strategy is automatically detected beetween fzf and vim's `inputlist()` depending on is the fzf plugin is loaded or not.

### DIFF
--- a/autoload/phpactor/input/list.vim
+++ b/autoload/phpactor/input/list.vim
@@ -1,0 +1,40 @@
+function! phpactor#input#list#strategy()
+    if !has_key(g:, 'phpactor#input#list#strategy')
+        let g:phpactor#input#list#strategy = s:auto_detect_strategy()
+    endif
+
+    return g:phpactor#input#list#strategy
+endfunction
+
+function! phpactor#input#list#inputlist(label, choices, ResultHandler)
+    echo a:label
+    let choice = inputlist(s:add_number_to_choices(a:choices))
+
+    if (choice == 0)
+        throw "cancelled"
+    endif
+
+    call a:ResultHandler(a:choices[choice - 1])
+endfunction
+
+function! phpactor#input#list#fzf(label, choices, ResultHandler)
+    " sink works because "key" is converted to integer, so only the number is kept
+    call fzf#run({
+        \ 'source': s:add_number_to_choices(a:choices),
+        \ 'sink': {key -> a:ResultHandler(a:choices[key - 1])},
+        \ 'down': '30%',
+        \ 'options': ['--tiebreak=index', '--layout=reverse-list']
+    \ })
+endfunction
+
+function! s:auto_detect_strategy()
+    if get(g:, 'loaded_fzf', 0)
+        return 'fzf'
+    endif
+
+    return 'inputlist'
+endfunction
+
+function! s:add_number_to_choices(choices)
+    return map(copy(a:choices), {key, value -> key + 1 .') '. value})
+endfunction
--- a/autoload/phpactor/input/list.vim
+++ b/autoload/phpactor/input/list.vim
@@ -1,9 +1,9 @@
 function! phpactor#input#list#strategy()
-    if !has_key(g:, 'phpactor#input#list#strategy')
-        let g:phpactor#input#list#strategy = s:auto_detect_strategy()
+    if !has_key(g:, 'phpactorInputListStrategy')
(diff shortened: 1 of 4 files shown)


plugin/phpactor.vim
```diff
@@ -708,38 +635,53 @@ function! phpactor#_rpc_dispatch(actionName, parameters)
     throw "Do not know how to handle action '" . a:actionName . "'"
 endfunction
 
-function! phpactor#_rpc_dispatch_input(type, parameters)
-    " Remove any existing output in the message window
-    execute ':redraw'
+function! phpactor#_rpc_dispatch_input_handler(Next, parameters, parameterName, result)
```
dantleech: Should `Next` be lower case?

plugin/phpactor.vim
```diff
@@ -708,38 +635,53 @@ function! phpactor#_rpc_dispatch(actionName, parameters)
...
+            \ inputParameters['label'],
+            \ inputParameters['choices']
+        \ ])
+    elseif 'list' == input['type']
+        let TypeHandler = function('phpactor#input#list', [
+            \ inputParameters['label'],
+            \ inputParameters['choices']
+        \ ])
+    elseif 'confirm' == input['type']
+        let TypeHandler = function('phpactor#input#confirm', [
```
dantleech: Would also prefer to be lower case for consistency

dantleech:
Looks good :) I like delegating to the new files and the anonymous functions. Just the above bug with confirm and the question about CS.

plugin/phpactor.vim
elythyr: If I remember correctly yes.
A public function should start with a letter in upper case.
Exception for autoloaded function where everything is lower case (most of the time).
But I'm not 100% sure.

I will look into the error you mentioned.

elythyr:
That's fixed.
Nasty combination between copy/paste and small memory... Forgot to adapt the code of the confirm() function ><

I also spotted another bug while debugging, I removed the handling of the cancelled exception.
If you open the menu and press Escape you will be able to reproduce it.
That's the first commit.

Do you want me to rebase everything in a single commit since it should have already be done in the first one ?

elythyr:
CS stands for "Case Style" ?
If that's it then:
`:h user-functions` gave:
> The function name must start with an uppercase letter, to avoid confusion with
builtin functions.  To prevent from using the same name in different scripts
avoid obvious, short names.  A good habit is to start the function name with
the name of the script, e.g., "HTMLcolor()".

But to be exact the `Next` is actually a `Funcref`, with `:h E704`:
> A Funcref variable must start with a capital, "s:", "w:", "t:" or "b:".  You
can use "g:" but the following name must still start with a capital.  You
cannot have both a Funcref variable and a function with the same name.

In the case you mentioned I don't know if the rule still applies.
But I find those rules complicated enough so I always use a capital letter to for my Funcref, standards rules us all :)

autoload/phpactor/input/list.vim
```diff
@@ -0,0 +1,40 @@
+function! phpactor#input#list#strategy()
+    if !has_key(g:, 'phpactor#input#list#strategy')
+        let g:phpactor#input#list#strategy = s:auto_detect_strategy()
```
dantleech: can we add this setting to the plugin doc? https://github.com/phpactor/phpactor/blob/develop/doc/vim-plugin.md#configuration 

Would it make more sense to have `phpactorInputListStrategy` as the global name to be consistent (more consistent) with the existing options?

elythyr: I will try to do the doc tomorrow.
For the name, I had the same idea but I wanted to keep it as close as the autoload functions as possible.
But since the configuration is not autoloaded it makes more sense.

elythyr: Done & Done

I also added a custom error message in case of invalid strategy.
And gave the possibility to provide a custom strategy if a user want to create it's own.

dantleech:
Looks good, already using this locally and it's a huge improvement!

We can update the changelog before the next release.

Issue #771, opened 2019-06-07
InputList allow multi selection


elythyr:
Hi,
I think it would be great to be able to select multiple items in a list in some cases.
For example, when we want to override a method we can only choose one.
On the contrary when we want to add a use statement we want to only select one item.

My idea here is to add a parameter to the `ListInput' class which will tell if the list accepts multiple items to be selected.
So in the `OverrideMethodHandler`, for example, we could set it to true.
Then in the Vim plugin we can pass another argument to the list functions to let them know about it.

It will have no impact on the vim `inputlist` but with fzf we can allow multiple selection.
The way fzf works, it will call the sink function for each selected item.
So one RPC call for each item.

Which allow to reuse existing handlers since we wouldn't have to modify them to accept multiple results.

I already have a proposal for the implementation:
https://github.com/elythyr/phpactor/tree/list-multi-selection
https://github.com/elythyr/rpc-extension/tree/list-multi-selection

dantleech:
Yes, this would be really great!

Do you want to make a PR for those changes? We can then take it further (I guess should be simple to support multiple ops in most in the RPC handlers).

elythyr:
The PR is ready on my forks.
For the handlers, I guess it's just making a loop going through the parameters and reusing the new `SourceCode` object every time.

I didn't want to change to many things at first, if you prefer this way I will do it, just let me know.
Plus I don't have a global vision of all the handlers and if it's applicable to all of them.

I will open a PR on this repository and the rpc-extension reposioty, and we can discuss this there.

PR #773, opened 2019-06-08
List multi selection

elythyr:
PR for https://github.com/phpactor/phpactor/issues/771

So we still have to decide if we keep the actual behavour of multiple RPC calls or if we adapt the handlers.

Plus there is two commits that have nothing to do with this PR.
One that adds modlines to the vim files, I don't know how you feel about that.
I personally use 2 spaces for tab in vimscript and it was a pain for me to change it every time I opened a file.
Another one that removes a variable declaration that was forgotten when you push the `target` parameter for the `open_file` action if I remember correctly.

So do you wan those commits in this PR, in a separate one or not at all ?

### DIFF
--- a/lib/Extension/CodeTransformExtra/Rpc/OverrideMethodHandler.php
+++ b/lib/Extension/CodeTransformExtra/Rpc/OverrideMethodHandler.php
@@ -73,7 +73,7 @@ class OverrideMethodHandler extends AbstractHandler
         }
 
         $newCode = $arguments[self::PARAM_SOURCE];
-        foreach ($this->methods($arguments) as $methodName) {
+        foreach ((array) $arguments[self::PARAM_METHOD_NAME] as $methodName) {
             $newCode = $this->overrideMethod->overrideMethod(
                 SourceCode::fromString((string) $newCode),
                 (string) $class->name(),
@@ -134,16 +134,5 @@ class OverrideMethodHandler extends AbstractHandler
         sort($methodNames);
 
         return array_combine($methodNames, $methodNames);
-    }
-
-    private function methods(array $arguments): array
-    {
-        $methods = $arguments[self::PARAM_METHOD_NAME];
-
-        if (is_string($methods)) {
-            return [$methods];
-        }
-
-        return $methods;
     }
 }
--- a/lib/Extension/CodeTransformExtra/Rpc/OverrideMethodHandler.php
+++ b/lib/Extension/CodeTransformExtra/Rpc/OverrideMethodHandler.php
@@ -72,16 +72,19 @@ class OverrideMethodHandler extends AbstractHandler
             return $this->createInputCallback($arguments);
         }
 
-        $transformedCode = $this->overrideMethod->overrideMethod(
-            SourceCode::fromString($arguments[self::PARAM_SOURCE]),
-            (string) $class->name(),
-            $arguments[self::PARAM_METHOD_NAME]
-        );
+        $newCode = $arguments[self::PARAM_SOURCE];
+        foreach ($this->methods($arguments) as $methodName) {
+            $newCode = $this->overrideMethod->overrideMethod(
+                SourceCode::fromString((string) $newCode),
+                (string) $class->name(),
+                $methodName
+            );
+        }
 
         return UpdateFileSourceResponse::fromPathOldAndNewSource(
             $arguments[self::PARAM_PATH],
             $arguments[self::PARAM_SOURCE],
-            (string) $transformedCode
+            (string) $newCode
         );
     }
 
@@ -131,5 +134,16 @@ class OverrideMethodHandler extends AbstractHandler
         sort($methodNames);
 
         return array_combine($methodNames, $methodNames);
+    }
+
+    private function methods(array $arguments): array
+    {
+        $methods = $arguments[self::PARAM_METHOD_NAME];
+
+        if (is_string($methods)) {
+            return [$methods];
+        }
+
+        return $methods;
     }
 }
--- a/lib/Extension/CodeTransformExtra/Rpc/OverrideMethodHandler.php
+++ b/lib/Extension/CodeTransformExtra/Rpc/OverrideMethodHandler.php
@@ -66,7 +66,7 @@ class OverrideMethodHandler extends AbstractHandler
             self::PARAM_METHOD_NAME,
             sprintf('Methods from "%s"', $parentClass->name()),
             $this->methodChoices($parentClass)
-        ));
+        )->setAllowMultipleResults(true));
 
         if ($this->hasMissingArguments($arguments)) {
             return $this->createInputCallback($arguments);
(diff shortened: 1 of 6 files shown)


lib/Extension/CodeTransformExtra/Rpc/OverrideMethodHandler.php
```diff
@@ -66,7 +66,7 @@ public function handle(array $arguments)
             self::PARAM_METHOD_NAME,
             sprintf('Methods from "%s"', $parentClass->name()),
             $this->methodChoices($parentClass)
-        ));
+        )->setAllowMultipleResults(true));
```
dantleech: `set` and `get` are not used in Phpactor. The "responses" are immutable value objects. I think the convention here for an optional would be `withMultiple(true)` which would return a new instance, see https://github.com/phpactor/rpc-extension/blob/master/lib/Response/OpenFileResponse.php#L85

I'd handle optional values differently next time, but this is the way it's done now...

lib/Extension/CodeTransformExtra/Rpc/OverrideMethodHandler.php
```diff
@@ -121,6 +121,7 @@ private function parentClass(ReflectionClass $class)
 
     private function methodChoices(ReflectionClass $parentClass)
     {
+        /** @todo filter methods already implemented in the current class */
```
dantleech: :+1:

dantleech: could _possibly_ add a new API method in worse reflection, e.g. `$methods = $parentMethods->excluding($classMethods->belongingTo($className))`

Otherwise eqaually fine to quickly implement here (not necessarily in this PR)

elythyr:
> Dumb question: How do you load / reload the Phpactor plugin? Not sure I'm doing it correctly...

I struggle with that too... Most of the time I used another tmux window with a dummy php file, so I just close and open Vim again...
Otherwise there is [Scriptease](https://github.com/tpope/vim-scriptease) plugin with `Runtime` command that might work.

lib/Extension/CodeTransformExtra/Rpc/OverrideMethodHandler.php
elythyr: It's not the response that is modified but the `ListInput` instance.
Which is later wrapped into a response with `createInputCallback`.
But if you still prefer an immutable approach I'm fine with that, just let me know.

lib/Extension/CodeTransformExtra/Rpc/OverrideMethodHandler.php
elythyr: Is this the coding style comment you were referencing earlier ?
The `/** @todo` instead of `// TODO ` ?
If yes I will change it, only saw the `// TODO` afterward

dantleech: No, but "//” would be more consistent, I was referring to the setMultipleAllowed method call

lib/Extension/CodeTransformExtra/Rpc/OverrideMethodHandler.php
dantleech: Yes, immutable would be the general approach here, I guess there might be an example? (Away from computer now). But in any case I _think_ what I suggest above would apply

dantleech:
Works great :) I don't think there are currently any other refactorings that can benefit from this now -- but I can imagine changing the f.e. the [generate accessors](https://phpactor.github.io/phpactor/refactorings.html#generate-accessor) to operate based on a list -- would allow to generate accessors for all properties f.e.

Otherwise, I merged the other PR, and just have to wait for a green build here :)

elythyr:
Well the generate accessors based on a list of properties is already done.
I was going to bring the PR once this PR would have been merged.
Because it already use the multi list and it was simpler for me to just wait a bit :)

My turn for a Dumb question: Should I update composer dependencies and push the new composer.lock to trigger Travis ?
Honestly I don't even get how we can have the ~0.2 version when there is only 0.1.x tags on the repository ><

elythyr:
Well this will have to wait, I updated the rpc-extension package and done one last test.
And I realize it's not working correctly.
When the second RPC call is made, it will receive the source without the changes from the first one.
So the code is not updated correctly.

Meaning I will have to change the handlers to accept multiple results and the vim plugin to make only one call with `sink*`

dantleech:
> Should I update composer dependencies and push the new composer.lock to trigger Travis ?

Yep

> I don't even get how we can have the ~0.2 version when there is only 0.1.x tags on the repository

There should be a [branch alias](https://getcomposer.org/doc/articles/aliases.md#branch-alias) in the `phpactor/rpc-extension` `composer.json`, it should be set to `0.2 => dev-master` or similar. This means that, when `minimum-stability: dev` is used in Phpactor's `composer.json`, and ~0,2 is required, then it will always checkout `dev-master`.

> Meaning I will have to change the handlers to accept multiple results and the vim plugin to make only one call with sink*

Ah, I was going to suggest that we iterated over the choices in the handler...

elythyr:
> Ah, I was going to suggest that we iterated over the choices in the handler...

I don't see how to do it.
In `phpactor#_rpc_dispatch_input_handler` we can't really do anything because we don't know if next will make the RPC call or if there is an other input to submit to the user.
Check there if the funcref is the RPC call seems pretty ugly.
And if we have two list inputs with multi enable, how do we handle it then ?
Same thing in `phpactor#rpc` we would have to through all the parameters check which ones are list and make one call per item. With many multi-list it could be challenging.

I think it's ok to change the handles since they will need to be change in order to activate the the multi selection for the input list anyway.
It's just a check for the type of the argument and a loop to add.

lib/Extension/CodeTransformExtra/Rpc/OverrideMethodHandler.php
```diff
@@ -72,16 +72,19 @@ public function handle(array $arguments)
             return $this->createInputCallback($arguments);
         }
 
-        $transformedCode = $this->overrideMethod->overrideMethod(
-            SourceCode::fromString($arguments[self::PARAM_SOURCE]),
-            (string) $class->name(),
-            $arguments[self::PARAM_METHOD_NAME]
-        );
+        $newCode = $arguments[self::PARAM_SOURCE];
+        foreach ($this->methods($arguments) as $methodName) {
```
dantleech: Would prefer to to only pass the `$arguments[self::PARAM_METHOD_NAME]` to make it more clear -- also this could be simplified to simply `foreach ((array) $arguments[self::PARAM_METHOD_NAME]`

elythyr: Never type cast into an array, good to know!

dantleech:
Great work :) I created https://github.com/phpactor/phpactor/pull/774 to add the CHANGELOG and some documentation.

dantleech:
Also announced the change on Twitter, not sure if you have an account:
https://twitter.com/phpactor/status/1137620559179784192

PR #774, opened 2019-06-09
GH-773: Updated changelog and documentation

dantleech:


elythyr:
I should have done the documentation part with my PR sorry.
Before merging it, could you also add in the multi selection part that `CTRL-A` and `CTRL-D` can be used to select/unselect every item in a list ?

PR #775, opened 2019-06-09
Bugfix 770

### LINKED ISSUE #770: :call phpactor#ChangeVisibility() "bug: leaves cursor down one line in the buffer
This bug occurs if the cursor is inside a PHP method or at the signature, and I call that Vim function. It does successfully change the visibility of the method though.
phpactor 0.12.0
NVIM v0.3.1
Ubuntu 16.04.6 LTS

elythyr:
Fixes #770

Or at least an attempt :)
We would need some tests to check more scenarios.
I'm not familiar with vim tests, I saw there is some in the `plugin` directory.
Are they still up to date ?
Is it possible to check the cursor position within a test ?

If yes I might try to create some so we could add every failing scenario and keep being sure everything works correctly.

As I explain in my last commit, I decided a totally different approach.
I realize that if I struggle so much it was mainly because I was trying to figure out a way to localize and move a "virtual" cursor which would have represented the line that was deleted.
It's event hard to explain so I guess it was indeed a bad idea!

Then I got a new idea, what I instead of focusing on the deleted line I focus on the remaining one.
The one where the cursor ends up after the deletion.
If we memorize the difference of lines between this one and the deleted one, then we just have to "reposition" the cursor relatively to it's current position.

Does it make sense to you ?
Once the transformation is done the cursor should still be X lines above the first unchanged one.

### DIFF
--- a/plugin/tests/apply_text_edits/replace_line_which_move.vader
+++ b/plugin/tests/apply_text_edits/replace_line_which_move.vader
@@ -1,0 +1,49 @@
+Include: utils.vader
+
+Given php (Original source):
+  <?php
+
+  class Foo
+  {
+      /**
+       * @var string
+       */
+      private $foo:
+
+      private function foo()
+      {
+          return $this->foo();
+      }
+  }
+
+Execute (Goes to the method line and changes the visibility):
+  10
+  call Apply(
+    \ Delete(10),
+    \ Insert(10, '    /**'),
+    \ Insert(11, '     * @var int $i'),
+    \ Insert(12, '     */'),
+    \ Insert(13, '    private function foo(int $i)')
+  \ )
+
+Then (Then cursor should still be on the method signature):
+  call AssertCursorOnLine(13)
+
+Expect php (The docblock should be added and the signature changed):
+  <?php
+
+  class Foo
+  {
+      /**
+       * @var string
+       */
+      private $foo:
+
+      /**
+       * @var int $i
+       */
+      private function foo(int $i)
+      {
+          return $this->foo();
+      }
+  }
--- a/plugin/phpactor.vim
+++ b/plugin/phpactor.vim
@@ -366,6 +366,8 @@ function! phpactor#_applyTextEdits(path, edits)
     call phpactor#_switchToBufferOrEdit(a:path)
 
     let postCursorPosition = getpos('.')
+    let curLine = postCursorPosition[1]
+    let curLineBeforeDeletion = v:null
 
     for edit in a:edits
 
@@ -388,6 +390,12 @@ function! phpactor#_applyTextEdits(path, edits)
         let linesToDelete = end['line'] - start['line']
         if linesToDelete > 0
             call execute('normal ' . linesToDelete . 'dd')
+
+            if end['line'] < curLine " Delete above the cursor
+                let curLine -= linesToDelete " Move the cursor up
+            elseif start['line'] < curLine " Delte the cursor line
+                let curLineBeforeDeletion = curLine " Memorize it
+            endif
         endif
 
         if edit['text'] == "\n"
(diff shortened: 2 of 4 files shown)


dantleech:
Seems good based on the scenarios I can think of :)

> I'm not familiar with vim tests, 

We are using [Vader](https://github.com/junegunn/vader.vim).

The tests are working (they are executed on Travis, see `.travis/vim-plugin-test.sh` for how) you can write "unit" or "functional" tests (it just runs a script and compares a buffer to an expected state I think). You can get the cursor position with `getpos('.')`.

So _if_ you want to add tests, we could:

- Test the change visibility operation only (there are already examples of applying refactorings in tests).
- Add a "unit" test, and provide our text edits directly in Vimscript. I think this should be the easiest.

But up to you, I'm happy either way.

elythyr:
>     * Add a "unit" test, and provide our text edits directly in Vimscript. I think this should be the easiest.

That's more what I had in mind.
Calling directly `phpactor#_applyTextEdits` since its not private.
It's easiest and it's also provide the possibility to add any kind of `edits` without having to actually call phpactor.

I will try to find some time.

elythyr:
So I gave it a try but I have a problem.
Vader always delete a "paragraph" instead of just a line. And not a "paragraph" in the way vim describe it.
This is the code:
```vim
Given php (Original source):
    <?php

    class Foo
    {
        /**
         * @var string
         */
        private $foo;

        private function foo()
        {
            return $this->foo();
        }
    }

Execute (position on the method line and change the visibility):
    10
    let posBefore = getpos('.')
    " call phpactor#_applyTextEdits(
    "     \ expand('%:p'),
    "     \ [{
    "     \     'start': {'character': 0, 'line': 9},
    "     \     'text': '',
    "     \     'end': {'character': 0, 'line': 10}
    "     \ }, {
    "     \     'start': {'character': 0, 'line': 9},
    "     \     'text': '      public function foo()',
    "     \     'end': {'character': 0, 'line': 9}
    "     \ }, ]
    " \ )
    " silent execute printf('keepjumps %d,%dd _', 10, 10)
    6d
    let posAfter = getpos('.')

Expect php (The visibility to be changed):
    <?php

    class Foo
    {
        /**
         * @var string
         */
        private $foo;

        public function foo()
        {
            return $this->foo();
        }
    }

Execute (Assert the cursor is still on the same line):
    AssertEqual posBefore, posAfter
```

If you test it, you see that the `6d` deletes all the docblock tag instead of just the line in the middle of it.
`10d` deletes all the method block instead of just the signature ><

I don't quite understand what append...

elythyr:
I added a few tests, more to have a test system in place than for really testing a lot of scenarios.
I think it's more efficient to add them every time there is a new bug than to write a lot of early tests.

So I think it's ready for a test drive on develop :D

plugin/tests/apply_text_edits/replace_line_which_move.vader
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+Execute (Goes to the method line and changes the visibility):
+  10
+  call Apply(
+    \ Delete(10),
+    \ Insert(10, '    /**'),
+    \ Insert(11, '     * @var int $i'),
+    \ Insert(12, '     */'),
+    \ Insert(13, '    private function foo(int $i)')
```
dantleech: might be a stupid question but why don't we just have the above `Given` and then do `10d` and envoke the change visiblity, like wioth the `insert use` test?

```
Given php (source with no namespace):
  <?php
  
  class Foobar
  {
      public function foobar(Phpactor $phactor)
      {
      }
  }
Do (put the cursor over an existing class name and add use):
  /Phpactor\<Enter>
  :call phpactor#UseAdd()\<Enter>

Expect php (the use statement to be inserted):
  <?php

  use Phpactor\Phpactor;

  
  class Foobar
  {
      public function foobar(Phpactor $phactor)
      {
      }
  }
```

dantleech: Ah, ok, this isn't testing change visibility :)

dantleech:
Thanks! This issue has always bugged me and great to see the VIM plugin tests evolving. I updated the CHANGELOG. Let's see how it fares in develop :)

PR #778, opened 2019-06-10
Generate accessors

### LINKED ISSUE #776: Generate accessors by name
Works great :) I don't think there are currently any other refactorings that can benefit from this now -- but I can imagine changing the f.e. the [generate accessors](https://phpactor.github.io/phpactor/refactorings.html#generate-accessor) to operate based on a list -- would allow to generate accessors for all properties f.e.

Otherwise, I merged the other PR, and just have to wait for a green build here :)

_Originally posted by @dantleech in https://github.com/phpactor/phpactor/pull/773#issuecomment-500141170_

As discuss there I already had started to work on it.
But I realized that I might be taking the problem from the wrong side.

What I did is change the handler to make the offset optional and add another optional parameter for the properties names.
If names are provided I use WorseReflection to find the property.
Then I use the position and the visibility to retrieve the offset of the property, because the start of the position if before the visibility but the offset start after it. So I add the visibility length to the start position to get the offset.

Two things bugs me here:
* It feels kind of hacky to "build" the offset from scratch
* The handler should only be a bridge between the RPC extension and the actual transformers

So I came up with another solution.
I still handled both offsets and name in the handler.
But I changed the `GenerateAccessor` API to provide two functions:
* `generateFromOffset`
* `generateFromPropertyName`

The `generateFromPropertyName` has the responsibility to use WorseReflection to find the property and create the `SymbolContext` used by the transformer based on the position.
It feels less hacky since a `Symbol` can be build form a position.
But it's a bit more changes and I sill don't have a good view of the global project.

So I was wondering what is your thoughts about that and how you would prefer to implement it.

P.S.: If the second option is choose, it might also be a good idea to be able to generate a `SymbolContext` from a `ReflectionMember`.
Something like `SymbolContext::fromReflection($member)` ?
Or would it be against the original design ?

elythyr:
Fixes #776 

That's were I am at the moment.
I will start removing everything relative to the offset.
If we decide to remove the loading of the source from WR, then we should also remove the path from the parameters I think.

### DIFF
--- a/lib/Extension/CodeTransformExtra/Rpc/GenerateAccessorHandler.php
+++ b/lib/Extension/CodeTransformExtra/Rpc/GenerateAccessorHandler.php
@@ -2,18 +2,29 @@
 
 namespace Phpactor\Extension\CodeTransformExtra\Rpc;
 
+use InvalidArgumentException;
 use Phpactor\CodeTransform\Domain\Refactor\GenerateAccessor;
+use Phpactor\Extension\Rpc\Response\Input\ListInput;
 use Phpactor\MapResolver\Resolver;
 use Phpactor\Extension\Rpc\Response\UpdateFileSourceResponse;
 use Phpactor\CodeTransform\Domain\SourceCode;
 use Phpactor\Extension\Rpc\Handler\AbstractHandler;
+use Phpactor\WorseReflection\Core\Reflection\ReflectionClass;
+use Phpactor\WorseReflection\Core\Reflection\ReflectionProperty;
+use Phpactor\WorseReflection\Reflector;
 
 class GenerateAccessorHandler extends AbstractHandler
 {
     const NAME = 'generate_accessor';
     const PARAM_OFFSET = 'offset';
+    const PARAM_NAME = 'name';
     const PARAM_SOURCE = 'source';
     const PARAM_PATH = 'path';
+
+    /**
+     * @var Reflector
+     */
+    private $reflector;
 
     /**
      * @var GenerateAccessor
@@ -21,8 +32,10 @@ class GenerateAccessorHandler extends AbstractHandler
     private $generateAccessor;
 
     public function __construct(
+        Reflector $reflector,
         GenerateAccessor $generateAccessor
     ) {
+        $this->reflector = $reflector;
         $this->generateAccessor = $generateAccessor;
     }
 
@@ -33,27 +46,112 @@ class GenerateAccessorHandler extends AbstractHandler
 
     public function configure(Resolver $resolver)
     {
+        $resolver->setDefaults([
+            self::PARAM_OFFSET => null,
+            self::PARAM_NAME => null,
+        ]);
         $resolver->setRequired([
             self::PARAM_PATH,
             self::PARAM_SOURCE,
-            self::PARAM_OFFSET,
         ]);
     }
 
     public function handle(array $arguments)
     {
-        $sourceCode = $this->generateAccessor->generateAccessor(
-            SourceCode::fromStringAndPath(
-                $arguments[self::PARAM_SOURCE],
-                $arguments[self::PARAM_PATH]
-            ),
-            $arguments[self::PARAM_OFFSET]
-        );
+        $class = $this->class($arguments[self::PARAM_SOURCE]);
+
+        if (!$arguments[self::PARAM_OFFSET]) {
+            $this->requireInput(ListInput::fromNameLabelChoices(
+                self::PARAM_NAME,
+                sprintf('Properties from "%s"', $class->name()),
+                $this->propertiesChoices($class)
+            )->withMultiple(true));
+        }
+
+        if ($this->hasMissingArguments($arguments)) {
+            return $this->createInputCallback($arguments);
+        }
+
+        $originalSource = $arguments[self::PARAM_SOURCE];
+        $sourceCode = SourceCode::fromStringAndPath($originalSource, $arguments[self::PARAM_PATH]);
+
+        foreach ($this->propertiesOffsets($arguments, $class) as $offset) {
+            $sourceCode = $this->generateAccessor->generateAccessor($sourceCode, $offset);
+        }
 
         return UpdateFileSourceResponse::fromPathOldAndNewSource(
             $sourceCode->path(),
-            $arguments[self::PARAM_SOURCE],
+            $originalSource,
             (string) $sourceCode
         );
+    }
+
+    /**
+     * @param string $source
+     *
+     * @return ReflectionClass
+     */
+    private function class($source)
+    {
+        $classes = $this->reflector->reflectClassesIn($source);
+
+        if ($classes->count() === 0) {
+            throw new InvalidArgumentException(
+                'No classes in source file'
+            );
+        }
+
+        if ($classes->count() > 1) {
+            throw new InvalidArgumentException(
+                'Currently will only generates accessor by name in files with one class'
+            );
+        }
+
+        return $classes->first();
+    }
+
+    private function propertiesOffsets(array $arguments, ReflectionClass $class): array
+    {
+        $offset = $arguments[self::PARAM_OFFSET];
+
+        if (is_int($offset) || ctype_digit($offset)) {
+            return [$offset];
+        }
+
+        $offsets = [];
+        foreach ((array) $arguments[self::PARAM_NAME] as $propertyName) {
+            $offsets[] = $this->propertyOffset($class, $propertyName);
+        }
+
+        return $offsets;
+    }
+
+    private function propertyOffset(ReflectionClass $class, string $propertyName): int
+    {
+        $property = $class->properties()->get($propertyName);
+
+        $start = $property->position()->start();
+        $visibility = $property->visibility();
+
+        // start is right before the visibility
+        // and the offset start at the first space after the visibility
+        $offset = $start + strlen($visibility) + 1;
+
+        return $offset;
+    }
+
+    private function propertiesChoices(ReflectionClass $class): array
+    {
+        // Select only those from the current class because the accessor generator
+        // is not able to work with the parent class at the time
+        $properties = $class->properties()->belongingTo($class->name());
+
+        $propertiesNames = array_map(function (ReflectionProperty $property) {
+            return $property->name();
+        }, iterator_to_array($properties));
+
+        sort($propertiesNames);
+
+        return array_combine($propertiesNames, $propertiesNames);
     }
 }
--- a/lib/Extension/CodeTransformExtra/Rpc/GenerateAccessorHandler.php
+++ b/lib/Extension/CodeTransformExtra/Rpc/GenerateAccessorHandler.php
@@ -16,7 +16,6 @@ use Phpactor\WorseReflection\Reflector;
 class GenerateAccessorHandler extends AbstractHandler
 {
     const NAME = 'generate_accessor';
-    const PARAM_OFFSET = 'offset';
     const PARAM_NAMES = 'names';
     const PARAM_SOURCE = 'source';
     const PARAM_PATH = 'path';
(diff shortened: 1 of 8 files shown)


autoload/phpactor/mapping.vim
```diff
@@ -0,0 +1,101 @@
...
+        call phpactor#mapping#map('<C-w><C-]>', '<Plug>phpactorGotoDefinitionVsplit', 'n')
+    endif
+
+    " Put in last to override any default mapping if needed
+    for mapping in get(g:, 'phpactorCustomMappings', [])
+        call call('phpactor#mapping#map', add(copy(mapping), v:true))
+    endfor
+endfunction
+
+" vim: et ts=4 sw=4 fdm=marker
```
dantleech: I guess this is a leak from the other PR?

lib/Extension/CodeTransformExtra/Rpc/GenerateAccessorHandler.php
```diff
@@ -33,27 +46,86 @@ public function name(): string
...
+        if ($this->hasMissingArguments($arguments)) {
+            return $this->createInputCallback($arguments);
+        }
+
+        $originalSource = $arguments[self::PARAM_SOURCE];
+        $newSource = SourceCode::fromStringAndPath($originalSource, $arguments[self::PARAM_PATH]);
+
+        if ($offset = $arguments[self::PARAM_OFFSET]) {
+            $newSource = $this->generateAccessor->generateFromOffset($newSource, $offset);
+        } else {
```
dantleech: Note: As a general rule `else` is treated as a code smell in this project, so would always prefer to delegate to a method with an early return. (but I guess we remove the offset anyway).

lib/Extension/CodeTransformExtra/Rpc/GenerateAccessorHandler.php
```diff
@@ -33,27 +46,86 @@ public function name(): string
...
+        }
+
+        return $classes->first();
+    }
+
+    private function propertiesChoices(ReflectionClass $class): array
+    {
+        // Select only those from the current class because the accessor generator
+        // is not able to work with the parent class at the time
+        $properties = $class->properties()->belongingTo($class->name());
```
dantleech: :+1: Typically I can't think of situations where I want to generate an accessor for a parent class.

dantleech:
> then we should also remove the path from the parameters I think.

The path is required in all cases I think -- to know which buffer to update in the update source response.

elythyr:
> The path is required in all cases I think -- to know which buffer to update in the update source response.

Make sense

autoload/phpactor/mapping.vim
elythyr: I think so, I had remove my own mappings for the mapping branch in prevision of a PR.
So I rebased this PR on top of it to still some mappings during dev.
I just rebased the branch on top of the current develop and pushed it back.

lib/Extension/CodeTransformExtra/Rpc/GenerateAccessorHandler.php
elythyr: I had the same reaction when I wrote it but I didn't gave it more thought since the generator was not able to do it.
In general it should be a bad idea to gave access to a parent attribute anyway.

Sorry for the wait, I didn't saw you reviews last time...

tests/Unit/Extension/CoreTransform/Rpc/GenerateAccessorHandlerTest.php
```diff
@@ -4,46 +4,124 @@
...
+}
+PHP;
     const PATH = '/path/to';
-    const OFFSET = 1234;
-    const CONSTANT_NAME = 'FOOBAR';
+    const FOO_NAME = 'foo';
+    const BAR_NAME = 'bar';
+    const PROPERTIES_CHOICES = [self::FOO_NAME => self::FOO_NAME, self::BAR_NAME => self::BAR_NAME];
+    const GENERATE_ACCESSOR_ACTION = 'generate_accessor';
+    const CURSOR_OFFSET = 57;
```
dantleech: Here I would generally use a data provider, rather than constants, as normally you may want to vary some of these values for each test, rather than gradually building up the class constants over the tests lifetime. This _could_ also reduce the amount of code, as there is only one test with many inputs?

lib/Extension/CodeTransformExtra/Rpc/GenerateAccessorHandler.php
```diff
@@ -42,18 +58,72 @@ public function configure(Resolver $resolver)
...
+            (string) $newSource
         );
     }
+
+    /**
+     * @param string $source
+     *
+     * @return ReflectionClass
+     */
+    private function class($source)
```
dantleech: We don't add docblocks in this project unless they add value -- here the type hint should be `string` and the return type should be `ReflectionClass` (an exception might be if we want to declare that the return value is an array of `Foo` (e.g. `@return Foo[]`) or a collection `Foo` of `Bar` (`@return Foo<Bar>`).

lib/Extension/CodeTransformExtra/Rpc/GenerateAccessorHandler.php
```diff
@@ -42,18 +58,72 @@ public function configure(Resolver $resolver)
...
+    {
```
dantleech: This method is duplicated from `OverrideMethodHandler`. I'm thinking it would be better to not repeat this.

Would it be worth creating some, "introspection" helper/s (e.g. `ClassAtOffset`) and implement it in both this and the override method handler? Or maybe there is a better way? Could simply be something local to this extension to start with.

tests/Unit/Extension/CoreTransform/Rpc/GenerateAccessorHandlerTest.php
```diff
@@ -4,46 +4,124 @@
...
 use Phpactor\Extension\Rpc\Response\UpdateFileSourceResponse;
 use Phpactor\CodeTransform\Domain\Refactor\GenerateAccessor;
 use Phpactor\Extension\CodeTransformExtra\Rpc\GenerateAccessorHandler;
 use Phpactor\Tests\Unit\Extension\Rpc\HandlerTestCase;
+use Phpactor\WorseReflection\ReflectorBuilder;
 
 class GenerateAccessorHandlerTest extends HandlerTestCase
 {
-    const SOURCE = '<?php echo "foo";';
+    const SOURCE = <<<'PHP'
```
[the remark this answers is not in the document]
elythyr: Will change

tests/Unit/Extension/CoreTransform/Rpc/GenerateAccessorHandlerTest.php
```diff
@@ -4,46 +4,124 @@
...
-            self::OFFSET
-        )->willReturn(SourceCode::fromStringAndPath('asd', '/path'));
+        $action = $this->handle(self::GENERATE_ACCESSOR_ACTION, [
+            'source' => self::SOURCE,
+            'path' => self::PATH,
+            'offset' => self::CURSOR_OFFSET,
+        ]);
+
+        /** @var InputCallbackResponse $action */
+        $this->assertInstanceOf(InputCallbackResponse::class, $action);
```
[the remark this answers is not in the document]
elythyr: Will remove it

tests/Unit/Extension/CoreTransform/Rpc/GenerateAccessorHandlerTest.php
elythyr: Don't remember exactly, just find a test with constants so I did it too to not have to rewrite everything.
But since then the test change enough so I've almost rewrite it already :D

lib/Extension/CodeTransformExtra/Rpc/GenerateAccessorHandler.php
elythyr: My mistake, I normally always type hint...

lib/Extension/CodeTransformExtra/Rpc/GenerateAccessorHandler.php
elythyr: What about adding it in the reflector interface ?
We already have some variation like `reflectClassLike`, what about a `reflectClassAt` ?

dantleech: The current way would be to use the collection for that (e.g. `$collection->atPosition(123)` -- plausibly it could be done for all collections, but unsure of the implementation. The general idea is to have all the "query" methods on the collection.

dantleech: ah, found an existing example :) https://github.com/phpactor/worse-reflection/blob/master/lib/Bridge/TolerantParser/Reflection/Collection/ReflectionMemberCollection.php#L39

elythyr: Ok, two things that came to mind.
Would not it be better to also add `Position::contains()` method, it feels like it's not the first time I see this condition in different places.

Does the class reflector handle anonymous class as well ?
I was thinking about this case this morning:
```php
class Foo
{
    public function foo()
    {
        $bar = new class {
            // <> cursor here
        };
    }
}
```

tests/Unit/Extension/CoreTransform/Rpc/GenerateAccessorHandlerTest.php
elythyr: I just take a look at it and I would like to let it like that if it's OK with you.
This is a really simple test class, using data providers would actually make more code.
Since the handler mainly forwards the information to a third party there is not a lot that will ever change.
I will not have the time this evening anyway so let me know for the next time.

dantleech: Yeah no worries, it's more about being consistent if anything (and I sometimes end up turning tests similar to this into data providers over time anyway :)

dantleech:
@kermorgant this is a slight change in the API (but not the protocol) for generate accessors. We have changed `generate accessor` to `generate accessors` - and is applied on the class rather than the property, will that effect the Emacs plugin?

 So I guess this is good to go otherwise -- some last things to do:

- [ ] Merge the code-transform PR
- [ ] Update the [documentation](https://github.com/phpactor/phpactor/blob/develop/doc/refactorings.md#generate-accessor)
- [ ] Update the CHANGELOG

dantleech:
Added docs and changelog in https://github.com/phpactor/phpactor/pull/785

kermorgant:
@dantleech not tested but I believe it will affect emacs integration (if we don't update phpactor.el, then it will still call `generate accessor`).

So I guess it'd be good to bump the rpc-version

elythyr:
> We have changed `generate accessor` to `generate accessors`

https://github.com/phpactor/phpactor/blob/0d8795c6f1ce3171e6f664884962cd9464821ab8/lib/Extension/CodeTransformExtra/Rpc/GenerateAccessorHandler.php#L18

So it didn't changed or are you talking about something else ?
I think that as long as the emac plugin is able to dispatch the input list response it should be all good, and for what I saw it seems to behave like the vim plugin.

Otherwise if I can change something to be compliant with the emac plugin just let me know.

dantleech:
> So it didn't changed or are you talking about something else ?

Ah ... I saw changes to the RPC handler and assued it was a BC break, but the first call remains the same, it's only the callback that changes? In which case it should be fine.

dantleech:
Replaced by #787

elythyr:
> > So it didn't changed or are you talking about something else ?
> 
> Ah ... I saw changes to the RPC handler and assued it was a BC break, but the first call remains the same, it's only the callback that changes? In which case it should be fine.

Yep only the callback response.

PR #779, opened 2019-06-10
POC: Integrating Maestro to co-ordinate packages.

dantleech:


config/maestro/composer.json
```diff
@@ -0,0 +1,24 @@
+{
```
dantleech: rm

PR #781, opened 2019-06-15
Only try to import a class when `import_class` is not empty

### LINKED ISSUE #780: Omni completion open fzf list to choose a class
Found a bug which should be related to the inputlists changes.

When using omni completion to complete a class name, the automatic import of the use statement open an inputlist.

I will look into it quickly.

elythyr:
Fixes #780


dantleech:
Thanks :)

elythyr:
You're welcome

PR #790, opened 2019-06-29
Search for class import candidates using full name

dantleech:
- the `name` is not passed from the VIM plugin, only the offset.
- Phpactor then determines the "word" under the cursor
- Word boundary chars do not include `\`

### DIFF
--- a/lib/Extension/CodeTransformExtra/Rpc/ImportClassHandler.php
+++ b/lib/Extension/CodeTransformExtra/Rpc/ImportClassHandler.php
@@ -15,6 +15,7 @@ use Phpactor\Extension\Rpc\Response\UpdateFileSourceResponse;
 use Phpactor\Extension\Rpc\Response\Input\TextInput;
 use Phpactor\CodeTransform\Domain\Refactor\ImportClass\ClassAlreadyImportedException;
 use Phpactor\CodeTransform\Domain\Exception\TransformException;
+use Phpactor\TextDocument\Util\WordAtOffset;
 
 class ImportClassHandler extends AbstractHandler
 {
@@ -64,7 +65,6 @@ class ImportClassHandler extends AbstractHandler
             self::PARAM_ALIAS => null,
         ]);
         $resolver->setRequired([
-            self::PARAM_NAME,
             self::PARAM_OFFSET,
             self::PARAM_SOURCE,
             self::PARAM_PATH,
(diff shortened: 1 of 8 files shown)


lib/Extension/CodeTransformExtra/Rpc/ImportClassHandler.php
```diff
@@ -74,12 +74,13 @@ public function configure(Resolver $resolver)
     public function handle(array $arguments)
     {
         if (null === $arguments[self::PARAM_QUALIFIED_NAME]) {
-            $suggestions = $this->suggestions($arguments[self::PARAM_NAME]);
+            $name = (new WordAtOffset('\s|%|\(|\)|\[|\]|:|;|\r|\r\n|\n'))($arguments[self::PARAM_SOURCE], $arguments[self::PARAM_OFFSET]);
```
dantleech: This utility is from `phpactor/text-document`, this regex is the default split pattern without the `\` char. It would probably make sense to add this pattern as a constant in that utility.

lib/Extension/SourceCodeFilesystemExtra/SourceCodeFilestem/Application/ClassSearch.php
```diff
@@ -36,6 +36,7 @@ public function __construct(FilesystemRegistry $filesystemRegistry, FileToClass
     public function classSearch(string $filesystemName, string $name)
     {
         $filesystem = $this->filesystemRegistry->get($filesystemName);
+        $name = str_replace('\\', '/', $name);
```
dantleech: This class isn't under test currently, and one of the "big" upcoming tasks would be to provide a dedicated service for finding classes and other things (e.g. functions), and this service would replace this code anyway...

lib/Extension/CodeTransformExtra/Rpc/ImportClassHandler.php
dantleech: Could now use `WordAtOffset(WordAtOffset::SPLIT_QUALIFIED_PHP_NAME)`

Einenlum:
:+1:

dantleech:
Sorry generally away from the computer atm, I think this just needs to be updated with the suggested constant.

Not really happy with the solution, but the class search stuff _should_ be replaced, so this is only a very harmless quick fix that will surely not last for the next several years.

dantleech:
@kermorgant this is an RPC break -- `name` is no longer required and it will error if provided. Let me know if you want to have a BC period.

dantleech:
Updaed but build is failing

dantleech:
Fixed tests and rebased

PR #796, opened 2019-07-08
Preserve original path if available when generating new class

dantleech:
Note, this is lacking any test. Will see if I get time to add an integration test here.

---
The ClassNew class can accept either an FQN or a file path, prior to
this fix if it was given a file path, it converted it to a class name,
but then used the class name to convert it back to the file path - which
is not 100% reliable, as it's possible for a class to exist at multiple
locations.

This fix preserves the file location if given.


przepompownia:
It works at least on my example.

PR #813, opened 2019-08-25
Add license information to composer.json

zonuexe:
The lack of "license" is warned by the `composer validate` command.
This change must be applied to composer.lock before release.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -3,6 +3,7 @@
     "description": "PHP refactoring and intellisense tool for text editors",
     "minimum-stability": "dev",
     "prefer-stable": true,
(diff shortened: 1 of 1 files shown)


dantleech:
Thanks, I'll make `composer validate` part of the release process in the future :+1:

zonuexe:
This warning appears on the Packagist page and may annoy some people.

[スクリーンショット 2019-08-26 6 14 35](https://packagist.org/packages/phpactor/phpactor)

dantleech:
There is now 0.13.2 with this fix, thanks

PR #823, opened 2019-10-13
Vim extract functions handles motions

elythyr:
This PR allow the function of the vim plugin allowing to extract some code to a method or an expression to also handle motions in order to provide a better user experience.

It's from my previous idea about idea default mappings.
Since it's not something you want here I made another plugin just to add the mappings on my side.
But I would still need this part to be merged in order to user motions.

For those who where interested in the default mappings you can get them from:
http://github.com/elythyr/phpactor-mappings

### DIFF
--- a/plugin/phpactor.vim
+++ b/plugin/phpactor.vim
@@ -165,6 +165,10 @@ function! phpactor#ExtractExpression(isSelection)
     call phpactor#rpc("extract_expression", { "path": phpactor#_path(), "offset_start": selectionStart, "offset_end": selectionEnd, "source": phpactor#_source()})
 endfunction
 
+function! phpactor#ExtractConstant()
+    call phpactor#rpc("extract_constant", { "offset": phpactor#_offset(), "source": phpactor#_source(), "path": phpactor#_path()})
+endfunction
+
 function! phpactor#ClassExpand()
     let word = expand("<cword>")
     let classInfo = phpactor#rpc("class_search", { "short_name": word })
--- a/plugin/phpactor.vim
+++ b/plugin/phpactor.vim
@@ -103,7 +103,7 @@ function! phpactor#Complete(findstart, base)
     endif
 
     return completions
-endfunc
+endfunction
 
 function! phpactor#_completeTruncateLabel(label, length)
     if strlen(a:label) < a:length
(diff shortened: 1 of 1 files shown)


plugin/phpactor.vim
```diff
@@ -145,24 +145,50 @@ endfunction
...
+        let positions.start = phpactor#_offset()
+        let positions.end = v:null
+    elseif a:type ==? 'v' " Visual mode
+        let positions.start = phpactor#_selectionStart()
+        let positions.end = phpactor#_selectionEnd()
+    else " Linewise or characterwise motion
+        let linewise = 'line' == a:type
+
+        let positions.start = s:getStartOffsetFromMark("'[", linewise)
+        let positions.end = s:getEndOffsetFromMark("']", linewise)
```
dantleech: So `v` and `True` are the same? What does linewise / characterwise do? (not familiar with this)

plugin/phpactor.vim
```diff
@@ -145,24 +145,50 @@ endfunction
...
+        let positions.start = phpactor#_selectionStart()
+        let positions.end = phpactor#_selectionEnd()
+    elseif a:1 ==? 'v' " Visual mode
+        let positions.start = phpactor#_selectionStart()
+        let positions.end = phpactor#_selectionEnd()
+    else " Linewise or characterwise motion
+        let linewise = 'line' == a:1
+
+        let positions.start = s:getStartOffsetFromMark("'[", linewise)
+        let positions.end = s:getEndOffsetFromMark("']", linewise)
```
dantleech: If this logic is the same as below, can we extract a common method?

plugin/phpactor.vim
dantleech: I guess the difference is that expression can work with just a single offset, not sure if method works like that - in theory it could extract the expression under the cursor.

plugin/phpactor.vim
dantleech: I guess the difference is that expression can work with just a single offset, not sure if method works like that - in theory it could extract the expression under the cursor.

plugin/phpactor.vim
elythyr: it's a vim notion so I will refer you to the doc, since it's been a while I don't remember exactly and it will be more accurate this way:
```vim
:h linewise
```

plugin/phpactor.vim
elythyr: Yes the important is not the two lines but the logic block just above.
And we don't handle expression and method extraction the same way.
Also rule of three: if there is only those 2 occurences I prefer to wait until refactoring.

dantleech:
Let's see :)

elythyr:
That was quick!

PR #824, opened 2019-10-13
Fix 6.0 typo

chinleung:


dantleech:
It was a joke because 6 was famously never released :eyes: but not very funny I guess :)

chinleung:
@dantleech Oh haha! That's why I thought it was a typo because it never existed. 🙃

PR #825, opened 2019-10-15
Bugfix 777

dantleech:


dantleech:
I've just updated the composer.lock and updated the CHANGELOG.

It looks good now @elythyr .

elythyr:
Great thx !

PR #828, opened 2019-10-18
Add keepjumps when using "goto"

elythyr:
Make a better user experience by preserving je jumplist

To see what I'm talking about, without the change go to a file.
Once there use the "goto" functionality to jump to a method definition in another file.
Then if you try to use `<C-O>` to go back you will stay in the file but your cursor will go to the previous position (inside this file).
By adding keepjumps, you will be take back right on the method call on which you used the "goto".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ Features:
   - [context-menu] Invoke menu for the nearest actionable node (i.e. you can
     invoke the context menu on whitespace now) - @elythyr
   - [class-mover] Jump to interface implementation
+  - [vim-plugin] Extract functions handles motions @elythyr
+  - [vim-plugin] Jumping to another file preserves the jumplist @elythyr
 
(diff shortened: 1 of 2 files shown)


dantleech:
Sounds perfect :) Can you add a CHANGELOG entry (and credit yourself as with the previous one?)

elythyr:
Ok will do it in this one for both

elythyr:
Done, I don't need the credit you made a lot more valuable job by your own already!

kminh:
With the latest version it seems that this issue has come back. `<c-o>` brings you back to the previous location in the newly opened file, instead of the original location where you execute the `phpactor#GotoDefinition()` method.

elythyr:
@kminh Can you open an issue about that plz ?
I can't test it right know and not everyone will read this merge PR.
So with an issue you've more chances that someone will fix it.

Issue #830, opened 2019-10-20
Vim - Use fzf to show the result of "find references"


elythyr:
This is a proposal to use fzf to show the result of the \find references` action.
Right now the quickfix is used and I don't find it very practical.
It's a great tool to achieve something on multiple buffers at once but it's not easy to navigate in.

So I was wondering what others might thing about that?

My idea so far is to use fzf to show the list of references, taking advantage of the preview feature to be able to see the references easly and select only those we choose.
I want to try to add a shortcut which will allow to populate the quickfix with the selection so we can still take advantage of it and be able to filter it.

PR #832, opened 2019-10-20
Fzf references

elythyr:
Implements #830 

So everyone can test from home :)

As it was for the list input, [fzf](https://github.com/junegunn/fzf) must be installed, otherwise the script fallback to quickfix list.
Forgot to mention that when in fzf you can select multiple entries, and press `ctrl-q` to build the quickfix list from the selection, so we don't loose the ability to do something like `:cdo` behind it.
In addition [fzf.vim](https://github.com/junegunn/fzf.vim) must also be installed as a vim plugin.
It's needed for the preview, I could had a degraded mode to handle it in phpactor plugin.
But its not the goal of this plugin so I prefer no to and let the user install everything if he wants.

There is a few tools handled by [fzf.vim](https://github.com/junegunn/fzf.vim) to colorized the preview, I personally use [bat](https://github.com/sharkdp/bat) (no it's not only for the name!)

And since a picture worth a thousand words:
fzf-references-preview


dantleech:
> Did I messed up with the picture or do you have to authorize it first ?

It looks like it needs authorization :)

elythyr:
Fixed the issue with the image, and add the fact that we can build the quickfix list from the fzf selection.
Forgot to talk about the first time :)

dantleech:
One thing I notice is that it doesn't jump to the reference when you open the file? (instead it goes to line 0)

elythyr:
I knew I forgot something !!!
Will be done before we can merge

plugin/phpactor.vim
```diff
@@ -583,31 +583,116 @@ function! phpactor#_rpc_dispatch(actionName, parameters)
...
-        if len(a:parameters['file_references']) == 1
-            let fileRefs = a:parameters['file_references'][0]
-            if -1 != match(fileRefs['file'], bufname('%') . '$')
-                return 
-            endif
+            return
         endif
 
-        execute ':cwindow'
+        let formated = s:align_pairs(keys(results), '^\(.\{-}:\d\+:\d\+:\)\s*\(.*\)\s*$', 100)
```
dantleech: would it make sense to separate this into a separate function?

dantleech: Possibly even another file/module (however autoloading works) if the new, associated, functions belong only to this strategy?

dantleech:
Have been using this for the past few days, and it's great :+1: 

We should add some documentation, maybe a new section [in the vim-plugin doc](https://github.com/phpactor/phpactor/blob/develop/doc/vim-plugin.md)? f.e. `Extras` and detail the enhancements possible by installing `fzf` and `bat` for find references, list menus etc?

elythyr:
Ok, I'll look into it

plugin/phpactor.vim
elythyr: I thought about that.
And I'm kind of relying onto the rule of three for a while now :)
It would make sense but I'm not sure yet in which scenario we could really reuse it.
It's really dependent on the object we provide to the sink, so refactor to make the code more readable yes, to reuse I'm not sure.
Since the all function that handles rcp responses was already like that I didn't even try.
But I could create another strategy like for the lists.
Maybe a"quickfix strategy" of some sort that will require a specific input to be handled properly.

What about opening a new issue for improvement so we can deliver this part already ?

dantleech: I think extracting the method might still be a good idea, not really that concerned about extracting a module `phpactor#fzf#build_quick_fix_list` or whatever is best?

dantleech:
> What is your feedback about the full screen ?

For me it's fine (so far, I haven't actually done much coding over the past days), but my instinct is still to expect the quick-fix list - and not sure in the long run if the new method would be quicker (it probably would be).

I'm not sure if we should make this a configuration option, and if so if it should default to on or off (if off we can recommend to enable it in the docs).

elythyr:
I would say it's not so much about being quick but being able to filter the results.
The quickfix doesn't allow that (without additional plugin), also if the user has fzf installed there is a strong chance he might want to use it.
For the default value, it's a possibility but since it's an enhancement requiring fzf to be installed and that doesn't break anything we could make it a default without too much problem.

dantleech:
> we could make it a default without too much problem.

yeah, I think it would make sense -- can we add a config option to disable it? we can then trial it in `develop` before committing.

Einenlum:
Just tested this feature and it's amazing :champagne: 
Thanks a lot @elythyr, great work!

plugin/phpactor.vim
elythyr: Done, I let your resolve if that corresponds to what you were expected

elythyr:
> Just tested this feature and it's amazing 
> Thanks a lot @elythyr, great work!

Happy to see that you enjoy it!

> > we could make it a default without too much problem.
> 
> yeah, I think it would make sense -- can we add a config option (enabled by default)? we can then trial it in `develop` before committing.

Done, I didn't want to go with a strategy there (like for the list) because I don't see this as an "alternative" to the quickfix but just an intermediate functionality.

I also added some documentation about the new feature and the "extras": https://github.com/elythyr/phpactor/blob/develop/doc/vim-plugin.md

I removed the trailing spaces in phpactor.vim, it was itching me for a while now :)

I added an header line in the fzf window to advertise about the `ctrl-q` shortcut.

autoload/phpactor/quickfix.vim
```diff
@@ -0,0 +1,13 @@
...
+    let shouldUseFzf = v:true == get(g:, 'phpactorUseFzfForQuickfix', v:true)
+
+    " fzf.vim is required for fzf#vim#with_preview()
+    if shouldUseFzf && exists("*fzf#complete")
+        call phpactor#fzf#quickfix(a:entries)
+    else
+        call setqflist(values(a:entries))
+        cw " Open only if there is recognized errors only
+    endif
+endfunction
```
dantleech: I like that we separated FZF functionality into a separate file, but not sure it makes sense to have this here? would it make sense to put this back into the main file?

doc/vim-plugin.md
```diff
@@ -197,6 +199,7 @@ let g:phpactorCustomInputListStrategy = 'InputListCustomStrategy'
 - `g:phpactorInputListStrategy`: Select a strategy for the [Input
   List](#input-list).
 - `g:phpactorCustomInputListStrategy`: Specify your own strategy.
+- `g:phpactorUseFzfForQuickfix`: Defines if fzf shoudl be used to build the quickfix, default: `v:true`.
```
dantleech: It's not really using FZF for quickfix, it's using FZF for file reference navigation.

Maybe:

```
g:phpactorFileReferenceNavigation=[quickfix|fzf|auto]`
```

Which makes it a bit more future proof.

Not related to this PR - but why do we have both `g:phpactorInputListStrategy` and `g:phpactorCustomInputListStrategy`? They could be combined into one or?

autoload/phpactor/fzf.vim
```diff
@@ -0,0 +1,109 @@
...
+    endif
+
+    let maxlen = min([maxlen, max])
+
+    return map(pairs, "printf('%-'.maxlen.'s', v:val[0]).' '.v:val[1]")
+endfunction
+
+function! s:quickfix_sink(results, actions, lines) abort
+  if 2 > len(a:lines)
+    return " Don't know how to handle this, should not append
```
dantleech: should this be an Exception ?

autoload/phpactor/quickfix.vim
elythyr: The goal is to have one function to build the quickfix so that we don't have to duplicate the logic to choose between fzf or vim quickfix list.
I think the quickfix is a great tool, fzf can be really nice to filter it and we might want it for some others use cases like `find implementation` and all.
I don't know if I make a lot of sense, if I'm not clear enough let me know and I will try again :)

doc/vim-plugin.md
elythyr: > It's not really using FZF for quickfix, it's using FZF for file reference navigation.

That's what I explained above, for me it's not just to find references.
It's more a filter before the quickfix, so it make sense to have this name.
That's also why I said that I didn't go for the strategy option here, I can't think of another option to do what I wanted to do with fzf.
Which is allow to filter the results that we will build the quickfix with and possibly shortcut the all thing if we are interested in only one result.



> Not related to this PR - but why do we have both `g:phpactorInputListStrategy` and `g:phpactorCustomInputListStrategy`? They could be combined into one or?

The first one decide which strategy to use.
The second one allow to add new strategies.
I prefer it separate because a user might want to provide multiple new strategies and choose the one he wants on the fly depending on the functionality called.
For example with a custom mapping you could choose the strategy before calling the phpactor function:
```vim
nmap XXX let g:phpactorInputListStrategy = 'custom1' | <Plug>PhpactorAction
```

I did not check the syntax but that's the idea.

autoload/phpactor/fzf.vim
elythyr: I guess, I replicate most of the `s:common_sink` from fzf here so I didn't think too much about it.
Since it should not append until we messed up in the code I guess it could not hurt.

elythyr: I figure out why it was done this way while creating the exception.
The goal is to ensure that we got two lines at least which mean one "action" and at least one line selected.
But if you enter a filter that doesn't match with anything but still want to press enter to leave the window then we would go an exception.
Which is not what we want, so I think we should leave it like that.

doc/vim-plugin.md
```diff
@@ -328,3 +331,56 @@ Use `<tab>` to toggle selection and CTRL-A/CTRL-D to select all/select none.
...
+```
+
+Extras
+------
+
+In order to get the best possible experience we suggests you a few extra tools
+that will make using this plugin a lot more appreciable.
+
+fzf & fzf.vim
+-------------
```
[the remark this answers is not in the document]
elythyr: Done

doc/vim-plugin.md
dantleech: > That's what I explained above, for me it's not just to find references.

But the quick fix is used for navigating file references, `quickfix` is just the mechanism that Phpactor uses to do that.

I think I'd prefer the strategy -- the new strategy is already quite specific - it requires `fzf.vim`, and it's not unimaginable that something interesting comes along which can be used as an alternative to `fzf`.

In other news, I notice that we don't define this config option at the top of `phpactor.vim` as we do for older options (e.g. `g:phpactorPhpBin`, `g:phpactorBranch` etc). Would it make sense to declare the options there with default values if they don't exist? I think it might be easier if all the plugin's config options are defined in one place.

autoload/phpactor/quickfix.vim
dantleech: So we have a single point to select the "file references" strategy, I think that's good :) Can it be deferentiated in its role from other functions in `phpactor.vim` ?

doc/vim-plugin.md
elythyr: > But the quick fix is used for navigating file references, `quickfix` is just the mechanism that Phpactor uses to do that.

Yes, but I want to use it for other tasks too that's why I want this bit of code to be reusable and the name of the config variable generic enough.
Otherwise we would have the same logic with a different variable every time we would like to offer this feature to the user.

> I think I'd prefer the strategy -- the new strategy is already quite specific - it requires `fzf.vim`, and it's not unimaginable that something interesting comes along which can be used as an alternative to `fzf`.

I agree, we could do it then. It doesn't feel really YAGNI to do it so early.
But I won't fight you on that, if you want a strategy now I will do it.


For the options I don't mind, I took the habit to always use the `get()` function in case the variable is not configured.

dantleech: :+1: I think it's little effort which helps future proof us in the longer term whilst also fitting in more with the precedent of `inputListStrategy`

Then let's move the config key to the top of the phpactor.vim for now, at least for consistency.

dantleech: > Yes, but I want to use it for other tasks too

I think we may have misunderstood - when I say "navigating file references" I am referring to what `quickfix` does without Phpactor, i.e. navigating between `file/line/col` positions.

So then I think I'd agree and it would make sense for the `FileReferenceNavigation` (or similar) method could be used for any Phpactor action that invokes a quick list.

autoload/phpactor/quickfix.vim
elythyr: Yes I would love to see the `plugin/phpactor.vim` getting smaller and smaller!
It makes it easier to find things and this file is load when vim is opened.
So the less thing there is in it the less we slow vim when starting even if it's just a bit.
Plus if we want to add a strategy and possibly later another strategy I really think it's better to centralize all the logic in one specific place.

elythyr:
Ok so there is a bit of a rework to add the strategy.
I wanted to organize the function in the same way that it was done for the list feature.
So I changed the file names and I also reorganize a bit.

I added `g:phpactorQuickfixStrategy`, it can take the value `fzf` or `vim` or a Funcref to provide a custom strategy.
I took a look at what I had done for the list strategy and I don't see the point of the "custom strategy" configuration.
So you're right to point it out, I still didn't touch it for now in case it would be used but I can if you wish.

autoload/phpactor/quickfix.vim
```diff
@@ -0,0 +1,166 @@
+function! phpactor#quickfix#strategy() abort
+    if !has_key(g:, 'PhpactorQuickfixStrategy')
+        return s:auto_detect_strategy()
+    endif
+
+    let Strategy = g:PhpactorQuickfixStrategy
```
dantleech: not blocking: should be lowercase S ?

dantleech: or this is a naming convention for functions?

doc/vim-plugin.md
```diff
@@ -328,3 +331,63 @@ Use `<tab>` to toggle selection and CTRL-A/CTRL-D to select all/select none.
...
+
+This plugin will allow us to use improved functionalities inside Vim. If you
+want to enjoy the full possibilities of both fzf and phpactor we strongly
+recommand you to install it!
+
+[bat](https://github.com/sharkdp/bat)
+---
+
+This is also a tool for the command-line and not a Vim plugin.
+It's ment to be used instead of the command `cat` and bring a lot to the table.
```
dantleech: `ment => meant`. Maybe: `It can be use as a powerful substutiture for `cat` - providing syntax highlighting and git integration.`

doc/vim-plugin.md
```diff
@@ -328,3 +331,63 @@ Use `<tab>` to toggle selection and CTRL-A/CTRL-D to select all/select none.
...
+
+It's used by default, among other possible tools, by fzf to print the preview
```
dantleech: `fzf` should be in backticks

doc/vim-plugin.md
```diff
@@ -328,3 +331,63 @@ Use `<tab>` to toggle selection and CTRL-A/CTRL-D to select all/select none.
...
+
+[fzf.vim](https://github.com/junegunn/fzf.vim)
+----------------------------------------------
+
+This is the actual [fzf](https://github.com/junegunn/fzf) plugin for vim. It
+requires you to have [fzf](https://github.com/junegunn/fzf) installed and
+configured.
+
+This plugin will allow us to use improved functionalities inside Vim. If you
+want to enjoy the full possibilities of both fzf and phpactor we strongly
```
dantleech: `fzf` and `phpactor` in backticks?

doc/vim-plugin.md
```diff
@@ -328,3 +331,63 @@ Use `<tab>` to toggle selection and CTRL-A/CTRL-D to select all/select none.
...
+[fzf](https://github.com/junegunn/fzf)
+--------------------------------------
+
+This is actually not a vim plugin but a tool for the command-line.
+It's shipped with a vim plugin that allows to use it inside Vim.
+
+If you have it installed and properly configured for Vim then Phpactor will
+make use of it to provide enhance functionalities, for instance:
+
+- [inputlist](#input-list) will take advantage of fzf
```
dantleech:
```
The following inputs will use FZF if available:

- [inputlist](#input-list) 
- [somethingelse](#something-else)
```

no need to have `will take advantage of` then.

doc/vim-plugin.md
```diff
@@ -328,3 +331,63 @@ Use `<tab>` to toggle selection and CTRL-A/CTRL-D to select all/select none.
...
+positions is given as a list that is directly usable by the function
+`setqflist()`. Not all informations are provided, depending on the context, so
+you should always check if the information you want is really defined before
+using it.
+
+Extras
+------
+
+In order to get the best possible experience we suggests you a few extra tools
+that will make using this plugin a lot more appreciable.
```
dantleech: I think just

```
In order to get the best experience we suggest you install a few extra tools.
```

doc/vim-plugin.md
```diff
@@ -328,3 +331,63 @@ Use `<tab>` to toggle selection and CTRL-A/CTRL-D to select all/select none.
...
+- `fzf`: Use [fzf](#fzf) to show allow you to filter the results.
+
+With the [fzf](#fzf) strategy you will still be able to get the result inside
+the quickfix by selecting the elements you are interested in and pressing
+`ctrl-q` to populate the quickfix with your selection and open it.
+
+You can use your own strategy by providing a `Funcref` to the
+`g:phpactorQuickfixStrategy` variable. When calling a strategy the list of
+positions is given as a list that is directly usable by the function
+`setqflist()`. Not all informations are provided, depending on the context, so
```
dantleech: `informations are` => `information is`. Not sure what is meant by `information` here (assume it relates to the hashmap that's passed to the func?). Maybe better to ommit this and just have it as a hidden feature for now? Otherwise it would be better to provide an example.

autoload/phpactor/quickfix.vim
```diff
@@ -0,0 +1,166 @@
+function! phpactor#quickfix#strategy() abort
+    if !has_key(g:, 'PhpactorQuickfixStrategy')
+        return s:auto_detect_strategy()
+    endif
```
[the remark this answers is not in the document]
elythyr: > Would still prefer to define all available keys in the same place, which is [here](https://github.com/phpactor/phpactor/blob/develop/plugin/phpactor.vim#L34) currently. Note also that options are `camelCase`.

Can't use camel case with Funcref, it must be pascal case, otherwise vim raise an error.

> Also still not _sure_ `quickfix` is the right term.

From the vim doctumentation:
> In Vim the quickfix commands are used more generally to find a list of positions in files.

The quickfix is just the way vim show positions in file in a way that the user can jump to those positions.

doc/vim-plugin.md
elythyr: Done, just not commited yet

doc/vim-plugin.md
elythyr: Done

doc/vim-plugin.md
elythyr: Why not, done

doc/vim-plugin.md
elythyr: Done

doc/vim-plugin.md
elythyr: Done

doc/vim-plugin.md
elythyr: Maked it `not all the keys are provided`
Because I'm afraid that just `information` is to generic and won't reflect properly what might be missing.

autoload/phpactor/quickfix.vim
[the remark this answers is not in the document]
dantleech: > The quickfix is just the way vim show positions in file in a way that the user can jump to those positions.

It's the way `vim` shows positions - Phpactor could use `quickfix` or not, it's an implementation for showing and jumping to positions. We could use FZF without using quickfix at all, but it's fine :)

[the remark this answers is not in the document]
elythyr: > We could use FZF without using quickfix at all, but it's fine, it makes sense

Absolutely, I just think we don't need a new word since we are naming a configuration variable for the vim plugin it's easier for the users to have the same language than vim. :)

dantleech:
> Would like us to move all the config options to a single place generally as detailed, and the option keys should start with a lowercase letter to be consistent with the ...

@elythyr would you be happy with that? if so I don't mind merging and doing it.

elythyr:
You mean outside of the `plugin/phpactor.vim` ?
I'm not sure about that, generally in plugins this file is like a front controller.
It will always be loaded and act as a single entry point (kind of).
So if I want to look for a configuration option default value, a mapping, a command, etc. on any plugin I will look there first.

I think we should move the logic out of this file and not the static things like options.
I didn't do it until now because I'm trying to not make big change in the infrastructure that already exists.
But if you agree we could do it progressively.
Abuse autoloading to reduce the amount of things to load when vim start and only load a feature when needed.

I hope it answer your question.

dantleech:
> You mean outside of the plugin/phpactor.vim ?

No - to keep them all in this file. I will create a PR :)

dantleech:
btw I announced this on twitter: https://twitter.com/phpactor/status/1192096001571643394 I don't know if you have an account that I can link in the future?

elythyr:
No I haven't :)

PR #837, opened 2019-10-30
Fix keepjumps for the goto feature

### LINKED ISSUE #836: #828 introduced a bug
Hi, it's about the #828 
It works great when you jump to another file, but when you jump to a position in the same file then you don't remember you're previous position.
I will try to provide a fix this evening after work.

It's not complicated but I just need a way to check if we changed file, if yes we append the `keepjumps` otherwise we don't use it.

elythyr:
Fixes #836


dantleech:
Thanks, didn't notice this :)

PR #839, opened 2019-11-06
Move option definition to phpactor.vim

dantleech:
@elythyr

### DIFF
--- a/plugin/phpactor.vim
+++ b/plugin/phpactor.vim
@@ -33,6 +33,10 @@ if !exists('g:phpactorCompletionIgnoreCase')
     let g:phpactorCompletionIgnoreCase = 1
 endif
 
+if !exists('g:PhpactorQuickfixStrategy')
+    let g:PhpactorQuickfixStrategy = v:null
+endif
+
 if g:phpactorOmniAutoClassImport == v:true
     autocmd CompleteDone *.php call phpactor#_completeImportClass(v:completed_item)
 endif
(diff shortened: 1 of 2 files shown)


plugin/phpactor.vim
```diff
@@ -33,6 +33,10 @@ if !exists('g:phpactorCompletionIgnoreCase')
     let g:phpactorCompletionIgnoreCase = 1
 endif
 
+if !exists('g:PhpactorQuickfixStrategy')
+    let g:PhpactorQuickfixStrategy = v:null
```
elythyr:
```suggestion
    let g:PhpactorQuickfixStrategy = function('phpactor#quickfix#fzf')
```

We don't need `v:null` since we already handle a default.
We cound even make `s:auto_detect_strategy()` public, something like `phpactor#quickfix#default_strategy()`
So we can use it to initialize the config variable and keep the same logic.

dantleech:
Idea is not to decouple and keep all the decisions in the module.

On 6 November 2019 19:53:18 EET, Camille Dejoye <notifications@github.com> wrote:
>elythyr requested changes on this pull request.
>
>
>
>> @@ -33,6 +33,10 @@ if !exists('g:phpactorCompletionIgnoreCase')
>     let g:phpactorCompletionIgnoreCase = 1
> endif
> 
>+if !exists('g:PhpactorQuickfixStrategy')
>+    let g:PhpactorQuickfixStrategy = v:null
>
>```suggestion
>    let g:PhpactorQuickfixStrategy = function('phpactor#quickfix#fzf')
>```
>
>We don't need `v:null` since we already handle a default.
>We cound even make `s:auto_detect_strategy()` public, something like
>`phpactor#quickfix#default_strategy()`
>So we can use it to initialize the config variable and keep the same
>logic.
>
>-- 
>You are receiving this because you authored the thread.
>https://github.com/phpactor/phpactor/pull/839#pullrequestreview-312627180

-- 

elythyr:
If you prefer it like that it's fine for me.

dantleech:
Cool, we can refactor it more generally later if required.

PR #843, opened 2019-11-24
Import missing classes

dantleech:
Import all unresolvable class names in the current namespace.

Note that this does not work with the FZF input stategy.

We return a collection of actions to import the missing classes. The `inputlist` blocks and works as expected, the `fzf` doesn't seem to block,and it seems the choice diaglogues are all opened at the same time.

@elythyr I couldn't see an obvious way to make `fzf` blocking (if that is indeed the issue):

TODO:

- [x] "temporarily" *hacked* the list strategy to use a simple list when processing a collection.
- [ ]  It's prompoting for imported classes (which would have been ambiguous?)
- [ ] Not related to this PR, but would be nice to fix the spacing issue when there are no existing class imports (as in screen recording below)


doc/refactorings.md
```diff
@@ -892,6 +893,21 @@ class Hello
 }
 ```
 
+Import Missing Classes
+----------------------
+
+Import all missing classes in the current file.
+
+- **Command**: **RPC Only**
+- **VIM plugin**: _Class context menu > Import Missing
```
greg0ire: This underscore makes things italics
```suggestion
- **VIM plugin**: Class context menu > Import Missing
```

dantleech: It should be italics, it's missing the last underscore..

elythyr:
> @elythyr I couldn't see an obvious way to make `fzf` blocking (if that is indeed the issue):

Not that I'm aware of, fzf is meant to work asynchronously so that for long process the user can start filter the results as soon as they arrived.

So if I understand correctly the handler will resolve all the class name which are not imported et return them as a list of action.
Then you want to filter this list and for each item kept execute every action to import the class.

Then it should be possible but we might have to rework the way the fzf version of the inputlist was made so that we can pass it our own "sink".
The idea is to do something like what I have done with the references:
- You create a dictionary with the value of each item to print as key and the action (as a funcref) as value
- You pass the keys to this dict to fzf so the user can filter
- You pass the dict to the "sink" so that you can retrieve the action for each selected item
- In the sink you call the action

Don't hesitate if I'm not clear or you need help.

dantleech:
With normal inputlist:

recording

With fzf method:

recording1

dantleech:
@elythyr I've added a [hack](https://github.com/phpactor/phpactor/pull/843/commits/63f41b5c49daf5aeee21db2d4ac45cb377c35250) to always use the input strategy when iterating over a collection as I don't know how to fix the FZF input.

elythyr:
I will try to find some time.
>It doesn't block and the next action is dispatched immediately.

That's what I don't understand, I though that we would receive the list of all unimported class as a collection of action allowing to import each one of them.

plugin/phpactor.vim
```diff
@@ -536,7 +546,9 @@ function! phpactor#_rpc_dispatch(actionName, parameters)
     " >> collection
     if a:actionName == "collection"
         for action in a:parameters["actions"]
+            let g:_phpactorRpcActionIsCollection = v:true
             let result = phpactor#_rpc_dispatch(action["name"], action["parameters"])
```
elythyr: So you call `phpactor#_rpc_dispatch` for each action received, before the user have filter them ?

dantleech: yes -- but this is where I'm confused - it does that already (it calls the callback handler when the items are selected). But I confess to not fully understanding everything that's going on :eyes:

elythyr: For me what appen is:
* We receive the collection of action
* In the loop we dispatch each action
* The rpc handler call fzf for each of them
  * The first fzf window is open
  * Once opened the second one is open (or not because already one windows exists but it doesn't matter)
  * and again until the end of the loop

So we must not use fzf in a loop, we need to treat it like the chain I mentioned.
So that the user choose the class for the first one, and only once he made a choice we call the next action

elythyr: I got a question, let's say that we are in a file with 3 classes not yet imported: A, B and C.
The classes A and C have multiple possible FQCN and need the user to choose, B on the other hand as only one possibility.

In the collection that the new handler returns, does all three actions are the same ?
Or do we have two actions to choose the right FQCN and one to import the only possibility for B ?

dantleech:
It will return three actions. 2 will be asking for a selection and the third will be an action to import the class. The other two actions will delegate to the standard import class handler AFAIRC (away from keyboard)

On 27 November 2019 19:04:20 CET, Camille Dejoye <notifications@github.com> wrote:
>elythyr commented on this pull request.
>
>
>
>> @@ -536,7 +546,9 @@ function! phpactor#_rpc_dispatch(actionName,
>parameters)
>     " >> collection
>     if a:actionName == "collection"
>         for action in a:parameters["actions"]
>+            let g:_phpactorRpcActionIsCollection = v:true
>let result = phpactor#_rpc_dispatch(action["name"],
>action["parameters"])
>
>I got a question, let's say that we are in a file with 3 classes not
>yet imported: A, B and C.
>The classes A and C have multiple possible FQCN and need the user to
>choose, B on the other hand as only one possibility.
>
>In the collection that the new handler returns, does all three actions
>are the same ?
>Or do we have two actions to choose the right FQCN and one to import
>the only possibility for B ?
>
>-- 
>You are receiving this because you authored the thread.
>https://github.com/phpactor/phpactor/pull/843#discussion_r351430837

-- 

dantleech:
Created an issue to resolve the FZF issue here: https://github.com/phpactor/phpactor/issues/845

PR #847, opened 2019-12-01
Unresolvable classes import #2

dantleech:
Seems I should not have merged the original PR due to the fact that it was functionally incomplete :/

- [ ] It's prompoting for imported classes (which would have been ambiguous?)
- [ ] Not related to this PR, but would be nice to fix the spacing issue when there are no existing class imports (as in screen recording below)


dantleech:
The first issue has been fixed in code-tranform.

 The second issue _is_ related to this PR and is due to the fact that the original source is used for each class import instead of using the most recently updated source (i.e. with the new imports).

This is a general issue, and as it's not too problematic, will create a new issue: https://github.com/phpactor/phpactor/issues/848

PR #857, opened 2019-12-07
Handle templates by php version

elythyr:
It goes along with https://github.com/phpactor/code-builder/pull/16 and need it to be merged first.

The code transform extension is still in the monolith so there is a bit of that here :)


dantleech:
Taken over as https://github.com/phpactor/phpactor/pull/869

PR #862, opened 2019-12-14
Integrates the code transform "style fixer" / sanitizer.

dantleech:
Integrates the code-builder CS fixer for updated code and also provides a CLI command to apply the CS fixer to single files -- which is for debugging more than anything else.

The CS fixer will be automatically applied to only those parts of the code which have been generated.


dantleech:
Looks like the code-transform library will need to be updated to use the CS-fixed code (maybe add a builder in `CodeBuilderBuilder::` in `code-builder` to simplify bootstrapping).

Issue #867, opened 2020-01-11
Can it import class name inline (for vim)


QzCurious:
I don't know if it only me wanting a way to transform a class name to its full name. And I think phpactor must good enough to do so. After seeking issues a few time. I found the functionality was [asked and implemented](https://github.com/phpactor/phpactor/issues/114) before!!

At this moment, I'm wondering if we, vim users, should dive into the code to explore how amaze it is. Or it's just accidentally lost its description. Hope these functionality can be described in the [site](https://phpactor.github.io/phpactor/).

After all, I'm don't want to be offensive to ask adding description for everything we can do with it, but want this project can be used efficiently by everyone.

dantleech:
Hi - yes you're correct. This isn't documented on the site currently - it's actually a VIM plugin only feature, would it make sense to add the shortcut to the example mappings?

```
nmap <Leader>e :call phpactor#ClassExpand()<CR>
```

and add it to the [refactorings](https://github.com/phpactor/phpactor/blob/develop/doc/refactorings.md) page?

Feel free to open a pull request :)

QzCurious:
I've made a [**Expand Class**](https://github.com/QzCurious/phpactor/commits/master) section in doc. And I alse modify the code that add a leading `\` to the result of class name to match the definition of [fully qualified name](https://www.php.net/manual/en/language.namespaces.rules.php).
I'd like to know which branch should I make a pull request to, master or develop?

_For the reason that why I just add a leading `\` instead of made a parameter to make it optional is that I think this is a facility but not dependency._

dantleech:
Great! "develop" is the branch

PR #871, opened 2020-01-12
The undocumented Expand Class

QzCurious:
- Add desciption for Expand Class in doc
- Make Expand class match definition of fully qualified name

### DIFF
--- a/doc/refactorings.md
+++ b/doc/refactorings.md
@@ -905,8 +905,8 @@ Expand the class name from unqualified name to fully qualified name.
 
 ### Motivation
 
-Although `use` statement make code cleaner, but sometimes the code would be more
-readable if we specify the full qualified name. For example, we might register
+Although importing classes can make code cleaner, sometimes the code can be more
+readable if the fully qualified name is specified. For example, we might register
 a list of listeners in a file.
 
 ### Before and After
--- a/doc/refactorings.md
+++ b/doc/refactorings.md
@@ -897,8 +897,7 @@ class Hello
 Expand Class
 ------------
 
-Instead of include `use` statement for a class under cursor,
-expand the class name from unqualified name to fully qualified name.
+Expand the class name from unqualified name to fully qualified name.
 
 - **Command**: _VIM function only_
 - **VIM plugin**: _VIM function only_
@@ -906,10 +905,9 @@ expand the class name from unqualified name to fully qualified name.
 
 ### Motivation
 
-Although `use` statement make code cleaner, but sometime the code would be more
-readable if we specify the full qualified name right in the place. For example,
-we might register a list of listener in a centralized file. We would be more
-pleasant to know the exact class right in the place.
+Although `use` statement make code cleaner, but sometimes the code would be more
+readable if we specify the full qualified name. For example, we might register
+a list of listeners in a file.
 
 ### Before and After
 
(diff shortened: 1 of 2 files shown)


doc/refactorings.md
```diff
@@ -893,6 +894,55 @@ class Hello
 }
 ```
 
+Expand Class
+------------
+
+Instead of include `use` statement for a class under cursor,
+expand the class name from unqualified name to fully qualified name.
```
dantleech: Could remove the first sentence, and just have `Expand the class name ..`?

doc/refactorings.md
```diff
@@ -893,6 +894,55 @@ class Hello
...
+- **Command**: _VIM function only_
+- **VIM plugin**: _VIM function only_
+- **VIM function**:`:call phpactor#ClassExpand()`
+
+### Motivation
+
+Although `use` statement make code cleaner, but sometime the code would be more
+readable if we specify the full qualified name right in the place. For example,
+we might register a list of listener in a centralized file. We would be more
+pleasant to know the exact class right in the place.
```
dantleech: - sometime => sometimes
- listener => listeners
- could remove "right in the place" and "centeralized"?
- could remove "We would be more pleasant to know the ..." sentence?

plugin/phpactor.vim
```diff
@@ -209,7 +209,7 @@ function! phpactor#ClassExpand()
 
     let line = getline('.')
     let char = line[col('.') - 2]
-    let namespace_prefix = classInfo['class_namespace'] . "\\"
+    let namespace_prefix = "\\" . classInfo['class_namespace'] . "\\"
```
dantleech: I'd rather revert this -- I actually don't use this expanding class names in PHP files (I always import) but rather in configuration files (e.g. Symfony DI container configurations). So while this "fixes" it in one respect, it would "break" it in others.

dantleech:
Thanks for this :)

QzCurious:
Done. Thanks for your kindness.

doc/refactorings.md
```diff
@@ -893,6 +894,53 @@ class Hello
...
+
+- **Command**: _VIM function only_
+- **VIM plugin**: _VIM function only_
+- **VIM function**:`:call phpactor#ClassExpand()`
+
+### Motivation
+
+Although `use` statement make code cleaner, but sometimes the code would be more
+readable if we specify the full qualified name. For example, we might register
+a list of listeners in a file.
```
dantleech: => Although importing classes can make code cleaner, sometimes the code can be more readable if the fully qualified name is specified.

dantleech: Does that make sense?

QzCurious: That sounds fine, I'll take it.
Let me ask that if the **"but"** kind of weird in there so you trying to drop it? (Sorry, I'm not so good at English)

dantleech: Yes, it's a bit weird -- it would work with `Blah statements make the code cleaner, but sometimes`. I guess it's either `although` or `but`. (disclaimer I'm a native speaker but not a language expert, so cannot say _why_, just it feels better :) )

dantleech:
Thanks :)

QzCurious:
Closes #867

Issue #882, opened 2020-02-09
[develop] FZF select all does not work


dantleech:
- Search for all references of a method
- References show in the FZF selector
- Press CTRL-Q to add all to the QF list
- Only first entry added to the QF list

przepompownia:
>     * Press CTRL-Q to add all to the QF list or select them individually
Ctrl-Q adds to QF list only items selected manually. It seems to work as expected for me (with the fresh fzf version).

przepompownia:
Ctrl-A is mapped to select all items before choice the action..

dantleech:
ah ok - maybe we can improve the text then

elythyr:
`Press CTRL-Q to add any selected result to the quickfix list` maybe ?

Issue #845, opened 2019-12-01
[develop] `inputlist` strategy is forced for collection actions when FZF is used


dantleech:
The FZF input selector does not work as expected with collection items apparently.

We introduced a hack to force the choice strategy to `inputlist` in #843. See this issue for details

chinleung:
Is this why my import list is no longer searchable? 🤔 Was about to make an issue for that. Basically, when I run `phpactor#UseAdd()`, I used to get a buffer with the list of namespaces and I could search through the list to quickly import the right namespace but now it is simply a list where I can only enter a number.

dantleech:
I guess you were using the `develop` branch?

Yes you can enable the FZF stuff as detailed here :) https://phpactor.github.io/phpactor/vim-plugin/experimental.html

chinleung:
> I guess you were using the `develop` branch?
> 
> Yes you can enable the FZF stuff as detailed here :) https://phpactor.github.io/phpactor/vim-plugin/experimental.html

Thanks! Yes I'm using the `develop` branch and that's exactly what I was looking for. It is working properly now. 😄

dshoreman:
This was posted [in a comment][1] on #1218 by @camilledejoye:

> We really need to take care of those lists issues, I still haven't found a solution for the fzf issue with multiple lists and I don't think we will ever found one...
> Maybe we could take a look at [liuchengxu/vim-clap](https://github.com/liuchengxu/vim-clap) instead.

@camilledejoye I haven't used it yet, but wouldn't it suffer the same issues as fzf?

The examples only seem to mention calling "commands" saved in .vimrc, but if we're able to call it with a command directly then it *could* work. If the issue with fzf is calling it in a loop though, how would that process go with vim-clap?

> Since it runs directly in vim I hope we won't have the same issues, it forces us to add an extra dependency though but it should be acceptable as an optional feature.

As it's a soft dependency we'd probably be ok (much like fzf). Since list.vim already checks for collections, we could update it to `isCollection && !hasVimClap` with an extra case for vim-clap in `auto_detect_strategy()`. For anyone without clap, it'd work as it did before

[1]: https://github.com/phpactor/phpactor/issues/1218#issuecomment-792553011

PR #883, opened 2020-02-09
Make FZF / BAT strategies "experimental"

dantleech:
FZF (and BAT) offers a good experience but they're not without it's drawbacks.

- FZF is async by design and we are using for a synchronous process #845 and not sure if we can fix this.
- Branching behavior based on the presence FZF (more maintainance)
- It adds an additional step when wanting to _navigate through_ file references (and it currently doesn't work on my machine #882 )
- We cannot depend on the behavior of FZF or BAT - we cannot depend on a specific version.

This PR:

- Removes the strategy auto-dection
- Simplifies the configuratoin value to be a single callable string
- Defaults to the basic VIM functionality.
- Updates the documentation to have a "vim plugin experimental" page detailing the FZF and BAT integrations.


plugin/phpactor.vim
```diff
@@ -37,8 +33,12 @@ if !exists('g:phpactorCompletionIgnoreCase')
...
 endif
 
-if !exists('g:PhpactorQuickfixStrategy')
-    let g:PhpactorQuickfixStrategy = v:null
+if !exists('g:phpactorQuickfixStrategy')
+    let g:phpactorQuickfixStrategy = 'phpactor#quickfix#vim'
+endif
+
+if !exists('g:phpactorInputListStrategy')
+    let g:phpactorInputListStrategy = 'phpactor#input#list#inputlist'
```
dantleech: Moved these configurations to the main `phpactor.vim` so all configs are in the same place, and made them more explicitly callbacks.

autoload/phpactor/quickfix.vim
```diff
@@ -1,45 +1,15 @@
-function! phpactor#quickfix#strategy() abort
```
dantleech: Removed auto-detection - I would prefer everybody had the same base experience - and right now the native capability is more stable.

elythyr:
I haven't tested but it's fined by me.
You could even remove it entirely from master it's not a problem as long as it is available on develop.

PR #889, opened 2020-02-23
VIM Plugin Resturcturing and help

dantleech:
- Use [vimdoc](https://github.com/google/vimdoc) to generate the vim help file (`doc/phpactor.txt`).
- Add commands (e.g. `:PhpactorImportClass` instead of `:call phpactor#ImportClass`) for all actions.
- Move things into `autoload` and `ftplugin` generally.
- Moves much VIM-plugin specific documentation to the VIM helpfile.
- Replaces the old VIM help (which was never updated)


doc/vim-plugin.md
```diff
@@ -4,37 +4,25 @@ currentMenu: vim-plugin
...
-- [Extras](#extras)
-
 Installation
 ------------
 
 **Prerequisites**:
 
 - [Composer](https://getcomposer.org/download)
-- PHP 7.0
+- PHP 7.2
```
dantleech: The next release will only support 7.2

Issue #747, opened 2019-03-14
Implement a loading message for the vim plugin


Einenlum:
In a big codebase, when we try to find references of a class member, PHPActor can be quite long to handle the task (which is normal). Then VIM hangs for maybe 30 seconds, with no possibility of interacting and no message.

Could it be possible to add a message from the plugin, to say that it is loading (to know that VIM is not just dead)?

elythyr:
Hi,

I'm wondering if asynchronous refactoring won't be dangerous.
We could overwrite some modifications done by PHPActor if the file is currently open in our editor and we modify it.

But it would be a great fixture to find references in the project.
Like Einenlum I work with a pretty large code base which make those tasks quite long.
It's a bummer to have to wait, not being able to do anything in Vim, in the meanwhile.

I will also use this opportunity to ask if it's possible to tell to PHPActor to ignore some directories when trying to refactor or looking for references ?
But keeping them managed by git ?
I didn't find anything about that for now but I just started my research :)

dantleech:
> I'm wondering if asynchronous refactoring won't be dangerous

Possibly, but it's already the case that open files can be changed by project-wide refactorings.

> I will also use this opportunity to ask if it's possible to tell to PHPActor to ignore some directories when trying to refactor or looking for references ?

It's not currently possible, although "easy" to implement (depending on the exact use case -- per refactoring or globally?). But let's open a separate ticket to discuss further :)

elythyr:
> It's not currently possible, although "easy" to implement. It also depends on the exact use case -- per refactoring or globally?. But let's open a separate ticket to discuss further :)

Thanks, I will keep searching and came back with a new issue if I still feel the need.

dantleech:
We can make find references and such things asynchronous with https://github.com/phpactor/phpactor/pull/890

mamazu:
The current version already shows progress of finding references. So many this issue can be closed now.

PR #890, opened 2020-02-23
[vim-plugin] Adding support for async calls

dantleech:
Adding support for making async calls to Phpactor (mainly to facilitate maintaining an indexer process).

Also adds:

```
:PhpactorIndexStart
:PhpactorIndexStop
```

to control the indexer provided by https://github.com/phpactor/workspace-query

TODO:

- [ ] Introduce request ID and/or original action name in RPC response (option to disallow concurrent requests to same RPC method? Ability to cancel job?)
- [ ] Support VIM 8 in addition to Nvim
- [ ] Nice to have: Support in-proces communication? (or at least the ability for the job to send notifications to the editor)
- [ ] Handle SIGTERM gracefully in Phpactor when running RPC (e.g. send termination message)
- [ ] Make asyncable commands async (e.g. find all references #747)

### DIFF
--- a/autoload/phpactor.vim
+++ b/autoload/phpactor.vim
@@ -311,6 +311,25 @@ function! phpactor#GenerateAccessors()
     call phpactor#rpc("generate_accessor", { "source": phpactor#_source(), "path": expand('%:p'), 'offset': phpactor#_offset() })
 endfunction
 
+let s:indexerJobId = v:null
+function! phpactor#IndexStart()
+    let s:indexerJobId = phpactor#nvim#asyncCall("index", {"watch": v:true})
+    if s:indexerJobId == v:null
+        return
+    endif
+
+    echo "Indexer starated with job ID " . s:indexerJobId
+endfunction
+
+function! phpactor#IndexStop()
+    if s:indexerJobId == v:null
+        echo "Indexer not running"
+        return
+    endif
+
+    call jobstop(s:indexerJobId)
+    echo "Indexer stopped"
+endfunction
 """""""""""""""""""""""
 " Utility functions
 """""""""""""""""""""""
--- a/autoload/phpactor.vim
+++ b/autoload/phpactor.vim
@@ -434,16 +434,8 @@ function! phpactor#rpc(action, arguments)
     let result = system(cmd, json_encode(request))
 
     if (v:shell_error == 0)
-        try 
-            let response = json_decode(result)
-        catch 
-            throw "Could not parse response from Phpactor: " . v:exception
-        endtry
 
-        let actionName = response['action']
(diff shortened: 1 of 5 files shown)


autoload/phpactor.vim
```diff
@@ -311,6 +311,25 @@ function! phpactor#GenerateAccessors()
     call phpactor#rpc("generate_accessor", { "source": phpactor#_source(), "path": expand('%:p'), 'offset': phpactor#_offset() })
 endfunction
 
+let s:indexerJobId = v:null
+function! phpactor#IndexStart()
+    let s:indexerJobId = phpactor#nvim#asyncCall("index", {"watch": v:true})
```
dantleech: Abstract the VIM/Neovim job control

ftplugin/php/commands.vim
```diff
@@ -120,3 +120,8 @@ command! -nargs=0 PhpactorChangeVisibility call phpactor#ChangeVisibility()
 ""
 " Generate accessors for the current class
 command! -nargs=0 PhpactorGenerateAccessors call phpactor#GenerateAccessors()
+
+""
+" Generate accessors for the current class
+command! -nargs=0 PhpactorIndexStart call phpactor#IndexStart()
```
dantleech: Remove spurious doc

autoload/phpactor/nvim.vim
```diff
@@ -0,0 +1,49 @@
...
+
+    let request = { "action": a:action, "parameters": a:arguments }
+
+    call chansend(job, json_encode(request))
+    call chanclose(job, 'stdin')
+
+    return job
+endfunc
+
+let s:stdout = []
```
przepompownia: I also still have no idea how to catch the response without using non-local scope (like the above). 

What do you think about creating a RPC server between vim and one-time phpactor commands? PHPCD mentioned already somewhere had [a such solution](https://github.com/przepompownia/phpcd.vim/blob/master/autoload/rpc.vim). 

I did not tested behaviour of vim-specific `ch_evalraw`.

dantleech: One idea of this PR was to allow messages to be sent _from_ Phpactor during a command's execution (i.e. to send progress notifications from the Indexer).

I'm not sure there are benefits to having an 2-way RPC connection - I can only imagine the use-case would be to cancel an operation - and in that case sending a `SIGTERM` from VIM should be enough?

dantleech: But in anycase it's worth concentrating on the language server solution for now - we can support most of Phpactor's operations (even refactoring) over LSP and it if it works well, it could replace the Phpactor VIM client one day.

PR #893, opened 2020-02-26
Allow to reuse open window when go to definition

### LINKED ISSUE #894: Please allow to reuse open window when going to definition etc.
Suppose the file containing the definition is already open in some window. When I run `:PhpactorGotoDefinitionTab` opens the second tab with the same file. Sometimes it may be expected behavior, but in general I want to use existing windows or tabs.

See #893

przepompownia:
In this minimal proposition we can enable this behavior by setting `g:phpactorUseOpenWindows` to `v:true` only (excluding calling `phpactor#_rpc_dispatch(actionName, parameters)` manually with the custom "use_open_window" parameter value). 

Closes #894

### DIFF
--- a/autoload/phpactor.vim
+++ b/autoload/phpactor.vim
@@ -314,23 +314,22 @@ endfunction
 """""""""""""""""""""""
 " Utility functions
 """""""""""""""""""""""
+function! s:isOpenInCurrentWindow(filePath)
+  return expand('%:p') == a:filePath
+endfunction
 
 function! phpactor#_switchToBufferOrEdit(filePath)
-    if expand('%:p') == a:filePath
-        " filePath is currently open
+    if <SID>isOpenInCurrentWindow(a:filePath)
         return v:false
     endif
 
     let bufferNumber = bufnr(a:filePath . '$')
 
-    if (bufferNumber == -1)
-        exec ":edit " . a:filePath
-        return v:true
-    endif
+    let command = (bufferNumber == -1)
+          \ ? ":edit " . a:filePath
+          \ : ":buffer " . bufferNumber
 
-    exec ":buffer " . bufferNumber
-
-    return v:true
+    exec command
 endfunction
 
 function! phpactor#_offset()
@@ -434,9 +433,9 @@ function! phpactor#rpc(action, arguments)
     let result = system(cmd, json_encode(request))
 
     if (v:shell_error == 0)
-        try 
+        try
             let response = json_decode(result)
-        catch 
+        catch
             throw "Could not parse response from Phpactor: " . v:exception
         endtry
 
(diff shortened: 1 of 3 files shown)


autoload/phpactor.vim
```diff
@@ -314,23 +314,22 @@ endfunction
...
 " Utility functions
 """""""""""""""""""""""
+function! s:isOpenInCurrentWindow(filePath)
+  return expand('%:p') == a:filePath
+endfunction
 
 function! phpactor#_switchToBufferOrEdit(filePath)
-    if expand('%:p') == a:filePath
-        " filePath is currently open
+    if <SID>isOpenInCurrentWindow(a:filePath)
```
dantleech: what is <SID> ?

ftplugin/php.vim
```diff
@@ -47,6 +47,8 @@ if !exists('g:phpactorInputListStrategy')
     let g:phpactorInputListStrategy = 'phpactor#input#list#inputlist'
 endif
 
+let g:phpactorUseOpenWindows = get(g:, 'phpactorUseOpenWindows', v:false)
```
dantleech: This syntax looks potentially better, but can we use the same as the above for consistency in this PR?

We are using [vimdoc](https://github.com/google/vimdoc) to generate the docuemntation. Can you add a description of this option as with the other options?

dantleech: (bonus if you can install `vimdoc` and update the help file by running `vimdoc . ` in the project root :) )

autoload/phpactor.vim
przepompownia: When you use a function that is defined as local to the script (`function s:foo()`), then using <SID> at the call as the prefix allows to avoid ambiguities especially when there exist a local function with the same name and both are run (non directly) from any external (as usually when using autoloaded functions) script. See `:help SID`.

From the technical point of view local functions are not fully protected from running from outside the script. If you run `:function` to see the list of defined functions then you may notice that local functions are also listed there. For example, at the moment I see a line `<SNR>120_string_encode(str) abort` (probably from vim-unimpaired)
and I can use it by `echo <SNR>120_string_encode('zxc')`.

ftplugin/php.vim
przepompownia: Vimdoc seems to be unavailable from pip repo but I have installed it manually without any problem. I don't know why vimdoc make this removal

```diff
-NOTE: This help is auto-generated from the VimScript using
-https://github.com/google/vimdoc. See
-https://phpactor.github.io/phpactor/developing.html#vim-help
-
```
and I have to restore it manually.

Please review all these changes for English - also function and variable names ;)

[Rector](https://github.com/rectorphp/rector) for vim is needed ;)

przepompownia:
Independently of global change of the default behavior I (at the moment) have no good idea how to improve the set of commands to one-time using with a custom "use_open_window" value.

ftplugin/php.vim
```diff
@@ -9,44 +9,36 @@ let g:phpactorInitialCwd = getcwd()
...
+
+""
+" Function to use when presenting a user with a choice of options. The default
+" is to use the VIM inputlist.
+let g:phpactorInputListStrategy = get(g:, 'phpactorInputListStrategy', 'phpactor#input#list#inputlist')
+
+""
+" When jump to the line of a file displayed in any existing window
+" reuse this window to avoid have more than one view of the same file.
+" The default is false.
```
dantleech: When jumping to a definition: if the file containing the definition is open in a window, jump to that window instead of switching buffers (?)

przepompownia: I cannot judge what version is more proper. Feel free to choose the better in your opinion.

autoload/phpactor.vim
dantleech: So if I understand correctly, we use the `s:` scope to make this a "private" function. I don't understand why we can't call `s:isOpenCurrentWindow` instead of `<SID>isOpenCurrentWindow` though?

ftplugin/php.vim
dantleech: yes, if what I said above is technically correct, then it could be better :+1:

ftplugin/php.vim
dantleech: > and I have to restore it manually.

I think I added it manually instead of adding it to the `autoload/phpactor.vim` source file ...

autoload/phpactor.vim
przepompownia: Yes, we may consider them "private" function (local to the script) remembering that we can still call it prefixed with a script number as above. 

I made basic test both on vim and neovim that suggested that using <SID> is unnecessary when a function is not called from a mapping. Indeed, the help text stands only about mapping, but it could sound more clear like [there](https://stackoverflow.com/a/16768478).

Thanks for the question.

dantleech:
@przepompownia will you also update the config documentation and regenerate the vimdoc or shall I?

przepompownia:
Shall you do it?

dantleech:
Thanks, I'll update the CHANGELOG and fix the doc :)

przepompownia:
Thanks for merging and for the whole work!

PR #895, opened 2020-02-26
Exit fzf immediatley with no result

przepompownia:
it also removes syntax error that revealed after running with '--exit-0'


przepompownia:
Independently of this fix it would be better to not call `fzf#run` (and its sink function) when there is nothing to choice.

dantleech:
I don't really understand the FZF integration very well, but trust this is a good change :)

przepompownia:
It then work like the standard way: does dot display any empty choice list (but still executes fzf binary and calls the sink function without need).

PR #897, opened 2020-02-29
Stable Context Menu Shortcuts

### LINKED ISSUE #896: [vim-plugin/rpc] Stable context menu short cuts
Currently the shortcuts for the context menu are [generated](https://github.com/phpactor/phpactor/blob/develop/autoload/phpactor/input.vim#L23) automatically based on the available choices.

The available choices change with every new feature, and so then do the shortcuts.

It would be better to define shortcuts in the RPC handler, or find another way of making them stable.

dantleech:
Allows the definition of key preferences to the context menu.

Note that we use VIM's `confirm` dialogue, which is case sensitive - which limits some our options when choosing keys.

Fixes #896

### DIFF
--- a/autoload/phpactor/input.vim
+++ b/autoload/phpactor/input.vim
@@ -33,6 +33,8 @@ function! phpactor#input#choice(label, choices, keyMap, ResultHandler)
     for choiceLabel in keys(a:choices)
         let buffer = []
 
+        " note that a:keyMap can be an empty list because PHP's json_decode
+        " can't tell the difference between an empty list and an empty dict
         if !empty(a:keyMap) && has_key(a:keyMap, choiceLabel) && !empty(a:keyMap[choiceLabel])
             let confirmLabel = s:determineConfirmLabelFromPreference(choiceLabel, a:keyMap[choiceLabel])
         else
--- a/autoload/phpactor/input.vim
+++ b/autoload/phpactor/input.vim
@@ -71,7 +71,8 @@ function! s:determineConfirmLabelFromPreference(choiceLabel, preference)
     endfor
 
     if foundShortcut == v:false
-        "call add(buffer, a:preference . '&')
+        " Could not find char in the label - add the shortcut at the end
+        call add(buffer, '&' . a:preference)
     endif
 
     return join(buffer, "")
--- a/autoload/phpactor/input.vim
+++ b/autoload/phpactor/input.vim
@@ -24,7 +24,6 @@ let s:usedShortcuts = []
 function! phpactor#input#choice(label, choices, keyMap, ResultHandler)
     let s:usedShortcuts = []
     let list = []
-    let choices = []
 
     if empty(a:choices)
         call confirm("No choices available")
@@ -41,7 +40,6 @@ function! phpactor#input#choice(label, choices, keyMap, ResultHandler)
         endif
 
         call add(list, confirmLabel)
-        call add(choices, choiceLabel)
     endfor
 
     let choice = confirm(a:label, join(list, "\n"))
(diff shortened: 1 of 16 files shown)


dantleech:
cc @elythyr @przepompownia as you seem to have better VimL skills than me :)

autoload/phpactor/input.vim
```diff
@@ -60,15 +54,56 @@ function! phpactor#input#choice(label, choices, ResultHandler)
     call a:ResultHandler(choices[choice - 1])
 endfunction
 
+function! s:determineConfirmLabelFromPreference(choiceLabel, preference)
+    let buffer = []
```
dantleech: not sure about variable scoping here -- should we use a scope prefix for local vars?

dantleech: they are local by default apparently :relieved:

przepompownia: Apparently (`:help l:` stands for edge cases but I cannot reproduce it, i.e. override `v:count` with local count), but it seems to be a good practice to explicitly define the scope (like method scope in PHP). If you use [vint](https://github.com/Vimjas/vint) (with [ALE](https://github.com/dense-analysis/ale) or [CoC](https://github.com/iamcco/diagnostic-languageserver/wiki/Linters) for example), you can notice that `ProhibitImplicitScopeBuiltinVariable` is disabled by default. This also does not look like common practice in most popular and advanced plugins.

autoload/phpactor/input.vim
```diff
@@ -32,19 +33,12 @@ function! phpactor#input#choice(label, choices, ResultHandler)
...
-                call add(buffer, '&')
-                let foundShortcut = v:true
-                call add(usedShortcuts, tolower(char))
-            endif
-
-            call add(buffer, char)
-        endfor
-
-        let confirmLabel = join(buffer, "")
+        if !empty(a:keyMap) && has_key(a:keyMap, choiceLabel) && !empty(a:keyMap[choiceLabel])
```
przepompownia:
```suggestion
        if has_key(a:keyMap, choiceLabel) && !empty(a:keyMap[choiceLabel])
```
`v:false == has_key({}, 'x')` is true. No need to check if a:keyMap is empty.

autoload/phpactor/input.vim
```diff
@@ -32,19 +33,12 @@ function! phpactor#input#choice(label, choices, ResultHandler)
...
+            let confirmLabel = s:determineConfirmLabelFromPreference(choiceLabel, a:keyMap[choiceLabel])
+        else
+            let confirmLabel = s:determineConfirmLabel(choiceLabel)
+        endif
 
         call add(list, confirmLabel)
         call add(choices, choiceLabel)
```
przepompownia: `choices` consist of all keys of `a:choises` and nothing else. It seems than you can simply assign `choices` to `keys(a:choices)` before this loop and iterate through it, isn't it?

autoload/phpactor/input.vim
dantleech: `keyMap` can be a `list` when empty due to using an `array` as the JSON source :thinking:

autoload/phpactor/input.vim
dantleech: indeed, well spotted.

autoload/phpactor/input.vim
dantleech: Nice, I didn't know there was a static analysis tool for VimL :)

autoload/phpactor/input.vim
przepompownia: Isn't `JSON_FORCE_OBJECT` helpful in such case?

dantleech: Well, we just switch to another problem then. There is probably a way to fix it on the PHP side, but I'm not that concerned.

przepompownia: Nothing too relevant to be concerned ;)

PR #898, opened 2020-02-29
Initial vint support

dantleech:
Updates `phpactor.vim` as far as I dare to fix issues reported by [Vint](https://github.com/Vimjas/vint)  - a CS tool for VimL.

Note that the VimLParser used by Vint does not understand the following (and we cannot ignore them it seems):

```
    let sink = {
        \ 'sink': {key -> a:ResultHandler(a:choices[key - 1])},
    \ }
```
```
unexpected token: > (see ynkdir/vim-vimlparser) 
```

or

```
  let items = map(copy(a:lines), {key, value -> a:results[value]})
```
```
unexpected token: , (see ynkdir/vim-vimlparser) 
```

There is an open pull request for this (only 5 days old): https://github.com/vim-jp/vim-vimlparser/issues/150

@elythyr I have to admit that I don't understand this syntax - can it be replaced with something else? :eyes:

It would be great to get Vint running in Travis, but the above seems to prevent us (cannot ignore parser errors).

### DIFF
--- a/autoload/phpactor.vim
+++ b/autoload/phpactor.vim
@@ -11,7 +11,7 @@
 "     https://github.com/google/vimdoc. See
 "     https://phpactor.github.io/phpactor/developing.html#vim-help
 
-function! phpactor#Update()
+function! phpactor#Update() abort
     let current = getcwd()
     execute 'cd ' . g:phpactorpath
     echo system('git checkout ' . g:phpactorBranch)
@@ -20,7 +20,7 @@ function! phpactor#Update()
     execute 'cd ' .  current
 endfunction
 
-function! phpactor#Complete(findstart, base)
+function! phpactor#Complete(findstart, base) abort
 
     let lineOffset = line2byte(line("."))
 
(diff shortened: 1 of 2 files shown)


autoload/phpactor.vim
```diff
@@ -10,8 +10,11 @@
 " NOTE: This help is auto-generated from the VimScript using
 "     https://github.com/google/vimdoc. See
 "     https://phpactor.github.io/phpactor/developing.html#vim-help
+" vint: -ProhibitUnnecessaryDoubleQuote
+" vint: -ProhibitImplicitScopeVariable
+" vint: -ProhibitAbbreviationOption
```
dantleech: The amount of changes required to fix all these is too much of a risk (I tried fixing them all manually, but things broke immediately).

blueyed: > " vint: -ProhibitUnnecessaryDoubleQuote

JFI: this check is actually too strict/noisy, see https://github.com/Vimjas/vint/issues/310.

blueyed: btw: in general I recommend to try/use vint master (also).

Issue #902, opened 2020-03-05
Cannot call phpactor#ContextMenu()


przepompownia:
I have updated phpactor with dependencies to check #901 
```
Phpactor returned an error: PHP Fatal error:  Uncaught TypeError: Argument 1 passed to Phpactor\Extension\ContextMenu\Model\ContextMenu::fromArray() must be of the type array
, null given, called in /home/user/.vim/bundle/phpactor/lib/Extension/ContextMenu/ContextMenuExtension.php on line 29 and defined in /home/user/.vim/bundle/phpactor/lib/E
xtension/ContextMenu/Model/ContextMenu.php:30                                                                                                                                 
Stack trace:                                                                                                                                                                  
#0 /home/user/.vim/bundle/phpactor/lib/Extension/ContextMenu/ContextMenuExtension.php(29): Phpactor\Extension\ContextMenu\Model\ContextMenu::fromArray()                    
#1 /home/user/.vim/bundle/phpactor/vendor/phpactor/container/lib/PhpactorContainer.php(79): Phpactor\Extension\ContextMenu\ContextMenuExtension->Phpactor\Extension\ContextM
enu\{closure}()                                                                                                                                                               
#2 /home/user/.vim/bundle/phpactor/vendor/phpactor/rpc-extension/lib/Registry/LazyContainerHandlerRegistry.php(40): Phpactor\Container\PhpactorContainer->get()             
#3 /home/user/.vim/bundle/phpactor/vendor/phpactor/rpc-extension/lib/RequestHandler/RequestHandler.php(26): Phpactor\Extension\Rpc\Registry\LazyContainerH in /home/user/.
vim/bundle/phpactor/lib/Extension/ContextMenu/Model/ContextMenu.php on line 30 
```
```
 $ phpactor  extension:update                                                                                                          
Loading composer repositories with package information                                                                                                                        
Updating dependencies                                                                                                                                                         
Your requirements could not be resolved to an installable set of packages.                                                                                                    
                                                                                                                                                                              
  Problem 1                                                                                                                                                                   
    - The requested package phpactor/completion-worse-extension 0.1.2 is satisfiable by phpactor/completion-worse-extension[0.1.2] but these conflict with your requirements o
r minimum-stability.                                                                                                                                                          
  Problem 2                                                                                                                                                                   
[...]

PR #903, opened 2020-03-05
Remove conflict markers and duplicated key

przepompownia:
related #902 

My json linter detected also duplicated key so I removed one of them (that without shortcut key).

### DIFF
--- a/lib/Extension/ContextMenu/menu.json
+++ b/lib/Extension/ContextMenu/menu.json
@@ -129,16 +129,6 @@
             "parameters": {
                 "offset": "%offset%",
                 "path": "%current_path%",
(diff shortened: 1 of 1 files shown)


dantleech:
Thanks :) We should really add an integration test for `menu.json`

dantleech:
Doh - there was one: https://travis-ci.org/phpactor/phpactor/jobs/658385930#L392 :man_facepalming: 

Note to self: don't push to develop.

przepompownia:
The test for installing a fully equipped language server will be appreciated too.

dantleech:
We already have that: https://github.com/phpactor/phpactor/blob/develop/config/travis/extensions-to-test

PR #907, opened 2020-03-09
Fix "VIM Plugin" example mappings and add missing command definitions

yaegassy:
@dantleech 

Thanks for this tools and vim plugin.

- [Vim] Fix typo, `phpactor#Naviagate()` -> `phpactor#Navigate()`
- [Vim] Added command definition for `PhpactorTransform`
- [Vim] Adjustments related to vim mapping
    - Command mapping contains `()`, so fix it.
    - `PhpactorImportAllClasses` command does not exist and has been removed from the mapping
    - Changed to map only php filetype

### DIFF
--- a/doc/phpactor.txt
+++ b/doc/phpactor.txt
@@ -191,20 +191,22 @@ Phpactor does not assume any mappings automatically, the following mappings
 are available for you to copy:
 >
 
-  nmap <Leader>u :PhpactorImportClass<CR>
-  nmap <Leader>ua :PhpactorImportAllClasses<CR>
-  nmap <Leader>mm :PhpactorContextMenu()<CR>
-  nmap <Leader>nn :PhpactorNavigate()<CR>
-  nmap <Leader>oo :PhpactorGotoDefinition()<CR>
-  nmap <Leader>oh :PhpactorGotoDefinitionHsplit()<CR>
-  nmap <Leader>ov :PhpactorGotoDefinitionVsplit()<CR>
-  nmap <Leader>ot :PhpactorGotoDefinitionTab()<CR>
-  nmap <Leader>K :PhpactorHover()<CR>
-  nmap <Leader>tt :PhpactorTransform()<CR>
-  nmap <Leader>cc :PhpactorClassNew()<CR>
-  nmap <silent><Leader>ee :PhpactorExtractExpression(v:false)<CR>
-  vmap <silent><Leader>ee :<C-U>PhpactorExtractExpression(v:true)<CR>
-  vmap <silent><Leader>em :<C-U>PhpactorExtractMethod()<CR>
+  augroup PhpactorMappings
+    au!
+    au FileType php nmap <Leader>u :PhpactorImportClass<CR>
+    au FileType php nmap <Leader>mm :PhpactorContextMenu<CR>
+    au FileType php nmap <Leader>nn :PhpactorNavigate<CR>
+    au FileType php nmap <Leader>oo :PhpactorGotoDefinition<CR>
+    au FileType php nmap <Leader>oh :PhpactorGotoDefinitionHsplit<CR>
+    au FileType php nmap <Leader>ov :PhpactorGotoDefinitionVsplit<CR>
+    au FileType php nmap <Leader>ot :PhpactorGotoDefinitionTab<CR>
+    au FileType php nmap <Leader>K :PhpactorHover<CR>
+    au FileType php nmap <Leader>tt :PhpactorTransform<CR>
+    au FileType php nmap <Leader>cc :PhpactorClassNew<CR>
+    au FileType php nmap <silent><Leader>ee :PhpactorExtractExpression<CR>
+    au FileType php vmap <silent><Leader>ee :<C-u>PhpactorExtractExpression<CR>
+    au FileType php vmap <silent><Leader>em :<C-u>PhpactorExtractMethod<CR>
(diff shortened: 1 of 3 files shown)


ftplugin/php/mappings.vim
```diff
@@ -4,19 +4,21 @@
...
+"     au FileType php nmap <Leader>oh :PhpactorGotoDefinitionHsplit<CR>
+"     au FileType php nmap <Leader>ov :PhpactorGotoDefinitionVsplit<CR>
+"     au FileType php nmap <Leader>ot :PhpactorGotoDefinitionTab<CR>
+"     au FileType php nmap <Leader>K :PhpactorHover<CR>
+"     au FileType php nmap <Leader>tt :PhpactorTransform<CR>
+"     au FileType php nmap <Leader>cc :PhpactorClassNew<CR>
+"     au FileType php nmap <silent><Leader>ee :PhpactorExtractExpression<CR>
+"     au FileType php vmap <silent><Leader>ee :<C-u>PhpactorExtractExpression<CR>
+"     au FileType php vmap <silent><Leader>em :<C-u>PhpactorExtractMethod<CR>
+"   augroup END
```
elythyr: The autocomand with PHP filetype are not useful in this context since you already are in an `ftplugin` that will be loaded only for PHP.
But I would instead add `<buffer`> so that the mappings are only defined for the current buffer, that will prevent any overlap with mappings from a plugin for another filetype.

doc/phpactor.txt
```diff
@@ -191,20 +191,22 @@ Phpactor does not assume any mappings automatically, the following mappings
 are available for you to copy:
 >
 
-  nmap <Leader>u :PhpactorImportClass<CR>
-  nmap <Leader>ua :PhpactorImportAllClasses<CR>
```
dantleech: This sohuld have been `:PhpactorImportMissingClasses`

doc/phpactor.txt
```diff
@@ -191,20 +191,22 @@ Phpactor does not assume any mappings automatically, the following mappings
...
-  nmap <Leader>oh :PhpactorGotoDefinitionHsplit()<CR>
-  nmap <Leader>ov :PhpactorGotoDefinitionVsplit()<CR>
-  nmap <Leader>ot :PhpactorGotoDefinitionTab()<CR>
-  nmap <Leader>K :PhpactorHover()<CR>
-  nmap <Leader>tt :PhpactorTransform()<CR>
-  nmap <Leader>cc :PhpactorClassNew()<CR>
-  nmap <silent><Leader>ee :PhpactorExtractExpression(v:false)<CR>
-  vmap <silent><Leader>ee :<C-U>PhpactorExtractExpression(v:true)<CR>
-  vmap <silent><Leader>em :<C-U>PhpactorExtractMethod()<CR>
+  augroup PhpactorMappings
```
elythyr: Like I said this should not be needed if the mappings are properly created.
Instead what about explaining that the mappings should be added as a ftplugin ?

dantleech: What's the advantage of having an `augroup` ?

dantleech: But this is for copy-pasting into your `.vimrc` - at least that's how I see them,

elythyr: The file `ftplugin/php/mappings.vim` so it's not obvious to me.
Also it's a lot better practice to not put it in the vimrc.
In this case we don't want these mappings when editing other files than PHP.

elythyr: The goal of not including mappings into phpactor was to let the user do it.
But we could still tell them how to do it properly.

If you put those mappings into your `vimrc` you will ends up having the mappings delcared for any other filetype.
Which will bring error since the plugin is now in the `ftplugin` directory the mapping will exists but not the commands.
Also by doing so we increase drastically the chances of conflict with another plugin mappings or even user defined mappings.

elythyr: When the file is reloaded `autocmd!` will delete all autocmd, so they are not declare twiced.

dantleech: I think this is outside the scope of this PR. This is just fixing the example mappings in the help.

dantleech: I would still avoid deciding on some default mappings and just suggest them.

People will come from different tools, have different shortcuts, and the defaults might conflict. I would rather people just have the ability to choose and update their mappings themselves.

ftplugin/php/mappings.vim
yaegassy: Added `<buffer>` mapping

yaegassy:
[VIM] Mapping adjustment

- Added `<buffer>` mapping
- Added `:PhpactorImportMissingClasses` mapping

doc/phpactor.txt
yaegassy: Added `:PhpactorImportMissingClasses` mapping

doc/phpactor.txt
elythyr: > I would still avoid deciding on some default mappings and just suggest them.

I know and I don't suggest to do so here.

> I think this is outside the scope of this PR. This is just fixing the example mappings in the help.

This PR changes the way the doc explains how to declare mappings.
I'm only suggesting to have only one `autocmd` that will call a function that will declare the mappings just as they are right now.

If it's still unclear feel free to resolve, no need to lost too much time on something that is commented anyway :)

dantleech:
Thanks :)

PR #908, opened 2020-03-09
Updated mappings

dantleech:
Add some missing mappings and fix some command names cc @yaegassy @elythyr

### DIFF
--- a/doc/phpactor.txt
+++ b/doc/phpactor.txt
@@ -201,10 +201,14 @@ are available for you to copy:
     au FileType php nmap <buffer> <Leader>ua :PhpactorImportMissingClasses<CR>
     au FileType php nmap <buffer> <Leader>mm :PhpactorContextMenu<CR>
     au FileType php nmap <buffer> <Leader>nn :PhpactorNavigate<CR>
-    au FileType php nmap <buffer> <Leader>oo :PhpactorGotoDefinition<CR>
-    au FileType php nmap <buffer> <Leader>oh :PhpactorGotoDefinitionHsplit<CR>
-    au FileType php nmap <buffer> <Leader>ov :PhpactorGotoDefinitionVsplit<CR>
-    au FileType php nmap <buffer> <Leader>ot :PhpactorGotoDefinitionTab<CR>
+    au FileType php,cucumber nmap <buffer> <Leader>o
+        \ :PhpactorGotoDefinition<CR>
+    au FileType php,cucumber nmap <buffer> <Leader>Oh
+        \ :PhpactorGotoDefinitionHsplit<CR>
+    au FileType php,cucumber nmap <buffer> <Leader>Ov
+        \ :PhpactorGotoDefinitionVsplit<CR>
+    au FileType php,cucumber nmap <buffer> <Leader>Ot
+        \ :PhpactorGotoDefinitionTab<CR>
     au FileType php nmap <buffer> <Leader>K :PhpactorHover<CR>
     au FileType php nmap <buffer> <Leader>tt :PhpactorTransform<CR>
     au FileType php nmap <buffer> <Leader>cc :PhpactorClassNew<CR>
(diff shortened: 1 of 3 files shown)


doc/phpactor.txt
```diff
@@ -204,9 +208,16 @@ are available for you to copy:
...
+    au FileType php nmap <buffer> <Leader>ci :PhpactorClassInflect<CR>
+    au FileType php nmap <buffer> <Leader>fr :PhpactorFindReferences<CR>
+    au FileType php nmap <buffer> <Leader>mf :PhpactorMoveFile<CR>
+    au FileType php nmap <buffer> <Leader>cf :PhpactorCopyFile<CR>
+    au FileType php nmap <buffer> <silent> <Leader>ee
+    :PhpactorExtractExpression<CR>
+    au FileType php vmap <buffer> <silent> <Leader>ee
+    :<C-u>PhpactorExtractExpression<CR>
+    au FileType php vmap <buffer> <silent> <Leader>em
+    :<C-u>PhpactorExtractMethod<CR>
```
dantleech: @yaegassy did you fix the whitespace manually or is my `vimdoc` behaving differently?

dantleech: Created an issue in vimdoc: https://github.com/google/vimdoc/issues/111

yaegassy: The whitespace has been removed manually.
I have never used `vimdoc`.

yaegassy: When using `vimdoc`, it seems that 1 line is wrapped when it becomes more than `80 characters`.

yaegassy: It was a temporary solution, but if I use `\` and limit it to 80 characters, it doesn't wrap in vimdoc either.

**ftplugin/php/mappings.vim**

```vim
"     au FileType php
"       \ nmap <buffer> <silent> <Leader>ee :PhpactorExtractExpression<CR>
"     au FileType php
"       \ vmap <buffer> <silent> <Leader>ee :<C-u>PhpactorExtractExpression<CR>
"     au FileType php
"       \ vmap <buffer> <silent> <Leader>em :<C-u>PhpactorExtractMethod<CR>
```

dantleech: Use I updated to use `\` :)

dantleech:
The goto definition mappings should also apply to `cucumber` (Gherkin) files for use with the `phpactor/behat-extension`

ftplugin/php/mappings.vim
```diff
@@ -7,19 +7,34 @@
 "   augroup PhpactorMappings
 "     au!
 "     au FileType php nmap <buffer> <Leader>u :PhpactorImportClass<CR>
```
elythyr: I'm not even sure this will works.
This file `ftplugin/php/mappings.vim` will only be loaded for files with the `php` filetype.
So the `FileType` event might already have been triggered and the auto commands won't do a thing...

dantleech: hm .. this is just documentation not a vimscript?

elythyr: Well I feel like someone reading it might try put this in a file with the same name and path.
Since it's already somewhere else in the doc, either delete this duplicate or fix it if it's indeed broken.

dantleech: What do you mean "already somewhere else"?

We are now generating the `doc/phpactor.txt` VIM help file from the source code. These files are the canonical source of truth.

dantleech: Note that in `:help phpactor` the path of this file is not visible.

elythyr: Ok, I wasn't aware of that... * shame *
Just ignore all that :)

dantleech: It's new :) https://phpactor.github.io/phpactor/developing.html#vim-help

Issue #921, opened 2020-03-15
Cannot open the context menu when cwd != workspace root


przepompownia:
Example: `https://github.com/przepompownia/phpcd.vim/blob/master/php/PHPCD.php` from one of abandoned projects but used to test phpactor behaviour. 

Trying to open `PhpactorContextMenu` ends with
```
Error from Phpactor: Could not find source with "Lvht\MsgpackRpc\Handler"
```
when I open PHPCD.php from outside its project. 

Problem does not exist only when the current directory is the project root directory. 

Another example, but on phpactor I described in https://github.com/phpactor/phpactor/issues/920#issuecomment-599264322 It also prevents from opening the menu.

dantleech:
See https://github.com/phpactor/phpactor/issues/660 it's currently expected that the project root is and stays the one in which VIM was started - but it might be possible to implement a strategy to somehow safely detect the project root (note we don't want to start refactoring the users home directory)

dantleech:
@przepompownia can be closed now?

Issue #660, opened 2018-12-04
Project Root Detection


gvanbeck:
Lots of people have issues where they cd to a directory inside of VIM, but the working directory remains the same.

We should address this either in the VIM plugin or by using a strategy to detect the project root.

original issue below

---

just updated the development branch. When running phpactor#Status() I get the wrong working dir. My current working dir is where the .phpactor.yml file is. The 'Work dir' is pointing to my other project.
(I changed the foldernames for privacy)

Info
----
Version: dafc795 (12 hours ago) 657-completion-do-not-suggest-statics-non-statics-659
Work dir: /Users/myhomefolder/Documents/Other/Directory

Diagnostics
-----------
[✔] Composer detected - faster class location and more features!
[✔] Git detected - enables faster refactorings in your repository scope!
[✔] XDebug is disabled. XDebug has a negative effect on performance.

Config files
------------
[✘] /Users/myhomefolder/.config/phpactor/phpactor.json
[✘] /Users/myhomefolder/.config/phpactor/phpactor.yml
[✘] /Users/myhomefolder/Documents/Current/.phpactor.json
[✔] /Users/myhomefolder/Documents/Current/.phpactor.yml

dantleech:
TRhe working dir sohuld be based on the folder in which you started VIM.

Can you confirm this is a regression (e.g. switching to 0.10.0 tag fixes the behavior?)

gvanbeck:
Switching to 0.10.0 did not fix it. After fiddling around I can confirm it was related to xolox/vim-session plugin. I changed the session though but the work dir still was wrong. Only closing of the session fixed it. I don't know if this can be usefull for you, if not just close this issue.

dantleech:
yeah this is a known issue for a few people, currenty the working dir is set for the duration of the VIM process. I am open to adding project directory detection if it can be done sensibly.

vivoconunxino:
Just as a suggestion; I use `vim-rooter` and never had problems with that. Here is my config:
```
"VIM-ROOTER
"---------------------------------
"what kind of files determine whether a directory is root
Plug 'airblade/vim-rooter'
let g:rooter_patterns = ['.git/','.phpactor.yml']
au BufEnter * Rooter

```

przepompownia:
Please test https://github.com/phpactor/phpactor/pull/923

dantleech:
This has been implemented in develop, closing.

PR #923, opened 2020-03-17
Determine the current workspace directory

przepompownia:


przepompownia:
Please pay attention on projects with subprojects (like phpactor) and on project inside non-unix path schemes.

dantleech:
> Please pay attention on projects with subprojects (like phpactor) and on project inside non-unix path schemes.

This would mean that, when performing an operation (e.g. find references) on a vendor class, only the vendor package would be searched?

elythyr:
I would think so because of the `composer.json` but if you remove it and let only `.git` it will works since the `.git` directory is not present in the vendor.
But it will still failed with submodules.

What is the point of the feature ? Is there a limitation with the current system ?
That could help me to get a better grasp on what you're trying to do here.

przepompownia:
> What is the point of the feature ? Is there a limitation with the current system ?
> That could help me to get a better grasp on what you're trying to do here.
> EDIT: Related to #660 ?
Directly it is related to https://github.com/phpactor/phpactor/issues/921

przepompownia:
It is not true that git index directories are always not present within root directories of dependencies. Phpactor may be the first hand example. Maybe we should be also able to overwrite the automatic choice by manual setting the workspace root in a local (per project) phpactor config.

However, I am not sure what the choice we can propose as the default.

elythyr:
I thought it was already the case with a `.phpactor` file or something like that ?

dantleech:
Where does the problem originate anyway? Is it that there is a plugin which changes the working directory or similar? Could we simply hook into something and update the Phpactor CWD?

dantleech:
btw another indicator of project root would be `phpactor.yml/phpactor.json`

przepompownia:
> Is it that there is a plugin which changes the working directory or similar?

Many similar tools (both on the server and client side) have a way to determine workspace root from the current file (based on configurable patterns) – especially when files from multiple project are open in one vim instance. 

Offtopic: multiple workspace folders are also supported by LSP. 

https://github.com/iamcco/vim-language-server (`projectRootPatterns`)
https://github.com/neoclide/coc.nvim/wiki/Using-workspaceFolders
https://microsoft.github.io/language-server-protocol/specification#workspace_workspaceFolders

> Could we simply hook into something and update the Phpactor CWD?

I'm not sure I'm understand: do you mean the global `g:phpactorInitialCwd` (used as a fallback in this PR)? Why to keep the one-time information in any global variable?

dantleech:
ok, reading the CoC documentation it makes sense to me :+1:

autoload/phpactor.vim
```diff
@@ -435,14 +435,41 @@ endfunction
 " RPC -->-->-->-->-->--
 """""""""""""""""""""""
 
+let g:phpactorProjectRootPatterns = ['composer.json', '.git']
```
dantleech: - Let's add these in https://github.com/phpactor/phpactor/blob/d5265a951494fe267e98b0a9956d83b35fbad06c/ftplugin/php.vim#L41 with some documentation (so it will find it's way into the VIM help.
- Add `.phpactor.json` and `.phpactor.yml` here?

autoload/phpactor.vim
```diff
@@ -435,14 +435,41 @@ endfunction
...
+let g:phpactorGlobalRootPatterns = ['/', '/home']
```
dantleech: can we use `$HOME` somehow? if the user had `.git` in their home directory Phpactor could potentially do some crazy things.

autoload/phpactor.vim
```diff
@@ -435,14 +435,41 @@ endfunction
...
+  while index(g:phpactorGlobalRootPatterns, l:directory) < 0
+    for l:pattern in a:rootPatterns
+      if (filereadable(l:directory .'/'. l:pattern))
+        return l:directory
+      endif
+    endfor
+    let l:directory = fnamemodify(l:directory, ':h')
+  endwhile
+
+  if index(g:phpactorGlobalRootPatterns, l:workspaceDir) >= 0
```
dantleech: `l:workspaceDir` is a local var? How is it defined at this point?

ftplugin/php.vim
```diff
@@ -40,6 +40,12 @@ let g:phpactorInputListStrategy = get(g:, 'phpactorInputListStrategy', 'phpactor
 " to that window instead of switching buffers.  The default is false.
 let g:phpactorUseOpenWindows = get(g:, 'phpactorUseOpenWindows', v:false)
 
+" The list of files that determine workspace root directory.
```
dantleech: maybe leave empty to assume the CWD or similar?

autoload/phpactor.vim
```diff
@@ -435,14 +435,38 @@ endfunction
...
 
+function! s:searchDirectoryUpwardForRootPatterns(initialDirectory, rootPatterns, fallbackDirectory)
+  let l:directory = a:initialDirectory
+  while index(g:phpactorGlobalRootPatterns, l:directory) < 0
+    for l:pattern in a:rootPatterns
+      if (filereadable(l:directory .'/'. l:pattern))
+        return l:directory
+      endif
+    endfor
+    let l:directory = fnamemodify(l:directory, ':h')
```
dantleech: this would still detect the `/home` is it had a global pattern in it or?

autoload/phpactor.vim
przepompownia: Done in ftplugin, I removed this redundant part.

autoload/phpactor.vim
```diff
@@ -435,14 +435,38 @@ endfunction
...
+  endwhile
+
+  if index(g:phpactorGlobalRootPatterns, l:directory) >= 0
+    let l:directory = a:fallbackDirectory
+  endif
```
przepompownia: Notice that there these patterns are used also to exclude them.

The one caveat is a case when `a:fallbackDirectory` is also one of them ;) 

What do you think about writing some tests?

autoload/phpactor.vim
przepompownia: See comment below

przepompownia:
Probably global root patterns should include '/' independently of user settings.

ftplugin/php.vim
przepompownia: > Add .phpactor.json and .phpactor.yml here?
> maybe leave empty to assume the CWD or similar?

If I understand, the one above sentences contradicts with each other. 

We would set the default empty and show `['composer.json', '.git', '.phpactor.json', '.phpactor.yml']` as an example in the docblock.

autoload/phpactor.vim
przepompownia: `l:directory` should be there, naturally :+1:

przepompownia:
I still cannot reproduce the errors reported by travis. Maybe something will come out after preparing travis docker container locally.

dantleech:
Can you run the Vader tests locally?

przepompownia:
Unfortunately, not. In opposite to CircleCi, where each build allows to use a SSH connection to its container, Travis requires to download the proper image in proper version and prepare an environment.  I downloaded some old, probably wrong image (about 2 GB) and gave up after I could not meet dependencies. 

A raw version of error output reveals more information than the highlighted one. I see an exception (path is not absolute) from the same file `source-code-filesystem/lib/Domain/FilePath.php` I fixed few days ago. I would like to change the extension to reveal the wrong path.

przepompownia:
I have manually made travis-like container with xenial and php7.2 and am able to reproduce the error. 

The error comes from the result `expand('%:p')` when the file does not exist yet. See `:help filename-modifiers`.

This file contains another places that can reveal this bug. Maybe they are not covered by any test.

przepompownia:
In the case of merge, squash option seems to be good idea because it consist of many not significant (as separate) commits.

dantleech:
Sorry, I forgot about this, will have a look tomorrow :crossed_fingers:

autoload/phpactor.vim
dantleech: Yes - testing is relatively easy with https://github.com/junegunn/vader.vim tests are in `tests/VimPlugin` can help out there if needed - there are several examples already. (we don't reallyt have the habbit of writing tests for the plugin, but we should)

autoload/phpactor.vim
```diff
@@ -353,7 +353,15 @@ function! phpactor#_source()
...
-    return expand('%:p')
+    let l:path = expand('%:p')
+
+    if filereadable(l:path) || stridx(l:path, '/') == 0
+      return l:path
+    endif
+
+    " todo if empty path
+    "
+    return printf('%s/%s', g:phpactorInitialCwd, l:path)
```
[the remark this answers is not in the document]
przepompownia: I needed a while to remember its origin after week. 

It came directly from one of test cases that did not pass. I have explained it in https://github.com/phpactor/phpactor/pull/923#issuecomment-600884367.

dantleech:
@przepompownia happy to merge this, or did you want to add some tests?

autoload/phpactor.vim
```diff
@@ -435,14 +443,53 @@ endfunction
 " RPC -->-->-->-->-->--
 """""""""""""""""""""""
 
+function! s:searchDirectoryUpwardForRootPatterns(initialDirectory, workspaceRootPatterns, fallbackDirectory)
+  if index(g:phpactorGlobalRootPatterns, '/') < 0
+    call add(g:phpactorGlobalRootPatterns, '/')
+  endif
```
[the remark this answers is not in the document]
przepompownia: I have explained it in doc comment added in ftplugin. It forces excluding system root directory `/` from treating as workspace root.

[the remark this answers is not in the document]
przepompownia: Technically, it checks if `g:phpactorGlobalRootPatterns` contains `/` (it may be not added by user). If not, it adds the system root to that list.

przepompownia:
> @przepompownia happy to merge this, or did you want to add some tests?

I'm afraid I may not have enough time, especially if I wanted to find a way to test local functions (like `s:searchDirectoryUpwardForRootPatterns`).

dantleech:
ok -- no worries. Let's see how it goes

Issue #941, opened 2020-04-06
Incorrect project root detection


dantleech:
```
2020-04-06 09:05:46] phpactor.DEBUG: Resolved path "%application_root%/vendor/jetbrains/phpstorm-stubs" to "/home/daniel/www/phpactor/phpactor/vendor/jetbrains/phpstorm-stubs" [] []
[2020-04-06 09:05:46] phpactor.DEBUG: Resolved path "%cache%/worse-reflection" to "/home/daniel/.cache/phpactor/worse-reflection" [] []
[2020-04-06 09:05:46] phpactor.DEBUG: Resolved path "%project_root%/vendor/autoload.php" to "/home/daniel/www/inviqa/lautsprecher-teufel/vendor/spryker-shop/catalog-page/vendor/autoload.php" [] []
[2020-04-06 09:05:46] phpactor.WARNING: Could not find autoloader "/home/daniel/www/inviqa/lautsprecher-teufel/vendor/spryker-shop/catalog-page/vendor/autoload.php" [] []
[2020-04-06 09:05:46] phpactor.DEBUG: Resolved path "%project_root%" to "/home/daniel/www/inviqa/lautsprecher-teufel/vendor/spryker-shop/catalog-page" [] []
[2020-04-06 09:05:46] phpactor.DEBUG: Resolved path "%cache%/index/%project_id%" to "/home/daniel/.cache/phpactor/index/catalog-page-f0ca3b" [] []
```

The vendor dir was resolved when trying to goto implemtnations instead of the project root.

The defaults root patterns are:

```
let g:phpactorProjectRootPatterns = get(g:, 'phpactorProjectRootPatterns', ['composer.json', '.git', '.phpactor.json', '.phpactor.yml'])
```

All of these could be matched in a vendor dir.

Any operation Phpactor attempts in third-party packages will likely not work. cc @przepompownia 

Resolved by locally setting:

```
let g:phpactorProjectRootPatterns = []
```

przepompownia:
Right, detecting by patterns that are present at the top of vendor packages does not make its job. 

We have not available the third method: manual selecting the root of the project (the first method does it implicitly by setting current working directory before calling to phpactor). It's advantage is avoiding a situation like presented above. It has also drawback if it bases on one global variable per whole vim process in the case when we need to have work on more than one project. 

Proposition: have more than one project root and assign each buffer handled by phpactor to exactly one of them. It implicitly would define "project" described only by its root at this moment. Probably assumptions:
[...]

PR #959, opened 2020-04-13
[WIP] Associate each php buffer with some project

przepompownia:
- define project by its primary root and relation of belonging a file to it
- define project repository
- define buffer matcher: if  given file does not belong to any project, ask user how to create the new project for this file


przepompownia:
Related to #941

przepompownia:
A this stage the result variable (`b:project`) is not used in `phpactor#rpc` but we can test how project root can be associated with the buffer - both by echoing `b:project` and executing `:PhpactorListProjects`.

My current instance of neovim has a bug with `inputlist()` - the user prompt does not appear and value is assigned noninteractively. I wanted to use any dialog function from phpactor but did not find anyone that returns selected  value. 

`s:matchFileToProject` requires rewrite to more clear form ;)

There are still no test that simulate user interaction.

ftplugin/php.vim
```diff
@@ -54,5 +54,13 @@ if g:phpactorOmniAutoClassImport == v:true
...
 endif
 
+let g:phpactorBufferMatcher = get(g:, 'phpactorBufferMatcher', phpactor#project#bufferMatcher#create(
+            \ phpactor#project#projectRepository#create(),
+            \ g:phpactorProjectRootPatterns,
+            \ g:phpactorGlobalRootPatterns,
+            \ g:phpactorInitialCwd
+            \ ))
+
+command PhpactorListProjects echo keys(map(g:phpactorBufferMatcher.getRepository().listProjects(), { key, val -> key}))
```
dantleech, line 64: should be in `ftplugin/php/commands.vim` :)

dantleech:
Looks good, I'll take a closer look tomorrow :)

przepompownia:
The current vim version used in Travis test does not support `v:t_dict` variable. At the moment I can replace it with `type({})`.

What do you think about switching to Vim 8, which can be needed once asynchronous calling to RPC were implemented?

autoload/phpactor/project/bufferMatcher.vim
```diff
@@ -0,0 +1,129 @@
...
+  let l:project = self.repository.findProjectContainingFile(a:file)
+
+  if type(l:project) == type({})
+    return l:project
+  endif
+
+  " early version to refactor or change
+  " heredoc are available since vim 8.1.1354
+  let l:message = printf(join([
+        \ 'Phpactor [RPC]',
```
elythyr, line 47: Is this related only to the RPC ?
If so I would propose to add it to the namespace so it's clear event by looking at the filsystem.
`autoload/phpactor/rpc/project/bufferMatcher.vim` maybe ?

autoload/phpactor/project/bufferMatcher.vim
```diff
@@ -0,0 +1,129 @@
...
+          \ printf('2. use initial working directory: "%s"', self.initialCwd)
+          \ ]
+
+  if v:null != l:rootDirByMarker
+    let l:item = '3. autodetected by root marker: '.l:rootDirByMarker
+    call add(l:choices, l:item)
+  endif
+
+  let l:choice = v:null
+  while index([0,1,2,3], l:choice) < 0
```
elythyr, line 73: It might be better to make it dynamic:
```suggestion
  while index(range(1, len(l:choices) - 1), l:choice) < 0
```

autoload/phpactor/project/bufferMatcher.vim
```diff
@@ -0,0 +1,129 @@
...
+    " bug: neovim assigns 0 without any interaction, so I used input()
+    let l:choice = inputlist(l:choices)
+    " let l:choice = str2nr(input(l:choices[0], join(l:choices[1:], "\n")))
+    " echo join(l:choices)."\n"
+    " let l:choice = str2nr(input('Enter number: '))
+    redraw
+  endwhile
+
+  if l:choice == 1
```
elythyr, line 82: What about having a list of function for each choices so we can dynamically call them and don't care in which order we previously define them ?
The goal is really to limit as much as possible the number of changes we will need to make when maintaining this code.
```vim
let l:actions = [
  \ { -> self.initialCwd },
  \ function('s:letTheUserChoose'),
  \ { -> self.initialCwd },
  \ { -> l:rootDirByMarker },
\ ]

let l:choice = 0
while
...
endwhile

call l:actions[l:choice]
```

autoload/phpactor/project/bufferMatcher.vim
```diff
@@ -0,0 +1,129 @@
...
+    let l:manualPath = input('Enter file path: ', l:initialDirectory, 'file')
+    redraw
+    if ! empty(glob(l:manualPath.'/'))
+      " todo validate path
+      let l:selectedDir = l:manualPath
+    endif
```
elythyr, line 88: What about asking until a valid path is given ?
In case the user made a typo and the path is invalid, we ask him again with the value he just gave us as default.
```suggestion
    let l:selectedDir = s:askForAPath(l:initialDirectory)
```
```vim
function! s:askForAPath(defaultValue) abort
  let l:manualPath = input('Enter file path: ', a:defaultValue, 'file')
  redraw

  if !empty(glob(l:manualPath.'/'))
    return l:manualPath
  endif

  " We might be able to echo here in order to let the user know that the path is invalid
  " If it does not work or is not clean enought, add a flag as parameter or the promp directly
  " And add a short message like "Invalid path\n" to preprend to the prompt
  return s:askPath(l:manualPath)
endfunction
```

autoload/phpactor/project/bufferMatcher.vim
```diff
@@ -0,0 +1,129 @@
...
+  if index(l:filesystemRootMarkers, '/') < 0
+    call add(l:filesystemRootMarkers, '/')
+  endif
+
+  return {
+        \ 'initialCwd': a:initialCwd,
+        \ 'projectRootMarkers': a:projectRootMarkers,
+        \ 'filesystemRootMarkers': l:filesystemRootMarkers,
+        \ 'repository': a:projectRepository,
+        \ 'getRepository': function('s:getRepository'),
```
elythyr, line 32: Without a private visibility it's a bit redundant, we might consider removing one of them ?

autoload/phpactor/project/bufferMatcher.vim
```diff
@@ -0,0 +1,129 @@
...
+  while index(a:filesystemRootMarkers, l:directory) < 0
+    if s:directoryMatchesToPatterns(l:directory, a:workspaceRootMarkers)
+      return l:directory
+    endif
+
+    let l:directory = fnamemodify(l:directory, ':h')
+  endwhile
+
+  return v:null
+endfunction
```
elythyr, line 119: It's a matter of taste and habit but personally I prefer to use recursion in those cases.
I find it more readable and easier to understand than a loop with break points in it.
Resolve if you don't like it, it's purely syntactic :)
You could even regroup the two conditions and gain 3 lines but I prefer readability.

```vim
function! s:searchDirectoryUpwardForRootMarkers(directory, workspaceRootMarkers, filesystemRootMarkers) abort
  if 0 <= index(a:filesystemRootMarkers, a:directory)
    return a:directory
  endif

  if s:directoryMatchesToPatterns(a:directory, a:workspaceRootMarkers)
    return a:directory
  endif

  return s:searchDirectoryUpwardForRootMarkers(
    \ fnamemodify(l:directory, ':h'),
    \ a:workspaceRootMarkers
    \ a:filesystemRootMarkers
  \ )
endfunction
```

autoload/phpactor/project/bufferMatcher.vim
```diff
@@ -0,0 +1,129 @@
...
+
+function s:directoryMatchesToPatterns(directory, patterns) abort
```
elythyr, line 121: Just in case, you might want to take a look at `finddir()` and `findfile()`, I don't know if we can use them because I'm not sure about all the rules we want to enforce with the filesystem markers and all.
But it's generally better for performances.

autoload/phpactor/project/project.vim
```diff
@@ -0,0 +1,39 @@
+function! phpactor#project#project#createFromRootPath(primaryRootPath) abort
+  let l:project = {
+        \ 'primaryRootPath': s:normalizeRootPath(a:primaryRootPath),
+        \ 'getPrimaryRootPath': function('s:getPrimaryRootPath'),
```
elythyr, line 4: As previously mention, a bit redundant but it might be just me.

autoload/phpactor/project/project.vim
```diff
@@ -0,0 +1,39 @@
...
+        \ }
+
+  return l:project
+endfunction
+
+function! s:getPrimaryRootPath() dict
+  return l:self.primaryRootPath
+endfunction
+
+function! s:containsFile(filename) dict
```
elythyr, line 15: `finddir()` and `filepath()` might be used here instead if we have full path and not just a filename as parameter.
If so maybe rename it at the same time to avoid confusion ;)

autoload/phpactor/project/bufferMatcher.vim
przepompownia: Yes, it is related only to used with RPC.  The whole phpactor code written in viml seems to be also related to RPC only ;) 

Most of this mechanism is as much phpactor independent as posible. It makes it easier to test and attach to or remove from phpactor. Maybe it would support root detection in any LSP client but I still have no knowledge how similar problems are resolved on both sides of LSP.

autoload/phpactor/project/bufferMatcher.vim
przepompownia: At the moment it is better. I hope that it will be possible to get rid of these internals for any better input list implementation.

autoload/phpactor/project/project.vim
```diff
@@ -0,0 +1,39 @@
...
+  endwhile
+
+  return v:true
+endfunction
+
+function s:normalizeRootPath(path) abort
+  let l:path = simplify(fnamemodify(a:path, ':p'))
+
+  " @todo better check if it is not an existing directory
+  if empty(glob(l:path.'/'))
```
elythyr, line 34: You got `isdirectory()` to help with that ;)

autoload/phpactor/project/bufferMatcher.vim
przepompownia: [Tlib](https://github.com/vim-scripts/tlib/blob/master/autoload/tlib/input.vim) input list can be good example, fzf maybe also, but we have good piece of code at the phpactor side that to reuse after small modification.

autoload/phpactor/project/projectRepository.vim
```diff
@@ -0,0 +1,47 @@
...
+        \ 'addProject': function('s:addProject'),
+        \ 'findProjectContainingFile': function('s:findProjectContainingFile'),
+        \ 'listProjects': function('s:listProjects'),
+        \ 'hasProjectWithPrimaryRoot': function('s:hasProjectWithPrimaryRoot'),
+        \ 'hasProject': function('s:hasProject')
+        \ }
+endfunction
+
+function! s:listProjects() dict abort
+  return copy(self.projects)
```
elythyr, line 14: I guess the idea was better safe than sorry with the `copy` but we can access it directly through the dictionary.
So could we imagine let the user the care to copy the projects before playing with them in order to avoid making copies every time the method is used (even if it's only to read) ?
And while at it, duplicate with `dict.projects`, choose one of the two if we decide to optimize and remove the `copy`.

ftplugin/php/commands.vim
```diff
@@ -128,3 +128,8 @@ command! -nargs=0 PhpactorGenerateAccessors call phpactor#GenerateAccessors()
 ""
 " Automatically add any missing properties to a class
 command! -nargs=0 PhpactorTransform call phpactor#Transform()
+
+""
+" List projects existing in this vim instance
+" (currently each project consist of its root directory only)
+command PhpactorListProjects echo keys(map(g:phpactorBufferMatcher.getRepository().listProjects(), { key, val -> key}))
```
dantleech, line 135: `command!`

autoload/phpactor/project/bufferMatcher.vim
przepompownia: Right: all fields are still public inevitably... The first line is needed for construction so we can get rid of the second (and replace references).

autoload/phpactor/project/projectRepository.vim
```diff
@@ -0,0 +1,47 @@
...
+    return v:null
+endfunction
+
+function! s:addProject(project) dict abort
+  " todo check type
+  if v:null isnot get(self.projects, a:project.getPrimaryRootPath(), v:null)
+    throw printf('Project "%s" already exists', a:project.getPrimaryRootPath())
+  endif
+
+  let self.projects[a:project.getPrimaryRootPath()] = a:project
```
elythyr, line 34: Given a project does not contain any data it's on no consequences if I try to register the same project twice, isn't it ?
If I'm not mistaken then I propose to fail silently to prevent user experience:
```suggestion
  " todo check type, here you could throw is the structure is not right, all keys must exist

  if get(self.projects, a:project.getPrimaryRootPath(), v:null) is v:null
    let self.projects[a:project.getPrimaryRootPath()] = a:project
  endif
```

plugin/phpactor.vim
```diff
@@ -0,0 +1,4 @@
+augroup phpactor
+  autocmd!
+  autocmd FileType php call phpactor#project#bufferMatcher#assignFileToProject(expand('<afile>'))
+augroup END
```
dantleech, line 4: Should this be moved to the `ftplugin` folder?

tests/VimPlugin/project/find_project.vader
```diff
@@ -0,0 +1,28 @@
...
+Execute (Ensure that the repository is empty):
+  Assert g:phpactorProjectRepository.isEmpty()
+
+Then (Find the project in an empty repository):
+  let rootPath = fnamemodify(g:projectDirs['SillyWalksProject'], ':p')
+  let file = rootPath . '/foo.strangeext'
+
+  let project = g:phpactorProjectRepository.findProjectContainingFile(file)
+
+  AssertEqual v:null, project
```
dantleech, line 12: Should this really be null? Maybe change the test description? (e.g. `Returns null if the repository is empty`)

autoload/phpactor/project/project.vim
przepompownia: I suspected that there is such function but I overlooked it browsing `help file-functions`. 

:+1:

autoload/phpactor/project/projectRepository.vim
```diff
@@ -0,0 +1,47 @@
...
+endfunction
+
+function! s:hasProject(project) dict abort
+  return get(self.projects, a:project.getPrimaryRootPath(), v:false) isnot v:false
```
elythyr, line 38: Here you don't have to check the all structure but at least that the function exist and is a function.
But in order to avoid repeating the same checks and to keep the knowledge at the same place I encourage your to create a function `phpactor#project#project#isValid(project)` that will do all the necessary tests.

tests/VimPlugin/project/find_project.vader
```diff
@@ -0,0 +1,28 @@
...
+
+Execute (Check if there is a project containing the file):
```
dantleech, line 14: `Returns project for a project file`?

tests/VimPlugin/project/find_project.vader
```diff
@@ -0,0 +1,28 @@
...
+
+  call g:phpactorProjectRepository.addProject(existingProject)
+
+  let project = g:phpactorProjectRepository.findProjectContainingFile(file)
+
+  AssertEqual existingProject, project
+
+  AssertEqual type({}), type(project)
+
+Execute (Ensure both projects have the correct type):
```
dantleech, line 25: Both projects?

plugin/phpactor.vim
```diff
@@ -0,0 +1,4 @@
+augroup phpactor
+  autocmd!
+  autocmd FileType php call phpactor#project#bufferMatcher#assignFileToProject(expand('<afile>'))
```
elythyr, line 3: Why not put this in a filetype plugin file instead of using an auto command ?

tests/VimPlugin/project/match_project.vader
```diff
@@ -0,0 +1,12 @@
+Include (Prepare project directories): lib/create_fixtures_directory.vader
```
dantleech, line 1: Is this test the same as `find_project.vader`?

ftplugin/php.vim
elythyr: Don't forget to make it a buffer command: `:h :command-buffer`

autoload/phpactor/project/projectRepository.vim
przepompownia: We can copy projects dictionary within `PhpactorListProjects` implementation instead.

autoload/phpactor/project/bufferMatcher.vim
elythyr: >  The whole phpactor code written in viml seems to be also related to RPC only ;)
:D That's a damn valid point !

autoload/phpactor/project/bufferMatcher.vim
elythyr: Also @dantleech prefer to avoid coupling with any other plugin or library.
And it's a good decision, so I would rather see us defining an interface contract.
I think there is already something close to that in `phpactor#input` for the lists.
Then in our side we only deal with what vim offer us `input`, `inputlist`, `confirm` and all that.
It's uggly and all but it's the same for anyone no matter what.

Then we could make some external plugins like `phpactor-vim-list-fzf`, that will only change the configuration variable so that phpactor vimscript will use it's own implementation for the list.

autoload/phpactor/project/bufferMatcher.vim
```diff
@@ -0,0 +1,129 @@
+function! phpactor#project#bufferMatcher#assignFileToProject(filename)
```
dantleech, line 1: Rename to `assignProjectToBuffer`?

plugin/phpactor.vim
przepompownia: `ftplugin/php.vim` is prevented from be sourced more than once. 

Maybe we can put it in another ftplugin file. 

Thanks to your above suggestion I discovered that `ftplugin/php/commands.vim` has no such prevention (commands are redefined each time).

autoload/phpactor/project/bufferMatcher.vim
```diff
@@ -0,0 +1,129 @@
...
+        \ 'repository': a:projectRepository,
+        \ 'getRepository': function('s:getRepository'),
+        \ 'matchFileToProject': function('s:matchFileToProject')
+        \ }
+endfunction
+
+function s:matchFileToProject(file) dict abort
+  let l:project = self.repository.findProjectContainingFile(a:file)
+
+  if type(l:project) == type({})
```
dantleech, line 40: What does `type({})` mean?

autoload/phpactor/project/project.vim
elythyr: If you got `fzf` on your compture and `fzf.vim` I think, you got a great `:Helptags` command to fuzzy find stuff in the doc :)

autoload/phpactor/project/bufferMatcher.vim
```diff
@@ -0,0 +1,129 @@
...
+  else
+    let l:selectedDir = self.initialCwd
+  endif
+
+  if v:null != l:selectedDir
+    let l:project = phpactor#project#project#createFromRootPath(l:selectedDir)
+    echomsg printf('Project with root "%s" has been created.', l:project.getPrimaryRootPath())
+
+    return l:project
+  endif
```
dantleech, line 100: maybe explicitly return `v:null` here? and early return:

```
if (v:null === l:selectedDir) {
    return v:null;
}

// happy path
```

autoload/phpactor/project/bufferMatcher.vim
```diff
@@ -0,0 +1,129 @@
...
+        \ 'initialCwd': a:initialCwd,
+        \ 'projectRootMarkers': a:projectRootMarkers,
+        \ 'filesystemRootMarkers': l:filesystemRootMarkers,
+        \ 'repository': a:projectRepository,
+        \ 'getRepository': function('s:getRepository'),
+        \ 'matchFileToProject': function('s:matchFileToProject')
+        \ }
+endfunction
+
+function s:matchFileToProject(file) dict abort
```
dantleech, line 37: rename to `resolveProjectForFile` ?

plugin/phpactor.vim
dantleech: > ftplugin/php.vim is prevented from be sourced more than once.

Why do we care about it being resourced? btw I use https://github.com/tpope/vim-scriptease to  resource plugins in development `:Runtime!`

przepompownia:
> Here are some feedback, don't hesitate to to say it if you feel I'm wrong.
> Especially when it comes to the logic because I still don't really get what we are trying to do :)

I would like to make RPC call with the correct working directory to avoid errors related to incorrect choice. 

I am going to write some test that simulates user experience (if it is possible to simulate an input). Maybe such tests will be more clear than my English ;)

Thanks for all comments, more are welcome :)

autoload/phpactor/project/bufferMatcher.vim
dantleech: We do of course have our `phpactor#input#` namespace...

autoload/phpactor/project/projectRepository.vim
elythyr: I didn't look much at the command to be honest and now that I do I don't get it, the `map()` seems of no use.
And returning only the list of paths doesn't seem that useful.
I would rather remove it for now and wait for a need to emerge before implementing it: [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)

And even we keep it there is no need to copy it.
The all point of making a copy is to make sure that we don't modify the projects we got internally, but this command only do "read actions".

autoload/phpactor/project/bufferMatcher.vim
dantleech: I think eventually people will be able to use either LSP or RPC and have the same feature set - will have to see what the trade offs and advanrages are in the long run. RPC is obviously much lighter and could eventually properly use the indexer too.

plugin/phpactor.vim
przepompownia: IMHO, part of mentioned above command definitions (as `PhpactorClassNew`, with given project root, or `PhpactorExtensionInstall`) would be moved from ftplugin to plugin.

dantleech:
> What do you think about switching to Vim 8

I think it's a good idea :)

plugin/phpactor.vim
elythyr: > ftplugin/php.vim is prevented from be sourced more than once.

I don't get that, it will be executed for each time a buffer filetype is defined as `php`.

> Thanks to your above suggestion

I think you refer to the `buffer` part for commands.
We should try to only impact one buffer at a time since we only care for php files (well behat too but that's not the point).

przepompownia:
@elythyr, @dantleech  did you test inputlist on neovim? I still am not prompted by inputlist on neovim (nightly release). 

`nvim /tmp/anything.php` should prompt for project root choice.

autoload/phpactor/project/bufferMatcher.vim
elythyr: It gave you the type number for a dictionary.
The constants `v:t_<type>` didn't always exists, fairly new even.
So that's how you can do it to be compatible with older versions:

elythyr:
> > What do you think about switching to Vim 8
> 
> I think it's a good idea :)

Switch to NeoVim <3

elythyr:
> @elythyr, @dantleech did you test inputlist on neovim? I still am not prompted by inputlist on neovim (nightly release).
> 
> `nvim /tmp/anything.php` should prompt for project root choice.

Didn't test it at all, just review the code and I think that will be it for this evening ;)

przepompownia:
> > What do you think about switching to Vim 8
> 
> I think it's a good idea :)


Using relatively fresh vim and nightly neovim I do not realize what versions we can expect from a common user but it seems that we can assume to require at least asynchroneous job support.

dantleech:
>  I do not realize what versions we can expect from a common user but it seems that we can assume to require at least asynchroneous job support.

At least VIM 8 should be fine, if anybodies using VIM 7 chances are they are also not running PHP 7.3 (the minimum supported version for the next release)

przepompownia:
@elythyr instead of switching, we should make tests both on vim and neovim.

autoload/phpactor/project/bufferMatcher.vim
dantleech: would it be better to check for `v:null`? `if (v:null !== l:project` ?

autoload/phpactor/project/project.vim
przepompownia: I am used to browse command or buffers history, buffers (using shortcuts) but forget about `:Helptags` ;)

przepompownia:
> I think it makes sense.
> 
> I can't get this to work however - my `b:project` seems always to be set to my initial CWD.

@dantleech does it occur also on Vim? I have this issue on neovim (`inputlist()` does not prompt).

dantleech:
Hmm, we use inputlist in other places without issue

On 14 April 2020 22:17:12 CEST, Tomasz N <notifications@github.com> wrote:
>> I think it makes sense.
>> 
>> I can't get this to work however - my `b:project` seems always to be
>set to my initial CWD.
>
>@dantleech does it occur also on Vim? I have this issue on neovim
>(`inputlist()` does not prompt).
>
>-- 
>You are receiving this because you were mentioned.
>https://github.com/phpactor/phpactor/pull/959#issuecomment-613659137

-- 

autoload/phpactor/project/bufferMatcher.vim
przepompownia: After looking how the upward search works, I would like to use it with tests.

przepompownia:
> Hmm, we use inputlist in other places without issue

It works if I put `echoerr` somewhere before ;)

autoload/phpactor/project/bufferMatcher.vim
przepompownia: Try `echo v:null != {}` (E735: Can only compare Dictionary with Dictionary). 

`isnot` can be used instead to compare with `v:null`. 

In some case of badly implemented function that should return `v:null` or a dictionary I got a number for example. Asking directly for dictionary type seems to be more accurate than comparing with `v:null` (although we can still get completely different object than we expect).

plugin/phpactor.vim
przepompownia: It seems to work.

autoload/phpactor/project/bufferMatcher.vim
przepompownia: ProjectAssigner instead of BufferMatcher?

tests/VimPlugin/project/match_project.vader
przepompownia: Indeed, both should have more expressive names. 
- `find_project` is a part of repository test, 
- `match_project` is a part bufferMatcher test.

tests/VimPlugin/project/find_project.vader
przepompownia: It was remains after removed assertion.

autoload/phpactor/project/projectRepository.vim
przepompownia: Maybe my need for ensure what projects already exist is more specific for development than normal usage ;)

przepompownia:
To refer to the rest of suggestions, especially those less obvious, I need the next portion of free time.

autoload/phpactor/project/bufferMatcher.vim
dantleech: What about `if !empty(l:project)`?

autoload/phpactor/project/bufferMatcher.vim
dantleech: or just `assigner`? as we have `project` in the namespace?

dantleech: `#phpactor#project#assigner#assignBufferToProject`?

autoload/phpactor/project/projectRepository.vim
dantleech: We could also imagine expanding this to "switch" projects (in the future, better to keep this PR small :) )

autoload/phpactor/project/bufferMatcher.vim
przepompownia: Similarly as above: it can fail if potentially badly implemented `self.repository.findProjectContainingFile()` returned non-zero number for example.

dantleech: but seems that accepting any dict as the result is effectively as bad (it doesn't make us safer). Who will implement these functions? I think as a general approach I'd rather have `if something was returned` logic than `if returned value is of the expected type, return` (if nothing else because it's easier to understand) ?

plugin/phpactor.vim
elythyr: https://github.com/phpactor/phpactor/blob/faecc433e98060dba1641fefa712eb3f8ceb43ef/ftplugin/php.vim#L1-L3

That's why it can be sourced only once.
I made a PR to clean the filetype plugin a bit:
https://github.com/phpactor/phpactor/pull/964

autoload/phpactor/project/bufferMatcher.vim
elythyr: It's not like if someone else is going to implement this, I think the test here is not necessary.
The repository is an internal dictionary if we messed up the implementation of the function we don't care to have one more error here.
I would even say that I prefer to have as many error as possible in order to be sure to see them.
In PHP it would be a `LogicException`, you don't want to hide those otherwise you're not able to know that something went wrong.

autoload/phpactor/project/bufferMatcher.vim
```diff
@@ -57,39 +57,33 @@ function s:matchFileToProject(file) dict abort
...
-    if isdirectory(l:manualPath)
-      " todo validate path
-      let l:selectedDir = l:manualPath
-    endif
-  elseif l:choice == 3
-    let l:selectedDir = l:rootDirByMarker
-  else
-    let l:selectedDir = self.initialCwd
-  endif
+  let l:selectedDir = l:choices[l:choice]['action']()
```
elythyr: This will break in case of interruption with `<C-c>`, you should initialize `l:choice` to the default value.
Both on line 79 when creating the variable and between line 80&81 to reset it in case of wrong choices.
Otherwise if I make an invalid choice and I interrupt it `l:choice` will still contains the invalid choice.

autoload/phpactor/project/bufferMatcher.vim
```diff
@@ -57,39 +57,33 @@ function s:matchFileToProject(file) dict abort
...
   endif
 
   let l:choice = v:null
-  while index(range(0, len(l:choices)), l:choice) < 0
-    " bug: neovim assigns 0 without any interaction, so I used input()
-    let l:choice = inputlist(l:choices)
-    " let l:choice = str2nr(input(l:choices[0], join(l:choices[1:], "\n")))
-    " echo join(l:choices)."\n"
-    " let l:choice = str2nr(input('Enter number: '))
+  while index(range(0, len(l:choices)-1), l:choice) < 0
```
elythyr:
```suggestion
  while index(range(1, len(l:choices)-1), l:choice) < 0
```
I would reject `0` since it's the prompt message.

autoload/phpactor/project/bufferMatcher.vim
```diff
@@ -57,39 +57,33 @@ function s:matchFileToProject(file) dict abort
...
+    let l:choice = inputlist(map(copy(l:choices), { number, item -> (number ? printf('%d: ', number) : '') . item['message'] }))
```
elythyr:
```suggestion
    let l:choice = inputlist(map(copy(l:choices), {
        \ number, item -> (number ? printf('%d. ', number) : '') . item['message']
    \ }))
```
Just to be easier to read and I put back the dot because I like it better :)

autoload/phpactor/project/projectRepository.vim
```diff
@@ -0,0 +1,42 @@
+function! phpactor#project#projectRepository#create() abort
```
elythyr, line 1: What about renaming the file in order to have a function like `phpactor#project#repository#create()`.
IMO repeating the `project` is unnecessary and disgracious :)

ftplugin/php/autocommands.vim
```diff
@@ -0,0 +1 @@
+call phpactor#project#assigner#assignProjectToBuffer(expand('%'))
```
elythyr, line 1: It's not an auto command, wrong file ?

autoload/phpactor/project/assigner.vim
```diff
@@ -0,0 +1,118 @@
...
+  if v:null != l:rootDirByMarker
+    let l:item = {
+          \ 'action': { -> l:rootDirByMarker },
+          \ 'message': 'autodetected by root marker: '.l:rootDirByMarker
+          \ }
+    call add(l:choices, l:item)
+  endif
+
+  let l:choice = v:null
+  while index(range(0, len(l:choices)-1), l:choice) < 0
```
elythyr, line 80:
```suggestion
  while index(range(1, len(l:choices)-1), l:choice) < 0
```
I would ignore `0` since it's the prompt I feel like it should not be a valid choice

autoload/phpactor/project/assigner.vim
```diff
@@ -0,0 +1,118 @@
...
+
+  if v:null != l:rootDirByMarker
+    let l:item = {
+          \ 'action': { -> l:rootDirByMarker },
+          \ 'message': 'autodetected by root marker: '.l:rootDirByMarker
+          \ }
+    call add(l:choices, l:item)
+  endif
+
+  let l:choice = v:null
```
elythyr, line 79: You should initialize it to the default value:
```suggestion
  let l:choice = 0
```

autoload/phpactor/project/assigner.vim
```diff
@@ -0,0 +1,118 @@
...
+  while index(range(0, len(l:choices)-1), l:choice) < 0
+    let l:choice = inputlist(map(copy(l:choices), { number, item -> (number ? printf('%d: ', number) : '') . item['message'] }))
```
elythyr, line 81: We also need to reset `l:choice` in case the value was out of range, otherwise if I select 5 then hit CTRL-C we will have an invalid index when calling the action:
```suggestion
    let l:choice = 0 " Reset to default to always have a valid value in case of interruption
    let l:choice = inputlist(map(copy(l:choices), { number, item -> (number ? printf('%d: ', number) : '') . item['message'] }))
```

autoload/phpactor/project/assigner.vim
```diff
@@ -0,0 +1,118 @@
...
+        \ self.filesystemRootMarkers
+        \ )
+
+  let l:choices = [
+        \ {
+        \ 'action': { -> self.initialCwd },
+        \ 'message': l:message
+        \ },
+        \ {
+        \ 'action': function('input', ['Enter file path: ', l:initialDirectory, 'file']),
```
elythyr, line 66: So you don't want to ask again if it's not a valid path ?
Typo are common even with completion I think we should help the user as much as possible.

autoload/phpactor/project/assigner.vim
```diff
@@ -0,0 +1,118 @@
...
+        \ }
+        \ ]
+
+  if v:null != l:rootDirByMarker
+    let l:item = {
+          \ 'action': { -> l:rootDirByMarker },
+          \ 'message': 'autodetected by root marker: '.l:rootDirByMarker
+          \ }
+    call add(l:choices, l:item)
+  endif
```
elythyr, line 77: Should not we fallback to `self.initialCwd` if we don't have a valid path ?

autoload/phpactor/project/assigner.vim
przepompownia: Did you try it along with `0` as the start value (which is btw the good idea)? After press Enter (default value) `l:choices` becomes `0` and the loop requires to be break.

przepompownia: Behaviour after pressing `Ctrl-C` requires more attention. In the current version it causes 
```
"/tmp/x2.php" [New File]                                                                                                                                                                                                                        
Phpactor [RPC]                                                                                                                                                                                                                                  
There is no project enabled for "/tmp/x2.php" yet.                                                                                                                                                                                              
Select the way of assign the project root                                                                                                                                                                                                       
If cancelled, "/home/_/.vim/pack/bundle/start/phpactor" will be used.                                                                                                                                                                      
1: manual (default "/tmp")                                                                                                                                                                                                                      
Type number and <Enter> or click with mouse (empty cancels):                                                                                                                                                                                    
Error detected while processing /home/_/.vim/pack/bundle/start/phpactor/ftplugin/php/autocommands.vim:                                                                                                                                     
line    1:                                                                                                                                                                                                                                      
Interrupted                                                                                                                                                                                                                                     
Error detected while processing /home/_/.vim/pack/bundle/start/phpactor/ftplugin/php/commands.vim:                                                                                                                                         
line    1:                                                                                                                                                                                                                                      
Interrupted                                                                                                                                                                                                                                     
Error detected while processing /home/_/.vim/pack/bundle/start/phpactor/ftplugin/php/mappings.vim:                                                                                                                                         
line    1:                                                                                                                                                                                                                                      
Interrupted                                                                                                                                                                                                                                     
Error detected while processing /home/_/.vim/pack/bundle/opt/_/ftplugin/php.vim:                                                                                                                                                      
line    1:                                                                                                                                                                                                                                      
Interrupted                                                                                                                                                                                                                                     
Error detected while processing /usr/share/vim/vim82/ftplugin/php.vim:                                                                                                                                                                          
line    1:                                                                                                                                                                                                                                      
Interrupted                                                                                                                                                                                                                                     
Interrupt: Press ENTER or type command to continue 
```
After moving calling back from ftplugin to autocommand, the error does not occur, but probably the rest of code is not executed (throwing anything gives no effect).

przepompownia: I still have no idea how to correctly handle keyboard interruption (in viml).

przepompownia: Outdated: `Interrupted` is an exception message, probably caught somewhere, so it did not look like an exception at first glance ;)

ftplugin/php/autocommands.vim
przepompownia: It is wrong but let wait to finish #965

przepompownia:
https://github.com/neovim/neovim/issues/12144 - it is not a bug.

I have this issue due to `F` added by default to `shortmess` in neovim. 
It gives similar efect like `silent call inputlist([1,2,3])`.

We then have at least two choices:
  - left the user the default choice. We ca also inform the user that it 
can be changed - I added `allowReassign` argument to 
`phpactor#project#assigner#assignProjectToBuffer()`,
  - encourage the user to `set shortmess-=F` (someother messages and 
interactions can be then unhide).

I do not see any better choice at the moment. <https://github.com/dantleech>

autoload/phpactor/project/assigner.vim
```diff
@@ -1,9 +1,9 @@
-function! phpactor#project#assigner#assignProjectToBuffer(filename)
+function! phpactor#project#assigner#assignProjectToBuffer(filename, allowReassign)
   if !exists('g:phpactorLoaded')
     return
   endif
 
-  if exists('b:project')
+  if exists('b:project') && (v:false is a:allowReassign)
```
elythyr, line 6: How can a buffer need to change project ?
If it was moved then is it still the same buffer ?
`filename`  will be different I think, not sure about what I'm saying so it needs to be checked.
If a file can be moved on the filesystem without changing buffer then we might want to detect that instead of providing a flag. Maybe stocking the filename in buffer variable and checking it against `a:filename`.

Talking about buffer variable make me wonder if we would not need a more unique name than `b:buffer`, it's a shared namespace so we should include `phpactor` in it at least. To limit the risk of collisions with other plugin.

przepompownia: I agree that this variable should be removed to something containing "namespace". 

At the moment the only reason to change is [a choice hidden from the user](https://github.com/phpactor/phpactor/pull/959#issuecomment-615840400): `inputlist()` non-interactively returns its default value so there is need to change it later.

elythyr: I didn't saw your previous explanation, sorry.

So for me the second option is a no go, we can't ask people to configure their editor so that they can use our tool, it's against the purpose IMO.

For the first option, as long as the default value used is configurable I don't see the problem.
And as I said I didn't do much except reading your code until know.
Stupidly just to see what was going on without try to understand how it will work.
One reason for that is that I never thought the user will be impacted.
If we need to handle some kind of project files grouping to be able to do our work: fine.
But the user should not be bother with that outside of configuration.
Then I start to think that what you did will ask the user in which project the file should go ?
I hope I mistaken because:
```gherkin
As a user
Given I use a tool to improve my editor
When this tool makes my editor ask me a question every time I open a file
Then I burn down the tool and send it to hell with a voodoo dance!
```

autoload/phpactor/fileutils.vim
```diff
@@ -0,0 +1,47 @@
...
+    if empty(a:forbiddenDirs) || ! phpactor#fileutils#containsDirectoryFrom(l:markerPath, a:forbiddenDirs)
+      return l:markerPath
+    endif
+  endfor
+
+  return v:null
+endfunction
+
+function! phpactor#fileutils#normalizePath(path) abort
+  let l:modifiers = isdirectory(a:path) ? ':p:h' : ':p'
```
elythyr, line 24: Should it not be ?
```suggestion
  let l:modifiers = !isdirectory(a:path) ? ':p:h' : ':p'
```

Otherwise I don't get it.
Also the name is confusing because you're doing things that is not expected, just from the name this is the behavior I expect from such a function:
```viml
echo normalizePath('~/vim/vimrc') " Print: /home/user/vim/vimrc
echo normalizePath('~/vim') " Print: /home/user/vim
```

przepompownia: Please wait a while for a vader test that may help you to understand the need to have this function (and eventually propose more accurate name).

autoload/phpactor/fileutils.vim
```diff
@@ -0,0 +1,47 @@
...
+  endfor
+
+  return v:false
+endfunction
+
+function! phpactor#fileutils#isSubdir(expectedSubdir, directory) abort
+  let l:expectedSubdir = phpactor#fileutils#normalizePath(a:expectedSubdir)
+  let l:directory = phpactor#fileutils#normalizePath(a:directory)
+
+  return stridx(l:expectedSubdir, l:directory) >= 0
```
elythyr, line 46: Since your "normalized" the paths they will both start from the root.
```suggestion
  return stridx(l:directory, l:expectedSubdir) == 0
```

autoload/phpactor/fileutils.vim
przepompownia: Please test
```vim
echo fnamemodify('/usr/bin', ':h') " gives /usr
echo fnamemodify('/usr/bin/', ':h') " gives /usr/bin
```

For directories `:p` adds a trailing slash if such did not existed before. We then can use `:h` like in the example above.

autoload/phpactor/fileutils.vim
przepompownia: Equality is sufficient. I overlooked the suggestion of changing the order, but not understand it. The current version works as expected.

autoload/phpactor/project/assigner.vim
przepompownia: Sadly, I can be not able to recognize the boundaries of irony in your reply ;) My English is such as you see and you can expect more misunderstandings ;)

I would like to allow provide the prompt for each new buffer that does not belong to any project. It is far from prompting for each file, its rather a rare case per the whole session. Right, probably we should allow disable this dialog completely and configure phpactor to use one or more non-interactive methods or  (as currently or in a configurable way, e.g. [`initialCwd`, `rootMarkers`]). 

As we discussed earlier, each method has downsides so allowing to interactive choose one of them is a good idea. 

I have no idea how to simulate user input into dialog with vader test but I in such case I we can do it manually and allow to continue automatically, as usual. Such test would be not for run by CI, but maybe it can help to understand what can happen from the user point of view.

przepompownia: > So for me the second option is a no go, we can't ask people to configure their editor so that they can use our tool, it's against the purpose IMO.

Only possibility to interactive choose the method depending (at the moment) on a presence `F` in `shortmess`. With `F` only the interactive choice is unavailable - not the whole phpactor. 

It would be good to minimize dependency of user vim settings. Notice that other tools can also rely on some of them. For example, [set hidden is suggested for using CoC](https://github.com/neoclide/coc.nvim/wiki/F.A.Q#my-vim-is-blocked-sometimes).

autoload/phpactor/fileutils.vim
elythyr: The order might be a mistake, you need to test it

autoload/phpactor/project/project.vim
przepompownia: I outsourced most of directory tree logic into `phpactor#fileutils` and tried to use those functions where it could be possible.

autoload/phpactor/project/assigner.vim
przepompownia: Resolved with this suggestion and handling exception too.

autoload/phpactor/fileutils.vim
przepompownia: See https://github.com/phpactor/phpactor/pull/959/files#diff-ce14ae62aff1c692b7268a0ce6f78a1a

autoload/health/phpactor.vim
```diff
@@ -19,6 +19,26 @@ function! s:check_diagnostics(diagnostics) abort
...
+
+  if v:true == g:phpactorAllowInteractiveProjectResolution
+    call health#report_warn('- interactive choice of method may be unavailable.')
+  endif
+
+  if index(g:phpactorNoninteractiveProjectResolvers, 'manual') >= 0
+    call health#report_warn('- manual input was enabled but it may not work')
+  endif
+endfunction
+
```
elythyr, line 41: I think we should improve that, the goal is that it should be understandable by someone not knowing how everything works.
It's an help to show the user how the plugin will behave and hint on how to configure it if needed.

So first, I would add an information stating which kind of root detection is used.
Then you could add one function per type of detection, each one will have the responsibility to add relevant information.

For instance, we should not bother someone with the `F` being part of `shortmessage` or not if it does not impact his current configuration.

You should also use the `{advice}` parameter for `health#report_warn` and `health#report_error` in order to explain why the `F` option is problematic and only in the case of the manual detection.

autoload/phpactor.vim
```diff
@@ -459,49 +459,13 @@ endfunction
...
     execute ':redraw'
 
     let request = { "action": a:action, "parameters": a:arguments }
 
-    let l:workspaceDir = s:searchDirectoryUpwardForRootPatterns(
-          \ fnamemodify(phpactor#_path(), ':h'),
-          \ g:phpactorProjectRootPatterns,
-          \ g:phpactorInitialCwd
-          \)
+    let l:workspaceDir = (get(b:, 'phpactorProject', v:null) isnot v:null) ? b:phpactorProject.primaryRootPath : g:phpactorInitialCwd
```
elythyr, line 468: This will solve the issue #985
But I think we can do even better, right it only makes possible to use phpactor root detection of the directory from where vim was open.

It should not be phpactor responsibility to resolve the directory in the first place, we should only ensure that we provide the root directory when sending a request.

I mentioned it because on this line I think we should consider the all root detection mechanism as external.
It would reduce the complexity by getting back to real need: provide a root directory.
Initially phpactor was using `g:phpactorInitialCwd`, to prevent any BC break we need to keep this variable with it's value being the directory from which vim was started.
From this point we want to be able to provide a dynamic directory depending on the buffer, meaning we will need to "ask" each buffer what is its root directory.
We can translate this need by
> We want to be able to apply different strategies at runtime

So instead of coupling this piece of code to any system, let's couple it to an abstract strategy that could be defined by the anyone.
```suggestion
    let l:workspaceDir = phpactor#getRootDirectory()
```

With
```vim
if !exists('g:PhpactorRootDirectoryStrategy')
  let g:PhpactorRootDirectoryStrategy = {-> g:phpactorInitialCwd}
endif

function! phpactor#getRootDirectory() abort
  let l:Strategy = get(b:, 'PhpactorRootDirectoryStrategy', g:PhpactorRootDirectoryStrategy)

  if type(function('type')) != type(l:Strategy)
    let l:Strategy = {-> g:phpactorInitialCwd}
  endif

  return l:Strategy()
endfunction
```

At this point the user is free to define both global and buffer local strategies.
We don't have any BC break since the default behavior is the same as before the first attempt to resolve the root directory.
Example of configuration:

#### To use with your system
```vim
" vimrc
let g:PhpactorRootDirectoryStrategy = {-> get(get(b:, 'phpactorProject', {}), 'primaryRootPath', g:phpactorInitialCwd)}
```

#### To use vim pwd (for users with [vim-rooter](https://github.com/airblade/vim-rooter) for instance)
```vim
" vimrc
let g:PhpactorRootDirectoryStrategy = function('getcwd')
```

The goal of this proposal is not to diminish your work at all, it's simply to provide a way of integrating in a extensible way.
So thanks for all what you have done anyone can use phpactor out of box with multiple projects in one vim instances.
Users having already a setup handling this notion can plug into phpactor to reuse what they have.
Users with no need for it will keep things as they were.

ftplugin/php.vim
```diff
@@ -3,5 +3,12 @@ augroup PhpactorInit
     autocmd CompleteDone <buffer> call phpactor#_completeImportClass(v:completed_item)
 augroup END
 
+let g:phpactorProjectAssigner = get(g:, 'phpactorProjectAssigner', phpactor#project#assigner#create(
+            \ phpactor#project#repository#create(),
+            \ g:phpactorProjectRootPatterns,
+            \ g:phpactorGlobalRootPatterns,
+            \ g:phpactorInitialCwd,
+            \ g:phpactorNoninteractiveProjectResolvers
+            \ ))
```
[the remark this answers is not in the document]
przepompownia: Right but I have doubt in which section it should land (configs or none). It does not keep any simple configuration.

autoload/phpactor/fileutils.vim
```diff
@@ -0,0 +1,47 @@
...
+    let l:subdir = phpactor#fileutils#normalizePath(l:subdir)
+    if phpactor#fileutils#isSubdir(l:subdir, l:directory)
+      return v:true
+    endif
+  endfor
+
+  return v:false
+endfunction
+
+function! phpactor#fileutils#isSubdir(expectedSubdir, directory) abort
```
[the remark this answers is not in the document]
przepompownia: Not. "Subdirectory" seems to exist as a single word.

ftplugin/php/autocommands.vim
```diff
@@ -0,0 +1 @@
+call phpactor#project#assigner#assignProjectToBuffer(expand('%'), v:false, g:phpactorAllowInteractiveProjectResolution)
```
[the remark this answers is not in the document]
przepompownia: As you remember, it originally was executed by an autocommand:
https://github.com/phpactor/phpactor/pull/959#pullrequestreview-393187970
;)

No problem for me to revert it and get advantages from displaying `:autocmd FileType`.

autoload/phpactor.vim
przepompownia: I kept such idea (without any implementation) as a backup in case of non-acceptance the whole work, but it should be the default approach - not backup. Your proposition sounds reasonable: gives an abstraction that contains a minimal implementation and can be replaced by any implementation maintained independently in some separate project. 

The whole implementation of this approach is yours so please make PR that closes this after merge. 

@dantleech @elythyr thanks for a lot of critical word.

dantleech: Thanks so much for all your work so far :) Even if this PR isn't merged as-is there is lots of valuable stuff here which we can take forward (especially with the Vader tests).

I've found it quite difficult to follow everything going on here, but I think we could do something like:

- Allowing project-root strategy to be selected (custom or phpactor CWD)
- Strategy for changing CWD if the user opens up a file != the filesystem branch of the current project root (with confirmation).

Either as part of this PR or new PR(s) - would that cover everything?

ftplugin/php/autocommands.vim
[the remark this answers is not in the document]
elythyr: Proof that I didn't thought hard enough about it, sorry :)

tests/VimPlugin/fileutils.vader
```diff
@@ -0,0 +1,46 @@
...
+  Assert ! phpactor#fileutils#isSubdir('/', '/usr')
+
+Execute (Test searching upward given root markers and forbidden dirs):
+  let rootMarkers = ['bar', g:marker]
+  let forbiddenDirs = []
+
+  AssertEqual g:project1Dir, phpactor#fileutils#searchDirectoryUpwardByRootMarkers(g:subdir, rootMarkers, forbiddenDirs)
+
+  let forbiddenDirs = [g:project1Dir.'/foo']
+  AssertEqual v:null, phpactor#fileutils#searchDirectoryUpwardByRootMarkers(g:subdir, rootMarkers, forbiddenDirs)
```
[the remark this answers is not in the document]
przepompownia: The purpose of excluding some directory from be a project root is an attempt to avoid indexing large directory that is supposed to be not a project root. `/` and '$HOME` are usual examples. If searching upward stops at one of excluded directories then there is no reasonable value to suggest and another resolver should be used. In non-interactive variant (`g:phpactorAllowInteractiveProjectResolution == v:false`) the next resolver will be used (`initialCwd` is always present independently of user settings).

It is not intended to disable any marker from be stop in moving upward so I agree with your scenario except the last assertion. 

However, if a fallback directory (initialCwd) is one of excluded (e.g. we start from `~/`) and will be used, then we cannot avoid it.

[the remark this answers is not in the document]
przepompownia: The case added by you is a good example to show that searching upward to any root marker is not always a good resolver. Assuming additionally that we start (and still have CWD) in home directory or one of excluded, we also have a good example to show that `initialCwd` is not god resolver for this case. Thus I added `manual` method and option to enable interactive choice from all methods.

autoload/phpactor.vim
przepompownia: Let this PR die in the current form and (before) implement @elythyr's proposition in a separate one. 

@elythyr would you like to make such PR? I can do it just as well.

tests/VimPlugin/fileutils.vader
```diff
@@ -0,0 +1,46 @@
...
+  AssertEqual g:project1Dir, phpactor#fileutils#searchUpwardBySingleMarker(g:markerFilePath, g:marker)
+  AssertEqual g:project1Dir, phpactor#fileutils#searchUpwardBySingleMarker(g:subdir, g:marker)
+  AssertEqual v:null, phpactor#fileutils#searchUpwardBySingleMarker(g:fixtureDir, g:marker)
+
+  execute "write " . g:project1Dir . '/foo/' .g:marker
+  AssertEqual g:project1Dir . '/foo', phpactor#fileutils#searchUpwardBySingleMarker(g:subdir, g:marker)
+
+Execute (Test phpactor#fileutils#isSubdir()):
+  Assert phpactor#fileutils#isSubdir('/usr/bin', '/usr')
+  Assert phpactor#fileutils#isSubdir('/usr/', '/usr')
```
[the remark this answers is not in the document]
przepompownia: Done

autoload/phpactor/project/bufferMatcher.vim
przepompownia: While I supposed to merge it in phpactor, I had not idea where to put such selector if it does not conform with (partially abstract) `phpactor#input#list` or any other. I deferred it in favor of resolving another issues.

autoload/phpactor.vim
elythyr: You can do it if you want.
The ideal would be to do it quickly enough so it will be in the next release.

Let me know if you won't have the time.

tests/VimPlugin/fileutils.vader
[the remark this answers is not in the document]
elythyr: OK so it's not meant to say "Ignore the vendor as project root and search upwards", it meant to say "Stop right now there is no interest to look anymore".

At first I thought it was an option to avoid taking the vendor directory into account, hence my misunderstanding.

autoload/phpactor/project/bufferMatcher.vim
przepompownia: To be equivalent it (not counting minor mistakes) should return `v:null` when encounter one of `a:filesystemRootMarkers`.

autoload/phpactor.vim
przepompownia: It can be unpredictable so we can assume I have not.

dantleech:
Now that we have the configurable root directory stragies we close this -- thanks for all the work @przepompownia :)

przepompownia:
Thanks also once again. The next step can wait for some free time.

PR #964, opened 2020-04-15
Fix ftplugin

elythyr:
I think we made some mistakes with our migration to a filetype plugin.
The idea is great, the way we did it need to be improved.

The main idea is that configuration are still global so they should not be in the `ftplugin` directory.
I know it's in contradiction with what I said in an issue...
But after thinking about it make sense for **global** configuration.
The rest, which can be related to a buffer only should stay in the `ftplugin` directory.

There is also the problem mentioned by @przepompownia saying that `ftplugin/php.vim` could be sourced only once.
That's because of the condition to check if phpactor is loaded.
But this condition has two main purpose:
* Avoid loading the file twice (could create some errors if we didn't do things correctly)
* Allow a user to disable the plugin from it's vimrc if he wants to

This is not related to filetype plugins and should not be there, also it does not follow the recommendation (see commit message).

As always, I tried to be as clear as possible in my commit messages so if you have any question you can check them our just ask here :)

### DIFF
--- a/autoload/phpactor.vim
+++ b/autoload/phpactor.vim
@@ -89,6 +89,10 @@ endfunction
 
 function! phpactor#_completeImportClass(completedItem)
 
+    if get(b:, 'phpactorOmniAutoClassImport', g:phpactorOmniAutoClassImport) != v:true
+      return
+    endif
+
     if !has_key(a:completedItem, "word")
         return
     endif
--- a/autoload/phpactor.vim
+++ b/autoload/phpactor.vim
@@ -11,6 +11,8 @@
 "     https://github.com/google/vimdoc. See
 "     https://phpactor.github.io/phpactor/developing.html#vim-help
 
+let s:_phpactorCompletionMeta = {}
+
 function! phpactor#Update()
     let current = getcwd()
     execute 'cd ' . g:phpactorpath
@@ -53,7 +55,7 @@ function! phpactor#Complete(findstart, base)
     let issues = result['issues']
 
     let completions = []
-    let g:_phpactorCompletionMeta = {}
+    let s:_phpactorCompletionMeta = {}
 
     if !empty(suggestions)
         for suggestion in suggestions
@@ -66,7 +68,7 @@ function! phpactor#Complete(findstart, base)
                         \ 'icase': g:phpactorCompletionIgnoreCase
                         \ }
             call add(completions, completion)
-            let g:_phpactorCompletionMeta[phpactor#_completionItemHash(completion)] = suggestion
+            let s:_phpactorCompletionMeta[phpactor#_completionItemHash(completion)] = suggestion
         endfor
     endif
 
@@ -92,11 +94,11 @@ function! phpactor#_completeImportClass(completedItem)
     endif
 
     let hash = phpactor#_completionItemHash(a:completedItem)
-    if !has_key(g:_phpactorCompletionMeta, hash)
+    if !has_key(s:_phpactorCompletionMeta, hash)
         return
     endif
 
-    let suggestion = g:_phpactorCompletionMeta[hash]
+    let suggestion = s:_phpactorCompletionMeta[hash]
 
     if !empty(get(suggestion, "class_import", ""))
         call phpactor#rpc("import_class", {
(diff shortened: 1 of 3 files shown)


autoload/phpactor.vim
```diff
@@ -11,6 +11,8 @@
 "     https://github.com/google/vimdoc. See
 "     https://phpactor.github.io/phpactor/developing.html#vim-help
 
+let s:_phpactorCompletionMeta = {}
```
dantleech, line 14: :+1:

plugin/phpactor.vim
```diff
@@ -1,8 +1,8 @@
-if exists('g:phpactorLoaded')
+if exists('g:loaded_phpactor')
```
dantleech, line 1: Would it make sense to name it `phpactorLoaded`? All the other global variables are prefixed with `phpactor` and are camelcase.

ftplugin/php.vim
```diff
@@ -1,57 +1,16 @@
...
-let g:phpactorGlobalRootPatterns = get(g:, 'phpactorGlobalRootPatterns', ['/', '/home'])
-
-if g:phpactorOmniAutoClassImport == v:true
-    autocmd CompleteDone *.php call phpactor#_completeImportClass(v:completed_item)
-endif
+function! s:import_on_complete_done(completed_item) abort
+    if get(b:, 'phpactorOmniAutoClassImport', g:phpactorOmniAutoClassImport) != v:true
+      return
+    endif
+    echom string(get(b:, 'phpactorOmniAutoClassImport', g:phpactorOmniAutoClassImport))
```
dantleech, line 5: Is this a debug statement?

plugin/phpactor.vim
elythyr: > it does not follow the recommendation (see commit message)

For me it's an exception we need to keep, in your vim try to expand `:echo g:loaded_` for completion you should see a bunch of them depending on how many plugins you have.
It's given like that in the vim documentation none the less.

ftplugin/php.vim
elythyr: it's possible, I don't remember using it but I will remove because it has nothing to do here ;)

elythyr: Removed.

By the way, do you think we should add the condition in `phpactor#_completeImportClass` instead of creating a temporary function as I did ?
The goal was to not change the way `phpactor#_completeImportClass` works in case someone wanted to use it but I'm not sure it really make sense.

dantleech:
It's fine to change it I think.

On 16 April 2020 09:35:17 CEST, Camille Dejoye <notifications@github.com> wrote:
>@elythyr commented on this pull request.
>
>
>
>> -" if contained within
>-let g:phpactorProjectRootPatterns = get(g:,
>'phpactorProjectRootPatterns', ['composer.json', '.git',
>'.phpactor.json', '.phpactor.yml'])
>-
>-""
>-" The list of directories that should not be considered as workspace
>root directory
>-" (in addition to '/' which is always considered)
>-let g:phpactorGlobalRootPatterns = get(g:,
>'phpactorGlobalRootPatterns', ['/', '/home'])
>-
>-if g:phpactorOmniAutoClassImport == v:true
>-    autocmd CompleteDone *.php call
>phpactor#_completeImportClass(v:completed_item)
>-endif
>+function! s:import_on_complete_done(completed_item) abort
>+    if get(b:, 'phpactorOmniAutoClassImport',
>g:phpactorOmniAutoClassImport) != v:true
>+      return
>+    endif
>+    echom string(get(b:, 'phpactorOmniAutoClassImport',
>g:phpactorOmniAutoClassImport))
>
>Removed.
>
>By the way, do you think we should add the condition in
>`phpactor#_completeImportClass` instead of creating a temporary
>function as I did ?
>The goal was to not change the way `phpactor#_completeImportClass`
>works in case someone wanted to use it but I'm not sure it really make
>sense.
>
>-- 
>You are receiving this because you commented.
>https://github.com/phpactor/phpactor/pull/964#discussion_r409342658

-- 

elythyr:
I tried to run vader tests since they are not in the pipeline.
There is 3 which are broken on develop, wee might want to fix them first.

dantleech:
Updated the changelog

elythyr:
Nice I will rebase #965

PR #965, opened 2020-04-15
Fix commands

elythyr:
This PR is about fixing the commands declared in `ftplugin/php/commands.vim`.

All the commands are global when most of them should be local to a buffer.
I moved the commands that should be global to `plugin/phpactor.vim` so we need to merge https://github.com/phpactor/phpactor/pull/964 and I need to rebase this branch before we can merge.

### DIFF
--- a/ftplugin/php/commands.vim
+++ b/ftplugin/php/commands.vim
@@ -1,130 +1,101 @@
 ""
-" Update Phpactor to the latest version using the branch
-" defined with @setting(g:phpactorBranch)
-command! -nargs=0 PhpactorUpdate call phpactor#Update()
-
-""
-" Clear the entire cache - this will take effect for all projects.
-command! -nargs=0 PhpactorCacheClear call phpactor#CacheClear()
-
-""
-" Show some information about Phpactor's status
-command! -nargs=0 PhpactorStatus call phpactor#Status()
-
-""
-" Dump Phpactor's configuration
-command! -nargs=0 PhpactorConfig call phpactor#Config()
-
-""
 " Extract a new method from the current selection
-command! -range=% PhpactorExtractMethod call phpactor#ExtractMethod()
+command! -buffer -range=% PhpactorExtractMethod call phpactor#ExtractMethod()
 
 ""
 " Extract the selected expression and assign it to a variable before (placing
 " it before the current statement)
-command! -range=% PhpactorExtractExpression call phpactor#ExtractExpression('v')
+command! -buffer -range=% PhpactorExtractExpression call phpactor#ExtractExpression('v')
 
 ""
 " Extract a constant from a literal
-command! -nargs=0 PhpactorExtractConstant call phpactor#ExtractConstant()
+command! -buffer -nargs=0 PhpactorExtractConstant call phpactor#ExtractConstant()
 
 ""
 " Expand the class name under the cursor to it's fully-qualified-name
-command! -nargs=0 PhpactorClassExpand call phpactor#ClassExpand()
+command! -buffer -nargs=0 PhpactorClassExpand call phpactor#ClassExpand()
 
 ""
 " Import the name under the cusor. If multiple options are available, you
 " are able to choose one.
-command! -nargs=0 PhpactorImportClass call phpactor#ImportClass()
+command! -buffer -nargs=0 PhpactorImportClass call phpactor#ImportClass()
 
 ""
 " Attempt to import all non-resolvable classes in the current class (based
 " on offset position)
-command! -nargs=0 PhpactorImportMissingClasses call phpactor#ImportMissingClasses()
+command! -buffer -nargs=0 PhpactorImportMissingClasses call phpactor#ImportMissingClasses()
 
 ""
 " Goto the definition of the class, method or function under the cursor. Open
 " the definition in the current window.
-command! -nargs=0 PhpactorGotoDefinition call phpactor#GotoDefinition()
+command! -buffer -nargs=0 PhpactorGotoDefinition call phpactor#GotoDefinition()
 
 ""
 " Goto type (class) of the symbol under the cursor.
-command! -nargs=0 PhpactorGotoType call phpactor#GotoType()
+command! -buffer -nargs=0 PhpactorGotoType call phpactor#GotoType()
 
 ""
 " As with @command(PhpactorGotoDefinition) but open in a vertical split.
-command! -nargs=0 PhpactorGotoDefinitionVsplit call phpactor#GotoDefinitionVsplit()
+command! -buffer -nargs=0 PhpactorGotoDefinitionVsplit call phpactor#GotoDefinitionVsplit()
 
 ""
 " As with @command(PhpactorGotoDefinition) but open in a horizontal split.
-command! -nargs=0 PhpactorGotoDefinitionHsplit call phpactor#GotoDefinitionHsplit()
+command! -buffer -nargs=0 PhpactorGotoDefinitionHsplit call phpactor#GotoDefinitionHsplit()
 
 ""
 " As with @command(PhpactorGotoDefinition) but open in a new tab
-command! -nargs=0 PhpactorGotoDefinitionTab call phpactor#GotoDefinitionTab()
+command! -buffer -nargs=0 PhpactorGotoDefinitionTab call phpactor#GotoDefinitionTab()
 
 ""
 " Load all implementations of the class under the cursor into the quick-fix
 " list.
-command! -nargs=0 PhpactorGotoImplementations call phpactor#GotoImplementations()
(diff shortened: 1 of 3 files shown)


ftplugin/php/commands.vim
```diff
@@ -1,130 +1,101 @@
...
-
-""
-" Remove an extension
-command! -nargs=1 PhpactorExtensionRemove call phpactor#ExtensionRemove(<q-args>)
+command! -buffer -nargs=0 PhpactorMoveFile call phpactor#MoveFile()
 
 ""
 " Create a new class. You will be offered a choice of templates.
-command! -nargs=0 PhpactorClassNew call phpactor#ClassNew()
+command! -buffer -nargs=0 PhpactorClassNew call phpactor#ClassNew()
```
przepompownia, line 74: Assuming we have implemented the idea of project root per buffer and that we can manually assign some project to non-php buffer (e.g. composer.json in a newly created project), we could move this command to `plugin`.

dantleech: Incidentally - there is also a case for using f.e. `PhpactorGotoDefinition` (and similar) in `.yml`, `.xml`, Gherkin files, etc.

dantleech: We probably need to decide which commands are global, and which only apply to buffers (or make them all available unconditionally)

elythyr: I went pretty quickly other the commands so I might have missed some for sure.
For the go to definition I didn't realized we could use them in non PHP files...
For the new class it's something that might be appreciated anywhere indeed no matter if the file is part of a project or not.

To be honest I never use the commands, go to definition is mapped on `<C-]>` and `<C-w><C-]>` and I don't use the new class functionality because I have snippets that can do that for me a lot better.
Just to say that I might not have a realistic vision of when to use them.

przepompownia: > For the new class it's something that might be appreciated anywhere indeed no matter if the file is part of a project or not.

Right, although providing the project root to this command (or use the assigned project root to the current buffer previously) would help resolve namespace (or file path if there would be possible to adding it given a namespace) and generate suggestions.

elythyr: I think this should be handle in `phpactor#newClass()` directly.
We can easly retrieve the current buffer and find out if it belongs to a project or not.

przepompownia: It would work if the current buffer belongs to any project. If the buffer is for example empty, special or help buffer providing the root can be still needed.

ftplugin/php.vim
```diff
@@ -1,58 +1,16 @@
...
+    endif
+    echom string(get(b:, 'phpactorOmniAutoClassImport', g:phpactorOmniAutoClassImport))
+
+    call phpactor#_completeImportClass(a:completed_item)
+endfunction
+
+augroup PhpactorInit
+    autocmd! * <buffer>
+    autocmd CompleteDone <buffer> call <SID>import_on_complete_done(v:completed_item)
+augroup END
```
przepompownia, line 13: Both definitions will be still redefined each time similarly like those from before change.

przepompownia: I did not test this behaviour yet. How it behaves with completion triggered by another client?

elythyr: > Both definitions will be still redefined each time similarly like those from before change.

I don't understand, which definition ? The auto commands ?
`autocmd! * <buffer>` will delete all auto commands from the group `PhpactorInit` linked to the current buffer.
That's the usual way of defining auto commands, you put them into a group so each time you want to define them, you first empty the all group and then declare them.
If they don't exists it does not throw an error, if they does then they are removed before being redefined.

przepompownia: Maybe I wrote it not clearly enough. I did not mean what's going on inside this augroup. I meant redefinition of the whole `augroup PhpactorInit` and `import_on_complete_done()`. 

The function could be moved somewhere into autoload. 

I know and use the common way of defining augroups, however thanks for explanation.

It is usually not need to use `autocmd FileType` while its action can be simply used in ftplugin. Why not to move whole augroup to `plugin`and use 
```vim
autocmd FileType php autocmd CompleteDone <buffer=abuf> call <SID>import_on_complete_done(v:completed_item)
```
in this case? I still not tested the above proposition ;)

elythyr: Ok so for the first point: `import_on_complete_done` was removed in the other PR.
I was waiting for @dantleech opinion about integrating the logic into `phpactor#_importOnCompletedone` or whatever the name is :)

I don't mind the `augroup` being "redefine", it's suppose to work like that, there will still be only one group with this name.

I don't see how it will be beneficial to move the auto command in the `plugin` directory just to plug it on another auto command.
I guess it's because the file won't be source twice ?
But it does not matter, `ftplugin/php.vim` will also be loaded only once per buffer.
And it could be reloaded without any issues, it's not wrong to reload a file.
For instance I reload my vimrc constantly to update my plugins because they are not loaded by default. As long as the file is properly written it's not harmful.

ftplugin/php/commands.vim
elythyr: So what about this list ?
I would like to be able to merge this so we can move on.

For `PhpactClassNew` I'm not against it, but I don't really have an opinion since I never use It.
For the navigation commands (goto and all that) I think it's better to declare them in their own `filetype/cucumber.vim` file.
Especially because this support came from an extension that is not bundled with phpactor by default.
We could even say that the extension should do it itself, but we don't have such a feature right now but we should separate things clearly.

ftplugin/php.vim
elythyr: > I did not test this behaviour yet. How it behaves with completion triggered by another client?

Exactly as it has always done, there is an internal variable that is only filled by `phpactor#Complete`, so no matter how the method is called it will only work after calling `phpactor#Complete`.
This variable is reset after the function is done (we might want to add some try/catch in the mixt but not related to this PR) and on each call to `phpactor#Complete`.

Bottom line, the changes made here have no impact on the current behavior.

> Both definitions will be still redefined each time similarly like those from before change.

As explained it's not an issue for me, so if it's OK for you I'll resolve this comment soon.

ftplugin/php/commands.vim
przepompownia: Independently of the place of the command definition I can assign a mapping or any custom command to `phpactor#ClassNew()` directly.

ftplugin/php/commands.vim
```diff
@@ -1,130 +1,101 @@
...
 " Attempt to import all non-resolvable classes in the current class (based
 " on offset position)
-command! -nargs=0 PhpactorImportMissingClasses call phpactor#ImportMissingClasses()
+command! -buffer -nargs=0 PhpactorImportMissingClasses call phpactor#ImportMissingClasses()
 
 ""
 " Goto the definition of the class, method or function under the cursor. Open
 " the definition in the current window.
-command! -nargs=0 PhpactorGotoDefinition call phpactor#GotoDefinition()
+command! -buffer -nargs=0 PhpactorGotoDefinition call phpactor#GotoDefinition()
```
dantleech, line 31: GotoDefinition can work on Gherkin files too with an extension -- possibly any filetype could work with Phpactor here.

ftplugin/php/commands.vim
```diff
@@ -1,130 +1,101 @@
...
 
 ""
 " Extract a constant from a literal
-command! -nargs=0 PhpactorExtractConstant call phpactor#ExtractConstant()
+command! -buffer -nargs=0 PhpactorExtractConstant call phpactor#ExtractConstant()
 
 ""
 " Expand the class name under the cursor to it's fully-qualified-name
-command! -nargs=0 PhpactorClassExpand call phpactor#ClassExpand()
+command! -buffer -nargs=0 PhpactorClassExpand call phpactor#ClassExpand()
```
dantleech, line 16: ClassExpand is useful on XML and YAML files (or any text document)

ftplugin/php/commands.vim
elythyr: Moved

ftplugin/php/commands.vim
elythyr: Moved

plugin/phpactor.vim
```diff
@@ -3,6 +3,8 @@ if exists('g:loaded_phpactor')
 endif
 
 let g:loaded_phpactor = 1
+
+" Config {{{
 let g:phpactorpath = expand('<sfile>:p:h') . '/..'
 let g:phpactorbinpath = g:phpactorpath. '/bin/phpactor'
```
dantleech, line 9: Unrelated - should probably deprecate this and create a documented `g:phpactorBinPath`

dantleech:
Is this ready to go now?

PR #987, opened 2020-04-29
Fix goto commands

camilledejoye:
As discussed the accumulation of commands to go to a definition was a bad idea.

##### Todo
- [x] Merge #965 first to avoid conflicts and rebase afterwards
- [x] Ensure the deprecated commands where in a previous release, delete them if not
- [x] Delete the deprecated autoload functions ? Since they are private I guess the deprecations does not affect the and I can delete immedialty, @dantleech ?
- [x] Improve the "window targets" to use valid vim commands like `edit`, `split` (ensure that the "window targets" are never sended by phpactor but only define in the plugin)
- [x] Use mods (`:h <mods>`) to control with more granularity how the window will be opened ? `toplef', `aboveleft`, etc.

### DIFF
--- a/autoload/phpactor.vim
+++ b/autoload/phpactor.vim
@@ -220,30 +220,6 @@ function! phpactor#GotoDefinition(...)
         \ "mods": mods,
     \ })
 endfunction
-""
-" deprecated, use @function(phpactor#gotodefinition) instead
-"
-" as with @function(phpactor#gotodefinition) but open in a vertical split.
-function! phpactor#GotoDefinitionVsplit()
-    echoerr "phpactor#GotoDefinitionVsplit is deprecated use phpactor#GotoDefinition('vsplit') instead"
-    call phpactor#GotoDefinition('vsplit')
-endfunction
-""
-" deprecated, use @function(phpactor#gotodefinition) instead
-"
-" as with @function(phpactor#gotodefinition) but open in an horizontal split.
-function! phpactor#GotoDefinitionHsplit()
-    echoerr "phpactor#GotoDefinitionHsplit is deprecated use phpactor#GotoDefinition('split') instead"
-    call phpactor#GotoDefinition('split')
-endfunction
-""
-" deprecated, use @function(phpactor#gotodefinition) instead
-"
-" as with @function(phpactor#gotodefinition) but open in a new tab.
-function! phpactor#GotoDefinitionTab()
-    echoerr "phpactor#GotoDefinitionTab is deprecated use phpactor#GotoDefinition('tabnew') instead"
-    call phpactor#GotoDefinition('tabnew')
-endfunction
 
 ""
 " @usage [target] [mods]
--- a/autoload/phpactor.vim
+++ b/autoload/phpactor.vim
@@ -186,17 +186,45 @@ function! phpactor#ImportMissingClasses()
     call phpactor#rpc("import_missing_classes", {"source": phpactor#_source(), "path": expand('%:p')})
 endfunction
 
-function! phpactor#_GotoDefinitionTarget(target)
+""
+" @default target=`focused_window`
+"
+" Goto the definition of the symbol under the cursor.
+" Open the definition in the [target] window, see @section(window-target) for
+" the list of possible targets.
+function! phpactor#GotoDefinition(...)
     call phpactor#rpc("goto_definition", {
                 \"offset": phpactor#_offset(),
                 \"source": phpactor#_source(),
                 \"path": expand('%:p'),
-                \"target": a:target,
+                \"target": a:0 ? a:1 : 'focused_window',
                 \'language': &ft})
 endfunction
-function! phpactor#GotoDefinition()
-    call phpactor#_GotoDefinitionTarget('focused_window')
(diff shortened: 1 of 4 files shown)


elythyr:
>  Improve the "window targets" to use valid vim commands like edit, split (ensure that the "window targets" are never sended by phpactor but only define in the plugin)
Use mods (:h <mods>) to control with more granularity how the window will be opened ? toplef', aboveleft`, etc.

Those two things are done but to be clean it requires a PR for `rpc-extension` and `reference-finder-rpc-extension` packages.
Until know phpactor was providing us with a `target` parameter.
But I think this should not be in here, I wonder if it wasn't done just as a way to provide the information to the `phpactor#_rpc_dispatch` function.
So these 2 PRs will remove every thing related to that in the side of phpactor so it will no longer be coupled to vim plugin.

That being said I just realize it will be a major BC break for the emac plugin...
@dantleech do you prefer to keep the packages as they are for now ?
It doesn't prevent the current PR to work, whatever target phpactor will provide in the response wlil be overridden.

doc/phpactor.txt
```diff
@@ -211,11 +255,11 @@ are available for you to copy:
...
         \ :PhpactorGotoDefinition<CR>
     au FileType php,cucumber nmap <buffer> <Leader>Oh
-        \ :PhpactorGotoDefinitionHsplit<CR>
+        \ :PhpactorGotoDefinition split<CR>
     au FileType php,cucumber nmap <buffer> <Leader>Ov
-        \ :PhpactorGotoDefinitionVsplit<CR>
+        \ :PhpactorGotoDefinition vsplit<CR>
     au FileType php,cucumber nmap <buffer> <Leader>Ot
-        \ :PhpactorGotoDefinitionTab<CR>
+        \ :PhpactorGotoDefinition tabnew<CR>
```
dantleech, line 262: Could possibly remove these mappings?

elythyr: I guess, it was there so I just updated it but we can remove it since there is another example already.

autoload/phpactor.vim
```diff
@@ -732,28 +797,7 @@ function! s:openFileInSelectedTarget(filePath, target, useOpenWindow, forceReloa
...
-    if a:target ==# 'hsplit'
-        exec ':split ' . a:filePath
-        return
-    endif
-
-    if a:target ==# 'new_tab'
-        exec ':tabnew ' . a:filePath
-        return
-    endif
+    execute printf('%s %s %s', a:mods, a:target, a:filePath)
```
dantleech, line 800: Can we validate the target here?

If I provide `tasd` as a target i get:

```
E492: Not an editor command:  tasd /home/daniel/www/phpactor/phpactor/vendor/psr/event-dispatcher/src/EventDispatcherInterface.php 
```

Can we instead throw an error listing possible targets:

```
invalid target "tasd` valid targets: "tabe", "vspl", // ... etc
```

dantleech: This is also important because I can pass "editor commands" e.g. `:PhpactorGotoDefinition bdelete`

dantleech: and if I execute:

```
:PhpactorGotoDefinition echo "asd"
```

I get:

```
E121: Undefined variable: home 
```

dantleech:
Sorry for taking so long to look at this, it looks really good. Would be good if we could:

- Remove the deprecated functions
- Validate the target (as commented)

elythyr:
No issue as you have seen I took even longer :)
I will have to get my head back in the code before answering your questions so I will save that thread and came back to it later.

autoload/phpactor.vim
elythyr: I see your point but I don't know a way to validate against every possibilities offered by the editor.
So if we miss some a user will be limited (at least until we fix it), and we will have to maintain it all the way.
So I would argue that it's up to the user to check vim capabilities first, this is barely a wrapper allowing as mush flexibility as possible for the user.

Now if you feel that we should actually protect the user against himself, I guess we can reuse `s:windowTargets` to validate the targets.
But for the modifiers I don't thinks it's necessary, vim should fail by itself if provided with an invalid modifier to the command... I hope :)

In a way I think I didn't want to bother with that because vim does not, try `toto /tmp/file` and you got the same error.
I guess vim follow the unix philosophy saying "you now what you do" and I'm fine doing the same thing.

Let me know if you want phpactor to be more user friendly than "unix oriented" :D

dantleech: Well - this is an unhandled error (undefined variable) - we should handle the error. If we can suggest what values the user can use, then all the better.

doc/phpactor.txt
elythyr: Done

autoload/phpactor.vim
elythyr: Done, check if it's OK with you.

camilledejoye:
Done a little rebase to fresh it up ;)

dantleech:
I get this error when using `:PhpactorNavigate`

image

dantleech:
traking in #1136

PR #1027, opened 2020-05-04
Add configurable root directory strategies

przepompownia:


przepompownia:
Closes #959 
https://github.com/phpactor/phpactor/pull/959#pullrequestreview-402619224

przepompownia:
https://github.com/przepompownia/vim-project still needs some work but can be used to test.

przepompownia:
Failed test remembers about upgrading minimal vim requirements:
https://github.com/vim/vim/commit/069c1e7fa9f45a665064f7f2c17da84d6a48f544

dantleech:
Happy to upgrade to 8

przepompownia:
The last commit uses unofficial `ppa:jonathonf/vim` repository. I am not sure if it is a good idea. We (assuming using Travis) could build vim from source but it probably would occur each time.

dantleech:
> I am not sure if it is a good idea.

What's the worst that could happen?

---

This looks good - what would be the next steps if any?

przepompownia:
I think that in worst case we will have to build vim from source (it takes time and resources) or change Travis to something more flexible (Travis seems to have limited choice of docker images). 

I would like to add a documentation howto integrate it with external tools (in addition to the existing with global variables declarations). What do you think about doing it in a separate PR and merge this as is (eventually with other fixes)? Currently it works without any BC break. 

I need some time to complete documentation in vim-project and (before) decide how to expose abstraction instead of referencing to b:project (that probably should be internal). 

If i had to document an integration with vim-rooter, I also need some time to test. Both things could be done with separate PR-s.

dantleech:
Sounds good :+1:

PR #1040, opened 2020-05-07
Update tag stack on GoToDefinition

Pashugan:
Easy going back with Ctrl+T. Have been using it for a couple of weeks now, seems to work well.


dantleech:
thanks, I'm not familiar with this currently -- @elythyr @Pashugan any opinions?

przepompownia:
I did not ever used tagstack but why not to provide this information? 

As a beginner in this I see that tagstack is created and I can go back with `Ctrl-T` or `:pop`.

```
:tags                                                                                                                                                                                                                                           
  # TO tag         FROM line  in file/text                                                                                                                                                                                                      
> 1  1 configureExtensionManager    69                                                                                                                                                                                                          
  2  1 PARAM_EXTENSION_VENDOR_DIR   211  $config[ExtensionManagerExtension::PARAM_EXTENSION_VENDOR_DIR] = $extensionDir = __DIR__ . '/../extensions';                                                                                           
  3  1 PARAM_EXTENSION_CONFIG_FILE    67  ~/.vim/pack/bundle/opt/phpactor/vendor/phpactor/extension-manager-extension/lib/ExtensionManagerExtension.php                                                                                         
  4  1 ListCommand       105  ~/.vim/pack/bundle/opt/phpactor/vendor/phpactor/extension-manager-extension/lib/ExtensionManagerExtension.php                                                                                                     
  5  1 version            60  ~/.vim/pack/bundle/opt/phpactor/vendor/phpactor/extension-manager-extension/lib/Command/ListCommand.php 
```
Trying to move forward (`:tag`) ends with 
```
E433: No tags file                                                                                                                                                                                                                              
E426: tag not found: configureExtensionManager 
```
Why the tags file is needed to move through tagstack? I have intentionally `set tags=` (empty). 

It seems to work, but only to move back.

Pashugan:
Sorry, my bad, I didn't explain it properly.

The tag file is normally needed when you use `Ctrl+]` or `:tag`, in that case the tag stack is handled automatically by vim.

I'm using the phpactor-mappings plugin that [re-maps](https://github.com/elythyr/phpactor-mappings/blob/master/ftplugin/php.vim#L71-L73) `Ctrl+]` to phpactor's `GotoDefinition`s. My patch just restores the conventional behaviour when you can go back through the tag stack by `Ctrl+T`.

This explanation made me think this should be added to phpactor-mappings as pushing tags on the tag stack might be an unwanted side effect for someone. Although for me it works perfectly well, feel free to abandon. Thanks.

przepompownia:
It seems that from those two plugins, phpactor is more accurate one to change tagstack than the other because jumping between places is made by phpactor. 

I did not work out if tag file requirement to move forward is a vim limitation or if it is possible to fix your PR to allow move forward. If moving back does not require existing tags file, then why moving forward does it?

Switching between places across windows may require some attention.

dantleech:
I navigate "back" using jump lists (`:help jump`) with `CTRL-O` and `CTRL-T`

przepompownia:
I also navigate this way. Tagstack seems to provide specific subset of jump places. It seems to allow skip steps not related to jumping between definitions (references etc.).

elythyr:
Hi, a bit late for the party but whatever:
> This explanation made me think this should be added to phpactor-mappings as pushing tags on the tag stack might be an unwanted side effect for someone.

For me it should definitely not be in phpactor, the use of `<C-]>` is simply to avoid learning a new mapping, which might be as good as it was previously now that we also have the ability to goto references, implementations, etc.

I think I understand your motivation but I have a hard time to relate to it because I never used `<C-Tab>`, only `<C-O>` and `<C-I>` to navigate the jumplist and `<C-^>` to go to the alternate file (which feels a bit like `<C-t>` to me).

The goal of the mappings plugin was to add some kind of sensible plugins in addition to an easy way to define them, I don't think it should do anymore than that.
I hope it make sense :)

dantleech:
ok, let's close this then. thanks @Pashugan for the idea :)

PR #1069, opened 2020-06-04
Auto-Function Import (language server)

dantleech:
Support for importing functions

recording1


elythyr:
Nice, that make me remember that functions are not completed where using a `\`.
For example typing `\array_me` does not provide any completion.

dantleech:
Same for classes?

elythyr:
> Same for classes?

 I think so but I'm not entirely sure and I don't have access to it on this computer...

PR #1081, opened 2020-06-18
Replace ocramius/package-versions with composer/package-versions

### LINKED ISSUE #1080: Conflicts Composer2
Currently [Composer2 is in preparation](https://php.watch/articles/composer-2), but we can't install Phpactor with Composer2.

```
% php --version
PHP 7.4.3 (cli) (built: May 26 2020 12:24:22) ( NTS )
Copyright (c) The PHP Group
Zend Engine v3.4.0, Copyright (c) Zend Technologies
    with Zend OPcache v7.4.3, Copyright (c), by Zend Technologies

% composer --version
Composer version 2.0-dev (2.0-dev+52afa5ef1603f56f3ba309920cd099d13563031d) 2020-06-17 15:54:06

% composer install
Installing dependencies from lock file (including require-dev)
Verifying lock file contents can be installed on current platform.
Your lock file does not contain a compatible set of packages. Please run composer update.

  Problem 1
    - ocramius/package-versions is locked to version 1.5.1 and an update of this package was not requested.
    - ocramius/package-versions 1.5.1 requires composer-plugin-api ^1.0.0 -> found composer-plugin-api[2.0.0] but it does not match the constraint.
  Problem 2
    - dantleech/what-changed is locked to version 0.1.0 and an update of this package was not requested.
    - dantleech/what-changed 0.1.0 requires composer-plugin-api ^1.1 -> found composer-plugin-api[2.0.0] but it does not match the constraint.

Use the option --with-all-dependencies to allow upgrades, downgrades and removals for packages currently locked to specific versions.

ocramius/package-versions only provides support for Composer 2 in 1.8+, which requires PHP 7.4.
If you can not upgrade PHP you can require composer/package-versions-deprecated to resolve this with PHP 7.0+.

You are using a snapshot build of Composer 2, which some of your plugins seem to be incompatible with. Make sure you update your plugins or report an issue to them to ask them to support Composer 2. To work around this you can run Composer with --ignore-platform-req=composer-plugin-api, but this may result in broken plugins and bigger problems down the line.
```

[ocramius/package-versions](https://github.com/Ocramius/PackageVersions) support Composer 2.x in https://github.com/Ocramius/PackageVersions/pull/133, but that requires `php:^7.4.0`.

pierredup:
Composer provides a fork of ocramius/package-versions which is compatible with both Composer 1 and 2 and PHP 7.0+. So switching to the forked version allows installation when using Composer version 2 and PHP < 7.4.

Fixes #1080

Note that when using Composer 2, you would still not be able to install all the dependencies as [dantleech/what-changed](https://github.com/dantleech/what-changed) depends on composer-plugin-api:^1.1. So that package would need to be updated first to be compatible with composer-plugin-api 2

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
         "composer/xdebug-handler": "^1.1",
         "dantleech/invoke": "^1.0.1",
         "monolog/monolog": "^1.0",
-        "ocramius/package-versions": "^1.0",
+        "composer/package-versions-deprecated": "^1.0",
         "php": "^7.3",
         "phpactor/class-mover": "^0.1.2",
         "phpactor/class-to-file": "^0.3.3",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "f4486cc3326ff144f9721a901f925dd5",
+    "content-hash": "2d96064aee631c91ea052a5c69c233c6",
     "packages": [
         {
             "name": "amphp/amp",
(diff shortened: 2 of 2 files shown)


dantleech:
Great, thanks. I'll added a note to update what-changed, so hopefully the next release will support composer 2: https://github.com/orgs/phpactor/projects/1

zonuexe:
@pierredup Thank you!

Issue #1105, opened 2020-09-30
Sort the result of "Find References" and related features


elythyr:
Hi,

That's something that bugs me for a while now so I wanted to leave a trace.
When looking for references, implementations, etc. in the code the result list is not sorted.
Which makes it quite difficult to quickly identify the different files impacted, I don't have a test project to gave an example but that's the kind of result we will get:
```
directory/something/File1.php:23
directory/somethingElse/File2.php:89
directory/something/File1.php:14
```

When it will make more sense to have something like:
```
directory/something/File1.php:14
directory/something/File1.php:23
directory/somethingElse/File2.php:89
```

PR #1109, opened 2020-10-11
Fix order references

elythyr:
Order the result of the find references feature.
Fix https://github.com/phpactor/phpactor/issues/1105

Update both the RPC handler and the `phpactor#quickfix#fzf` strategy, see the related commit for an explanation but the problem was that the current implementation did not preserve the order of the provided entries.

### DIFF
--- a/lib/Extension/ClassMover/Rpc/ReferencesHandler.php
+++ b/lib/Extension/ClassMover/Rpc/ReferencesHandler.php
@@ -266,27 +266,37 @@ class ReferencesHandler extends AbstractHandler
         string $source = null,
         string $replacement = null
     ) {
-        switch ($symbolContext->symbol()->symbolType()) {
-            case Symbol::CLASS_:
-                [$source, $references] = $this->classReferences($filesystem, $symbolContext, $source, $replacement);
-                break;
-            case Symbol::METHOD:
-                [$source, $references] = $this->memberReferences($filesystem, $symbolContext, ClassMemberQuery::TYPE_METHOD, $source, $replacement);
-                break;
-            case Symbol::PROPERTY:
-                [$source, $references] = $this->memberReferences($filesystem, $symbolContext, ClassMemberQuery::TYPE_PROPERTY, $source, $replacement);
-                break;
-            case Symbol::CONSTANT:
-                [$source, $references] = $this->memberReferences($filesystem, $symbolContext, ClassMemberQuery::TYPE_CONSTANT, $source, $replacement);
-                break;
-            default:
-                throw new \RuntimeException(sprintf(
-                    'Cannot find references for symbol type "%s"',
-                    $symbolContext->symbol()->symbolType()
-                ));
-        }
+        [$source, $references] = $this->doPerformFindOrReplaceReferences(
+            $symbolContext,
+            $filesystem,
+            $source,
+            $replacement,
+        );
 
         return [$source, $this->sortReferences($references)];
+    }
+
+    private function doPerformFindOrReplaceReferences(
+        SymbolContext $symbolContext,
+        string $filesystem,
+        string $source = null,
+        string $replacement = null
+    ) {
+        switch ($symbolContext->symbol()->symbolType()) {
+            case Symbol::CLASS_:
+                return $this->classReferences($filesystem, $symbolContext, $source, $replacement);
+            case Symbol::METHOD:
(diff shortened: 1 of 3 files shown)


elythyr:
Any news on the CI ?
I guess you're waiting to fix it before merging anything else ?

lib/Extension/ClassMover/Rpc/ReferencesHandler.php
```diff
@@ -268,19 +268,46 @@ private function performFindOrReplaceReferences(
...
             case Symbol::METHOD:
-                return $this->memberReferences($filesystem, $symbolContext, ClassMemberQuery::TYPE_METHOD, $source, $replacement);
+                [$source, $references] = $this->memberReferences($filesystem, $symbolContext, ClassMemberQuery::TYPE_METHOD, $source, $replacement);
+                break;
             case Symbol::PROPERTY:
-                return $this->memberReferences($filesystem, $symbolContext, ClassMemberQuery::TYPE_PROPERTY, $source, $replacement);
+                [$source, $references] = $this->memberReferences($filesystem, $symbolContext, ClassMemberQuery::TYPE_PROPERTY, $source, $replacement);
+                break;
             case Symbol::CONSTANT:
-                return $this->memberReferences($filesystem, $symbolContext, ClassMemberQuery::TYPE_CONSTANT, $source, $replacement);
```
dantleech, line 277: could rename this method to f.e. `doPerformFindOrReplace` and decorate it with the sorting functionality.

lib/Extension/ClassMover/Rpc/ReferencesHandler.php
```diff
@@ -268,19 +268,46 @@ private function performFindOrReplaceReferences(
...
-            'Cannot find references for symbol type "%s"',
-            $symbolContext->symbol()->symbolType()
-        ));
+        return [$source, $this->sortReferences($references)];
+    }
+
+    public function sortReferences(array $fileReferences): array
+    {
+        // Sort the references for each file
+        \array_walk($fileReferences, function (array &$fileReference) {
```
dantleech, line 295: no need to use an FQN here?

dantleech:
CI should work again now - had to migrate from travis-ci.org to travis-ci.com

dantleech:
btw I'd sooner replace this functionality with the new [reference finder as used by the languaeg server](https://github.com/phpactor/language-server-phpactor-extensions/blob/d5c6c4d3d5f3e8683ea30a8908c348034517de7d/lib/LanguageServerReferenceFinder/Handler/ReferencesHandler.php#L26) finally - ideally if the abstraction could also fall back to the non-indexed strategy here. But first it should also support search+replace.

lib/Extension/ClassMover/Rpc/ReferencesHandler.php
elythyr: All the references reference the same name so I'm not sure on which FQN we could sort.
The only things we are provided with are the file path and the positions.
So we sort the references in each files by position and then all files by alphabetical order.

lib/Extension/ClassMover/Rpc/ReferencesHandler.php
elythyr: Done, if I understood what you meant ;)

elythyr:
> btw I'd sooner replace this functionality with the new reference finder as used by the languaeg server finally - ideally if the abstraction could also fall back to the non-indexed strategy here. But first it should also support search+replace.
There is already a PR for the LSP extension: https://github.com/phpactor/language-server-phpactor-extensions/pull/5

So switching will not change anything regarding the sorting.

elythyr:
I can't figure out why the CI tests don't pass.
I removed my local packages, clean my vendors, install the packages from the lockfile, symlink the vendor and the lock file to the indexer-extension (as done in the `testdeps` script) and all tests passes...

lib/Extension/ClassMover/Rpc/ReferencesHandler.php
dantleech: i mean the `array_walk` doesn't need to be absolutely referenced?

dantleech: also should be private?

elythyr: ho yes, bad habit from work with the `\` and public well... It seems I'm out of my game :)

elythyr:
Up here too, to have the same behavior with LSP and RPC

Issue #1110, opened 2020-10-12
Error in install


araujo0205:
Hello, 
I installed in vim base on this [link](https://phpactor.readthedocs.io/en/master/usage/vim-plugin.html)
After plug installed, I open a php file and execute `:PhpactorStatus`. But return this error:
```
executing job failed: No such file or directory
```
Any suggestions ?

I using Mac 10.15.7 and my vim is
VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Oct  7 2020 23:36:35) instaled by homebrew.

elythyr:
There is a bug on the command `:PhpactorStatus` at the time, can you try to execute this commande instead: `:call phpactor#rpc("status", {'type': 'formatted'})`

araujo0205:
Thanks, this work for me.

elythyr:
Great, can you close the issue then please ?

araujo0205:
is good update the docs with the new command ?

dantleech:
hmm, not sure we should close this. `:PhpactorStatus` should work. Probably we should revert the `terminal` change (sure we had a ticket about that somewhere)

elythyr:
Yes we have, I will take care of it this week-end.

elythyr:
Fix merged, we can close.

Issue #1085, opened 2020-07-07
vim plugin not working with option `set autochdir`  or when launching vim in a subfolder 


allan-simon:
I'm using neovim and I have in my init.vim the line `set autochdir` , and it seems to not play nice with phpactor 

is it possible for phpactor to try to find in the parent folders if there's a `.git` folder or composer file in order to  choose the root of the project  (like many other command line utilities do )  ?

dantleech:
Phpactor doesn't support this directly but you can provide an anonymous function to detect the project root by `g:PhpactorRootDirectoryStrategy` (`:help phpactor` or https://phpactor.readthedocs.io/en/develop/vim-plugin/man.html).

Maybe @przepompownia or @elythyr can recommend a way to use this in conjunction with a root-detection strategy?

allan-simon:
@dantleech  thanks for the quick answer, as I'm not familiar with anonymous function in vimscript, do you have a quick example of a dirty workaround that would just return always the same root path ?

dantleech:
Me neither in complete honesty, `:help Funcref`

I assume something like `g:PhpactorRootDirectoryStrategy  = function  () { return getcwd() }` for example.

allan-simon:
ok thanks, I've seen the "expr-lambda" help but adding in my init.vim

```
let g:PhpactorRootDirectoryStrategy = {-> '/home/allan/myproject' }
```

does not seem to work

`:Phpactorstatus` still return the same working directory :(

dantleech:
You'd need to restart VIM of course ?

allan-simon:
haha, I did :)

przepompownia:
@dantleech restarting is not needed. The function is executed each time we call to RPC. @allan-simon do you have the version that uses `let g:PhpactorRootDirectoryStrategy`? `:Phpactorstatus` or `:checkhealth` should print both the version and the working directory.

przepompownia:
One detail: if you have set the buffer equivalent `b:PhpactorRootDirectoryStrategy` then it takes precedence over the global. 

I noticed that we have undocumented this possibility.

allan-simon:
```
Version: 111c3e6 (4 days ago) 0.16.1-1083
Filesystems: simple
Working directory: /vagrant/src
```

I've tripled-checked , I have only ` g:PhpactorRootDirectoryStrategy` defined

Here's my init.vim
```
set mouse=a
" workaround https://github.com/neovim/neovim/issues/6041
set guicursor=
[...]

PR #1112, opened 2020-10-17
fix: phpactor#Status()

### LINKED ISSUE #1073: :PhpactorStatus doesn't use the bin path
`:PhpactorStatus` shows the status in the `term://` in neovim, however it does not use the `php` bin path, which means it doesn't work for anybody that actually runs < php 7.3 by default.

And possibly - it's using `phpactor` only if it's in the global path?

cc @elythyr

elythyr:
Fix #1073 
Relates to #1085 and #1110 

The issue was that the function was always using `php` to start PHP and the `phpactor` executable must be in the `PATH`.
Now the function use `g:phpactorPhpBin` to have the right `php` binary (like if the user wants to use a specific PHP version)
And `g:phpactorbinpath` to locate the `phpactor` binary properly.

### DIFF
--- a/autoload/phpactor.vim
+++ b/autoload/phpactor.vim
@@ -303,12 +303,13 @@ endfunction
 
 function! phpactor#Status()
     if exists(':terminal')
+        let l:cmd = g:phpactorPhpBin . ' ' . g:phpactorbinpath . ' status'
         if has('nvim')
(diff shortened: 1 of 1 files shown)


dantleech:
thanks!

elythyr:
You're welcome, that was my mistake anyway ;)

Issue #1113, opened 2020-10-21
Index files outside project directory


norival:
Hi!

I was wondering if it is possible to index files outside project directory in order to have completion and code navigation. I'm working on projects that depend on other php files that are in a different location (these projects are not using composer).

Everyone here is using phpstorm and it is possible to add these files as 'External librairies' to have autocompletion of classes and methods and code navigation.

Thanks! :)

elythyr:
Hi,

It's currently not possible.
I guess we could add this feature, like we did in order to index the stubs, but I'm not sure it's a good idea right now.
The issue is that I'm almost certain we still have features that do not rely on the index.

So if we add this possibility we will have inconsistent behavior because the features relying on the index will be aware of the external code while other features won't.

@dantleech You have a better vision of project as a all, do you think it would behave properly ?
Maybe we can add the feature but specify that it is only supported when used as a LSP server ?
Even though some RPC features start using the indexer so we will still have the same issue.

I realize that the [indexer configuration](https://phpactor.readthedocs.io/en/develop/reference/indexer.html#configuration) as not been updated, only the [configuration reference](https://phpactor.readthedocs.io/en/develop/reference/configuration.html#indexer-stub-paths)
Also, it's not possible to override or change the value of `indexer.stub_paths` because of:
https://github.com/phpactor/phpactor/blob/6ee1603677757a1a9d42dfbd9f924347d6196a3e/lib/Phpactor.php#L79
I guess that if we don't want it to be configurable by the user we should not expose it in the configuration, since it's made as a list I think you intended it to be configurable.

So to resume:
* Is it possible to add external sources to the indexer without having a strange behavior ?
* If so, maybe we should rename the `stub_paths` to something like `external_sources`
* Should the user be able to add its own stub paths ? (to fix the current behavior)

dantleech:
Yes, adding it to stub paths would accomplish what is required I think. Maybe it should be renamed, and maybe the phpstorn stubs can be added programatically rather than via config

norival:
Hi! Thank you for your answers and all the details :) It would be great to have this feature. Is there any way I can help?

dantleech:
For now you could try setting:

```
{
    "indexer.stub_paths": [
        "/absolute/path/to/phpactor/vendor/jetbrains/phpstorm-stubs",
        "/path/to/your/external/files"
    ]
}
```

elythyr:
[...]

PR #1122, opened 2020-11-03
Do not overwrite stub_paths config if key exists

norival:
I'm not sure this is the right way to do it.
See #1113

### DIFF
--- a/lib/Phpactor.php
+++ b/lib/Phpactor.php
@@ -76,12 +76,8 @@ class Phpactor
         $config = $loader->load();
         $config[CoreExtension::PARAM_COMMAND] = $input->getFirstArgument();
         $config[FilePathResolverExtension::PARAM_APPLICATION_ROOT] = self::resolveApplicationRoot();
-        $config[IndexerExtension::PARAM_STUB_PATHS] = isset($config[IndexerExtension::PARAM_STUB_PATHS])
-            ? array_push(
-                $config[IndexerExtension::PARAM_STUB_PATHS],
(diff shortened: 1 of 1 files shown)


lib/Phpactor.php
```diff
@@ -76,9 +76,12 @@ public static function boot(InputInterface $input, string $vendorDir): Container
...
         $config[CoreExtension::PARAM_COMMAND] = $input->getFirstArgument();
         $config[FilePathResolverExtension::PARAM_APPLICATION_ROOT] = self::resolveApplicationRoot();
-        $config[IndexerExtension::PARAM_STUB_PATHS] = [
+        $config[IndexerExtension::PARAM_STUB_PATHS] = isset($config[IndexerExtension::PARAM_STUB_PATHS])
+            ? array_push(
+                $config[IndexerExtension::PARAM_STUB_PATHS],
                 self::resolveApplicationRoot() . '/vendor/jetbrains/phpstorm-stubs'
-        ];
+            )
+            : [ self::resolveApplicationRoot() . '/vendor/jetbrains/phpstorm-stubs' ];
```
dantleech, line 84:
```
$config[IndexerExtension::PARAM_STUB_PATHS] = array_merge([
     IndexerExtension::PARAM_STUB_PATHS = [],
], $config);
$config[IndexerExtension::PARAM_STUB_PATHS][] = self::resolveApplicationRoot() . '/vendor/jetbrains/phpstorm-stubs';
```

maybe better? finally we should do something else, but this is OK for now.

norival: Yes, better :)

dantleech:
Thanks!

PR #1127, opened 2020-11-15
Fix/#1085 status does not use correct working directory

elythyr:
Since I have no memory, as my wife love to remind me, on my previous fix I forgot to deal with the missing `--working-dir` option.

This PR fix that, when testing it I realized that the candidates we add to identify config files at the root of the project were still using `getcwd` instead of the command line option if provided, so I fixed that too.

### DIFF
--- a/autoload/phpactor.vim
+++ b/autoload/phpactor.vim
@@ -303,7 +303,12 @@ endfunction
 
 function! phpactor#Status()
     if exists(':terminal')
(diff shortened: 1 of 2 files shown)


elythyr:
@dantleech the CI breaks when starting `config/travis/testexts`
I don't get the problem, both `/composer.json` and `/extensions/extensions.json` have `"minimum-stability": "dev"`

dantleech:
Thanks, I think I fixed the build :crossed_fingers: (not related to this PR)

PR #1131, opened 2020-11-22
2020 11 22

dantleech:
```
    [2020-11-21 10:04:32] a2789e00 web-flow fix properties inferred types (#73)  * fix properties inferred types    * fix smoke test error on the CI    * updat...
    [2020-11-21 17:31:11] b8596002 web-flow Extract properties and methods from class' traits (#74)
    [2020-11-22 19:30:32] af58fccb web-flow Test performance regressions
    with PHPBench
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,9 @@ Improvements:
   - [language-server] Support document symbols (f.e. showing code outline for document)
   - [language-server] Support symbol highlighting
   - [code-tranform] Generated accessors automatically `ucfirst` the property name when prefix is used. - @einenlum
+  - [worse-reflection] Improved inference for property types - @elythyr
+  - [worse-reflection] Include virtual members from traits - @scisssssssors
+  - [code-tranform] fix invalid missing property diagnostic (#1126) - @elythyr
 
 Bug fixes:
 
--- a/composer.lock
+++ b/composer.lock
@@ -959,6 +959,10 @@
                 }
             ],
             "description": "Composer plugin that provides efficient querying for installed package versions (no runtime IO)",
+            "support": {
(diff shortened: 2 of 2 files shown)


dantleech:
image

@camilledejoye although the `$this` issue seems fixed, properties still do not have inferred types -- is this what you expect?

camilledejoye:
I noticed this too but I was not sure how to interpret it.
The type show here is the "best" type I guess, we only return one type to the LSP client.
But if you try to complete `$this->foo->` the public members from `FooBar` are shown.

dantleech:
hmm, I don't get any public methods there.
```
class FooBar implements Foo
{
    public function bar() {}
}

class Bar
{
    private Foo $foo;

    public function __construct(Foo $foo)
    {
        $this->foo = $foo;
    }

    public function foo(): void
    {
        if (!$this->foo instanceof FooBar) {
            return;
        }

        $this->foo->
        // $this is now inferred as FooBar
    }
}
```

dantleech:
Oh ignore that -- the class was implementing a non-existing class. Indeed - I get the public methods.

dantleech:
I guess something is missing. 
image

i guess we use class reflection to get the property types when walking assignments etc.

dantleech:
Let's merge it - it fixes a bug anyway and not sure how easy it would be to fix - probably it needs a wider architectural solution.

camilledejoye:
I think the current fix is enough for now, it covers most of the use cases we will encounter.
So I would not remove it since it does not (yet at least ;) ) bring any bug.
And I agree that some big changes would be needed to make it work in all cases, basically changing how we deal with types in general in the worse-reflection I guess...
Scared just to think about it :smile:

dantleech:
I'm scared about getting everything working on PHP8 and migrating from Travis which is only getting worse it seems.

Issue #1136, opened 2020-12-01
RPC Open File regression via. #987


dantleech:
Introduced in https://github.com/phpactor/phpactor/pull/987

When using `:PhpactorNavigate` we get the following error:

![image](https://user-images.githubusercontent.com/530801/100715149-ade93500-33ae-11eb-9391-0fd4e8a64460.png)

Due to the `Phpactor\Extension\Rpc\Response\OpenFileResponse`:

```
    const TARGET_FOCUSED_WINDOW = 'focused_window';
    const TARGET_VERTICAL_SPLIT = 'vsplit';
    const TARGET_HORIZONTAL_SPLIT = 'hsplit';
    const TARGET_NEW_TAB = 'new_tab';

```

Updating the targets however causes another problem:

```
Error detected while processing function phpactor#Navigate[1]..phpactor#rpc[24]..phpactor#_rpc_dispatch[124]..phpactor#_rpc_dispatch_input[43]..phpactor#input#choice[30]..phpactor#_rpc_dispatc
h_input_handler[3]..phpactor#rpc[24]..phpactor#_rpc_dispatch[56]..<SNR>164_openFileInSelectedTarget:
line   24:
E481: No range allowed: 0 edit spec/xxx/Bundle/StoreBundle/Mailer/Sender/SenderSpec.php
Error detected while processing function phpactor#Navigate[1]..phpactor#rpc:
line   24:
E171: Missing :endif
```

dantleech:
cc @camilledejoye

camilledejoye:
I'll try to look into it today

dantleech:
cool thanks, if you don't get a chance today I'll try and investigate myself or revert the PR temporarily

PR #1137, opened 2020-12-01
Fix OpenFile regression #1136

camilledejoye:
Fix regression #1136 introduced by #987


dantleech:
Great, was going to suggest making it backwards compatible - earlier I updated the OpenFile RPC action to use the new methods but it failed with another error - seems to work with this though :bow:

dantleech:
No just to wait a day or two for the build to start :cry:

dantleech:
Thanks

PR #1139, opened 2020-12-06
Add github actions

dantleech:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,147 @@
...
+              uses: shivammathur/setup-php@2.7.0
+              with:
+                  coverage: "none"
+                  php-version: "${{ matrix.php-version }}"
+                  tools: composer:v2
+            -
+                name: "Composer install"
+                uses: "ramsey/composer-install@v1"
+                with:
+                    composer-options: "--no-scripts"
```
Taluu, line 38: imo there's no need for this action to be used, just use `composer install --no-scripts`. Unless you really want to keep the cache ?

dantleech: I don't have much of a clue about this (this is mostly copied from PHPBench which was contributed by @OskarStark ). But isn't using the cache beneficial here?

Taluu: Hum, as you commit the composer.lock it may make sense.

OskarStark: Initially I translated Travis.yaml to github actions in phpbench and it was part of the Travis.yaml 😉

If it is working, let's keep it for now and maybe check afterwards in a stand-alone PR what will be the difference

.github/workflows/ci.yml
```diff
@@ -0,0 +1,147 @@
...
+
+            -
+                name: "Run phpstan/phpstan"
+                run: "vendor/bin/phpstan analyse --level=1 lib"
```
OskarStark, line 42: Maybe better to add a phpstan.neon.dist file and use this config file here?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,147 @@
...
+        runs-on: ubuntu-latest
+
+        continue-on-error: ${{ matrix.allow-failures }}
+
+        strategy:
+            matrix:
+                php-version: 
+                    - '7.3'
+                    - '7.4'
+                allow-failures: [false]
```
OskarStark, line 56: Can be removed if all are not allowed to fail 

Same for the other jobs

.github/workflows/ci.yml
```diff
@@ -0,0 +1,147 @@
...
+            -
+                name: "Run phpstan/phpstan"
+                run: "vendor/bin/phpstan analyse --level=1 lib"
+
+    PHPUnit:
+        name: "PHPUnit ${{ matrix.php-version }} + ${{ matrix.dependency }}"
+
+        runs-on: ubuntu-latest
+
+        continue-on-error: ${{ matrix.allow-failures }}
```
OskarStark, line 49: Can be removed in case all are not allowed to fail, this is false by default

.github/workflows/ci.yml
```diff
@@ -0,0 +1,147 @@
...
+              with:
+                  coverage: "none"
+                  php-version: "${{ matrix.php-version }}"
+                  tools: composer:v2
+
+
+            - name: "Composer install"
+              uses: "ramsey/composer-install@v1"
+              with:
+                  dependency-versions: "${{ matrix.dependency }}"
```
OskarStark, line 73: dependency variable is not defined

.github/workflows/ci.yml
```diff
@@ -0,0 +1,147 @@
...
```
OskarStark, line 106: Dependency variable not defined

dantleech:
Thanks, just for reference the contents of this PR is a POC for [a project I'm working on](https://github.com/dantleech/maestro2/blob/5b5c44ddccfa2e41b3bf3ae2bd8c95feb6d27ccf/example/pipeline/GithubActionsPipeline.php). I will make a better version of the template soon.

OskarStark:
You maybe want to check: https://github.com/sonata-project/dev-kit

We are managing all our sonata bundles

dantleech:
Indeed :) the idea is similar I think

PR #1217, opened 2021-03-03
Add regression test for `ClassSearch` including non-PHP files

### LINKED ISSUE #1216: Non-class results from `ClassSearch` slowing down `phpactor#UseAdd()`
In my project, test reports are saved by date/time in subdirectories of */tests/reports/{coverage,metrics}*.  
When I call `phpactor#UseAdd()`, it [also scans those reports][1] because */tests* is listed in composer's PSR 4 `autoload-dev` list.

If I trigger `UseAdd()` on a `User` class (for example), I will get the file I want *plus* two for each report (there's both a `User` and `System\User`). The end result is some 200 choices split across several pages, which takes over *4 seconds* to load.

---

My first thought was to exclude the path. I found #782, but then I noticed **the results aren't PHP files**, let alone PHP classes. They just *happen* to match the path and filename of the real class... with a `.html` extension tacked on.

After some digging, I found the `classSearch` method ([called][2] by [`ImportClassHandler`][3]) which uses the `named` filter:

https://github.com/phpactor/phpactor/blob/dac62ddadb2f5ed3127621cb28358957f613bc1a/lib/Extension/SourceCodeFilesystemExtra/SourceCodeFilestem/Application/ClassSearch.php#L41-L42

...but a call to `->named('Foo.php')` *[isn't an exact match][4]* - it effectively filters on `"/Foo.php*"`.

---

As far as I can tell, when performing a class search we're *only* interested in `.php` files, *not* `.php.bak` or `.php.html` etc. so we can just compare the full filename. I'll send a PR shortly with a fix, but it still needs cleaning up so in the meantime I have a couple...

## Questions

* Is there anywhere (other than `ImportClassHandler`) that uses `classSearch()` and relies on the current functionality?
* Instead of a closure, a new method on `FileList` would probably be better:
  * Since it's a "stricter `named()`", thoughts on `->namedExactly('Foo.php')` or similar?

[1]: https://github.com/phpactor/source-code-filesystem/blob/master/lib/Adapter/Composer/ComposerFileListProvider.php#L69-L71
[2]: https://github.com/phpactor/phpactor/blob/dac62ddadb2f5ed3127621cb28358957f613bc1a/lib/Extension/CodeTransformExtra/Rpc/ImportClassHandler.php#L153-L156
[3]: https://github.com/phpactor/phpactor/blob/dac62ddadb2f5ed3127621cb28358957f613bc1a/lib/Extension/CodeTransformExtra/Rpc/ImportClassHandler.php#L80
[4]: https://github.com/phpactor/source-code-filesystem/blob/b96d5b4014fbb6813c459a00d998e80c2bef9aaa/lib/Domain/FileList.php#L110-L116

dshoreman:
Adds a regression test for phpactor/source-code-filesystem#19 and updates the testdox around it.

---

**Original text:**

The `named()` method actually filters for "Foo*" which produces many
results that aren't PHP classes at all. This resolves the issue by
instead comparing against the *full* filename.

With this change, running `phpactor class:search 'User'` is close to 14
times faster, reduced down to ~0.3s from the 4 seconds it took before.

### DIFF
--- a/tests/Assets/Projects/Animals/lib/Badger.php.html
+++ b/tests/Assets/Projects/Animals/lib/Badger.php.html
@@ -1,0 +1,4 @@
+<p>This file emulates a coverage report to ensure only PHP files are found by ClassSearch.</p>
+<p>See Phpactor\Tests\System\Extension\SourceCodeFilesystem\Command\ClassSearchCommandTest</p>
+
+<p>Context: https://github.com/phpactor/phpactor/issues/1216</p>
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,7 @@
         "phpactor/reference-finder-extension": "^0.1.7",
         "phpactor/reference-finder-rpc-extension": "^0.1.5",
         "phpactor/rpc-extension": "^0.2.3",
-        "phpactor/source-code-filesystem": "^0.1.6",
+        "phpactor/source-code-filesystem": "^0.1.7",
         "phpactor/source-code-filesystem-extension": "^0.1.5",
         "phpactor/text-document": "^1.2.3",
         "phpactor/worse-reference-finder-extension": "^0.1.5",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "d37cd15970cabbd5fe0ba35b7cf1673d",
+    "content-hash": "55fd3c6e293b324c139ec34f98b7fd26",
     "packages": [
         {
             "name": "amphp/amp",
@@ -4111,16 +4111,16 @@
         },
         {
             "name": "phpactor/source-code-filesystem",
-            "version": "0.1.6",
+            "version": "dev-master",
             "source": {
                 "type": "git",
                 "url": "https://github.com/phpactor/source-code-filesystem.git",
-                "reference": "6ca0c64462b59c6d47f12fef5d4013229e346f68"
+                "reference": "860f83fd8df18be4cdbb5e3a1a3ae562afc3b923"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/phpactor/source-code-filesystem/zipball/6ca0c64462b59c6d47f12fef5d4013229e346f68",
-                "reference": "6ca0c64462b59c6d47f12fef5d4013229e346f68",
+                "url": "https://api.github.com/repos/phpactor/source-code-filesystem/zipball/860f83fd8df18be4cdbb5e3a1a3ae562afc3b923",
+                "reference": "860f83fd8df18be4cdbb5e3a1a3ae562afc3b923",
                 "shasum": ""
             },
             "require": {
@@ -4138,6 +4138,7 @@
                 "phpunit/phpunit": "^9.0",
                 "symfony/var-dumper": "^5.2"
             },
+            "default-branch": true,
             "type": "library",
             "extra": {
                 "branch-alias": {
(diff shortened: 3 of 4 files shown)


dantleech:
That's a pretty significant performance improvement!

I'd rather fix it in the [SCF library](https://github.com/phpactor/source-code-filesystem/blob/master/lib/Domain/FileList.php#L110)
```
    public function named(string $name): self
    {
        return new self(new RegexIterator($this->iterator, sprintf(
            '{/%s.*$}',
            preg_quote($name)
        )));
    }
```

looks strange to me (why does it match all chars after the given name (??)) - but the tests pass somehow. Maybe a new method is the way to go .. or ... I think you could do:

```
$filesystem->fileList(...)->filter(function (SplFileInfo $file) {
                return $file->getFileName() == 'Bar.php';
            })
```

dshoreman:
> I'd rather fix it in the [SCF library](https://github.com/phpactor/source-code-filesystem/blob/master/lib/Domain/FileList.php?rgh-link-date=2021-03-03T09%3A08%3A40Z#L110)
>
That's the plan, but doing it here first was an easy way to make sure CI passed. Once I've updated SCF, this PR will essentially only be updating tests.
 
> ```
>     public function named(string $name): self
>     {
>         return new self(new RegexIterator($this->iterator, sprintf(
>             '{/%s.*$}',
>             preg_quote($name)
>         )));
>     }
> ```
> 
> looks strange to me (why does it match all chars after the given name (??))

Exactly what confused me at first! I've been trying to think of possible other extensions that maybe should be allowed, but the only thing close is `.phpt` and those are a different format anyway.

> ...but the tests pass somehow. Maybe a new method is the way to go .. or ... I think you could do:
> 
> ```
> $filesystem->fileList(...)->filter(function (SplFileInfo $file) {
>                 return $file->getFileName() == 'Bar.php';
>             })
> ```

I had [tried that first](https://github.com/phpactor/phpactor/pull/1217/commits/efcbe94c6c78e6d302df4268141a0220dff84884) but unfortunately, we have to use the full path. If we don't, then searching with a FQCN will fail. Shame really... that would've brought the time down another 0.1s! :joy:

tests/Assets/Projects/Animals/lib/Badger.php.html
```diff
@@ -0,0 +1,26 @@
+<!DOCTYPE html>
+<html lang="en">
+ <head>
+  <meta charset="UTF-8">
+  <title>Code Coverage for /var/lib/Badger.php</title>
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+ </head>
+ <body>
+  <header>
+   <div class="container-fluid">
```
dshoreman, line 10: This could be shortened (or even an empty file) if you prefer

dantleech: Or maybe a line of text suggesting it's reason and referring to the test?

dshoreman: How's this?
```html
<p>This file emulates a coverage report to ensure only PHP files are found by ClassSearch.</p>
<p>See Phpactor\Tests\System\Extension\SourceCodeFilesystem\Command\ClassSearchCommandTest</p>

<p>Context: https://github.com/phpactor/phpactor/issues/1216</p>
```

I added the `<p>` tags just for the sake of formatting in case someone happens to double-click and it opens in a browser

dantleech: yes that's fine :)

dshoreman:
This is still pending a new version for SCF, but I've marked this ready for now since the composer update can always be done later

dantleech:
You can bump the SCF version number in composer.json and it will use dev-master (we use `minimum-stability: dev, prefer-stable: true`)

dshoreman:
Huh, neat. TIL! Thanks, I'll do that shortly

dshoreman:
@dantleech All done. Sorry about that :)

dantleech:
Thanks!

PR #1238, opened 2021-04-17
Rename class

dantleech:
This PR adds support for renaming classes (only the short names). It makes use of the existing legacy functionality, which simplifies maintainence and also makes integrating namespace / folder movement operations a bit easier as this already works in leagcy.

```
  phpactor/class-mover 374ee3d3..47166e1f

    [2021-04-17 15:21:24] 0d8d935f dantleech Migrating to use text edits
    [2021-04-17 15:35:34] 74053928 dantleech Use code updater to update classes
    [2021-04-17 15:39:03] 111f4b25 dantleech Bump dev to 2.0
    [2021-04-17 16:45:21] b80e83f4 dantleech Added extension
    [2021-04-17 16:51:38] 47166e1f dantleech Added test for container extension

  phpactor/language-server-phpactor-extensions f7f3ccd0..941fad52

    [2021-04-17 16:02:13] 02af2840 dantleech Adding class renamer
    [2021-04-17 16:36:21] 637a2257 dantleech Implement class renamer
    [2021-04-17 16:38:08] 68a19526 dantleech Applies CS fixes
    [2021-04-17 16:52:15] 654eb5e0 dantleech Register class renamer
    [2021-04-17 16:57:18] 941fad52 dantleech Updated phpbench config
```

cc @blademf

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Bug fixes:
 Features:
 
   - [language-server] Ensure workspace is indexed before finding references - @dantleech
+  - [language-server] Support for renaming class names (short only) - @dantleech
   - [language-server] Rename class members and variables - @BladeMF, @dantleech
   - [language-server] Basic support for workspace symbols.
   - [language-server] Added basic PHP linting by default.
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
         "composer/xdebug-handler": "^1.1",
         "dantleech/invoke": "^1.0.1",
         "monolog/monolog": "^1.0",
-        "phpactor/class-mover": "^0.1.4",
+        "phpactor/class-mover": "^0.2.0",
         "phpactor/class-to-file": "^0.4.0",
         "phpactor/class-to-file-extension": "^0.2.2",
         "phpactor/code-transform": "^0.4.0",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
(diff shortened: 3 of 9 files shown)


BladeMF:
@dantleech, a few questions:
1. do you have a list of the problems with that implementation (I remember you mentioned somewhere that it had some?)?
2. Does this return text edits to the LSP client or does it change the files directly?
3. Does it rename the file that the class is in?
4. Related to (3), what do we have left to implement to have full rename support (e.g., rename the file, detect aliased classes)?

dantleech:
> do you have a list of the problems with that implementation (I remember you mentioned somewhere that it had some?)?
no - there are issues with including/not including use statements in some special cases when classes are renamed to a new namespace. But that doesn't apply to this implementation.

> Does this return text edits to the LSP client or does it change the files directly?
It returns LSP edits

> Does it rename the file that the class is in?

No, and I'm not sure that's possible from the server side (we can't tell the client to close/open a file)

> full rename support

Basically we need to either _detect_ when a file is renamed, or the client needs to tell us. The `tsserver` will detect when a file is renamed and I think we can do the same (this will also allow you to move and entire directories and have all classes inside them updated).

I already started [experimenting](https://github.com/phpactor/language-server-phpactor-extensions/pull/34) with this

BladeMF:
> No, and I'm not sure that's possible from the server side (we can't tell the client to close/open a file), instead the client can rename the file.

It is possible. My implementation did this and it is relatively easy to do. We should probably just add the corresponding model to the Phactors's model (along with its TextEdit) and then convert once on the LSP side.

> I already started experimenting with this

I am pretty sure the client can tells us. I researched it, but didn't write it down anywhere, sadly. :-(

dantleech:
The client can tell us if an explicit `rename` happened on a file but *not a folder*. CoC/VIM does not provide the `willRename` event. So for now I'm going to try detecting it.

PR #1267, opened 2021-05-12
composer require --dev dms/phpunit-arraysubset-asserts

zonuexe:
It depends on unnecessary PHPUnit because the `dms/phpunit-arraysubset-asserts` package needed for development was mistakenly added to "require".

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -45,8 +45,7 @@
         "symfony/filesystem": "^4.2 || ^5.0",
         "symfony/yaml": "^4.3 || ^5.1",
         "thecodingmachine/safe": "~1.1.0",
-        "webmozart/glob": "^4.4",
-        "dms/phpunit-arraysubset-asserts": "^0.2.1"
+        "webmozart/glob": "^4.4"
     },
     "require-dev": {
         "dantleech/what-changed": "~0.4",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -60,7 +59,8 @@
         "symfony/process": "^2.7 || ^3.0 || ^4.0",
         "symfony/var-dumper": "^5.0",
         "sebastian/diff": "^4.0",
-        "phpspec/prophecy-phpunit": "^2.0"
+        "phpspec/prophecy-phpunit": "^2.0",
+        "dms/phpunit-arraysubset-asserts": "^0.3.0"
```
dantleech, line 63: we test the all the phpactor dependencies using this lock file, so easier to make this `^0.2.1`

dantleech:
whoops :flushed: it needs to be at the previous dep version for the tests to pass as noted above

composer.json
zonuexe: I'm sorry.  Fixed and force pushed https://github.com/phpactor/phpactor/commit/28a3869a1ab94b41b659ac5e7f0015b8f64744cb

dantleech:
Thanks :)

PR #1305, opened 2021-07-01
Pass basePath to PathFinder

wjzijderveld:
Make use of new named constructor in https://github.com/phpactor/path-finder/pull/11 
I noticed a failing tests, but I'll have to look into those at a later moment, hence the draft.


dantleech:
@wjzijderveld the path-finder lib has no been merged into this repo now..

PR #1308, opened 2021-07-14
Return 0 on ShutdownServer

### LINKED ISSUE #1307: Error 255 on every client exit
I am not sure whether this happens by intent or not. It seems like Phpactor returns a 255 code on exiting the language server.

My Setup
* NeoVim
* Phpactor via https://github.com/neovim/nvim-lspconfig/blob/master/CONFIG.md#phpactor

Steps to reproduce
* Open any php file
* Exit the editor

The shell displays `Client 1 quit with exit code 255 and signal 0`.

Phpactor Log (debug level) says

```
2021-07-14 08:43:27] phpactor.INFO: IN: {"method":"shutdown","jsonrpc":"2.0","id":2} []
[2021-07-14 08:43:27] phpactor.INFO: OUT:  {"id":2,"result":null,"error":null,"jsonrpc":"2.0"} []
[2021-07-14 08:43:27] phpactor.INFO: IN: {"method":"exit","jsonrpc":"2.0"} []
[2021-07-14 08:43:27] phpactor.CRITICAL: Exit called on STDIO connection, exiting the server [] []
```

In my opinion `Exit called on STDIO connection, exiting the server` shouldn't be `CRITICAL` is something to expect and it shouldn't be treated as critical.

Once this has been clarified I could have a look at the code.

weeman1337:
Couldn't find any simple way testing that :( 

closes #1307

### DIFF
--- a/lib/Application.php
+++ b/lib/Application.php
@@ -3,6 +3,7 @@
 namespace Phpactor;
 
 use Phpactor\Extension\Logger\Formatter\PrettyFormatter;
+use Phpactor\LanguageServer\Core\Server\Exception\ShutdownServer;
 use Symfony\Component\Console\Application as SymfonyApplication;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
@@ -52,6 +53,8 @@ class Application extends SymfonyApplication
 
         try {
             return parent::doRun($input, $output);
+        } catch (ShutdownServer $e) {
+            return 0;
         } catch (Exception $e) {
             if (
                 $input->hasArgument('command')


dantleech:
Sorry I merged this but it's wrong.

dantleech:
Reverted. The Application shouldn't know about the language server.

Instead you need to add the try/catch to the `StartCommand` in `language-server-extension`, I think.

dantleech:
Also, it's likely the behavior should really be that the server loop simply finishes on ServerException, then we don't need to try/catch anything. In which case this belongs in the `language-server` package

weeman1337:
Yeah - looked to easy anyway to be the proper solution :)

PR #1330, opened 2021-12-06
[Docs] Prerequisites PHP 7.3

gpoma:
PHP 7.2 was in the prerequisites but later in the file it states that you need at least the 7.3


dantleech:
thanks!

PR #1360, opened 2022-03-13
Generate accessors

dantleech:
https://user-images.githubusercontent.com/530801/158082491-0a5ea9e9-f4a0-4c37-be0d-88081abcf8d2.mp4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Changelog
 
 Features:
 
+  - [lsp] Added generate accessors code action
   - [lsp] Added extract constant code action
   - [extension] Removed the extension manager.
   - [extension] PHPStan and Psalm extensions are now included by default.
--- a/doc/lsp/code-actions.rst
+++ b/doc/lsp/code-actions.rst
@@ -28,5 +28,9 @@ List of currently available code actions:
 +---------------------------------------------+---------------------------------------+
 | :ref:`generation_extract_expression`        | ``quickfix.extract_expression``       |
 +---------------------------------------------+---------------------------------------+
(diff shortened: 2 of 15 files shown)


grtcdr:
Woah! Thank you for adding this.

One question though, what command are you calling to show the list of accessors?

dantleech:
Thanks, this is invoking an LSP code action on a sélection. So depends which language server client you use (e.g. coc or neovim LSP, or whatever)

grtcdr:
I'm using neovim's builtin LSP support.

I'll read the docs, thanks for your help :smile:

gbprod:
You should use

```
:'<,'>lua vim.lsp.buf.range_code_action()
```
😉

grtcdr:
That worked very nicely, although I don't get a popup as shown in the video, instead I get prompted to enter a number through the minibuffer. My configuration only has the bare minimum, so I suspect that's why.

dantleech:
see these two config options: https://phpactor.readthedocs.io/en/master/reference/configuration.html#code-transform-refactor-generate-accessor-prefix :)

dantleech:
> I don't get a popup as shown in the video,
that's because I'm using CoC, haven't tried the neovim client before.

grtcdr:
> see these two config options

Thank you, I'll add it to my config.

> that's because I'm using CoC, haven't tried the neovim client before.

I see.

Thank you, once again, for all your help! :)

camilledejoye:
You can look towards either https://github.com/glepnir/lspsaga.nvim or https://github.com/nvim-telescope/telescope.nvim for the popup effect.
lspsaga was not updated for a while but a fork which is up to date exists.

PR #1361, opened 2022-03-17
Fix prod composer dependencies

weeman1337:
The current `master` state doesn't work if you run `composer i --no-dev`.

- `symfony/filesystem` is still in use. It is installed as a transitive dependency when installing with dev-dependencies.
- `symfony/process` is in use. Moving it to regular dependencies.


dantleech:
Sorry, missed this. it's already fixed.

PR #1365, opened 2022-03-19
Update the type system

dantleech:
This PR will update the worse-reflection Phpactor type system. The ultimate aim of supporting generics, but this PR will only update to a saner system where generics can be implemented.

- The monolthic `Type::<type>` class has been changed to an interface.
- The `Type::` static constructors are switched to `TypeFactory::`
- There is a `TypeUtil` which provides a bridge to the previous behavior (e.g. `$type->short()` is now `TypeUtil::short($type)`)
- Each type has a dedicated class (`StringType`, `IntType`, etc)
- Types decide if they accept other types (e.g. parameter type can accept string as an argument etc)
- There is a stub GenericClassType class which stands in for the previous fake `Collection<Foo>` (`Type::collection()`)
- Class types are no longer "static" and have an embeded class reflector service, which will facilitate generics (or at least make it easier)


lib/CodeTransform/Adapter/WorseReflection/Transformer/CompleteConstructor.php
```diff
@@ -66,15 +68,10 @@ public function transform(SourceCode $source): TextEdits
...
-                    $className = $parameterType->className();
-                    if ($className) {
-                        $typeName = $class->scope()->resolveLocalName($className)->__toString();
-                    }
-
-                    if ($parameterType->isNullable()) {
-                        $typeName = '?' . $typeName;
+                if (TypeUtil::isDefined($parameterType)) {
+                    $typeName = TypeUtil::short($parameterType);
+                    if ($parameterType instanceof ClassType) {
```
dantleech, line 73: could be a `NullableType`

lib/Completion/Bridge/TolerantParser/WorseReflection/WorseClassMemberCompletor.php
```diff
@@ -98,25 +101,25 @@ public function complete(Node $node, TextDocument $source, ByteOffset $offset):
 
     private function populateSuggestions(SymbolContext $symbolContext, Type $type, bool $static, bool $completeOnlyName): Generator
     {
-        if (false === $type->isDefined()) {
+        if (false === TypeUtil::isDefined($type)) {
             return;
         }
 
-        if (null === $type->className()) {
+        if (!$type instanceof ClassType) {
```
dantleech, line 108: could be nullable

lib/Completion/Bridge/TolerantParser/WorseReflection/WorseParameterCompletor.php
```diff
@@ -161,11 +155,16 @@ private function reflectFunctionLike(TextDocument $source, Node $callableExpress
         $offset = $this->reflector->reflectOffset($source, $callableExpression->getEndPosition());
 
         if ($containerType = $offset->symbolContext()->containerType()) {
-            try {
-                $containerClass = $this->reflector->reflectClassLike($containerType->className());
-            } catch (NotFound $notFound) {
+            if (!$containerType instanceof ReflectedClassType) {
```
dantleech, line 158: nullable!

PR #1379, opened 2022-03-24
Complete class renaming

### LINKED ISSUE #1366: LSP rename: missing `RenameFile` operation in `documentChanges`
Currently when I try to rename some class form LSP client I have to rename its file manually. For a simple class with no references the server sends to the client `WorkspaceEdit` with a single `TextDocumentEdit` (as I guess from its format) inside `documentChanges`. It's enough to change the file content by some client's response handler. The original name remains untouched. 

After dumping such example edit in Lua format
```lua
{ -- WorkspaceEdit
  documentChanges = {
  { -- TextDocumentEdit
      edits = {
      {
          newText = "Arctgx\\PhpactorTestEnv",
          range = {
            ['end'] = {
              character = 32,
              line = 2
            },
            start = {
              character = 10,
              line = 2
            }
          }
        },
      {
          newText = "V",
          range = {
            ['end'] = {
              character = 7,
              line = 4
            },
            start = {
              character = 6,
              line = 4
            }
          }
        }
      },
      textDocument = {
        uri = "file:///home/arctgx/dev/arctgx/phpactor-test-env/src/Z.php",
        version = 7
      }
    },
  },
}
```
and adding `RenameFile` and applying such simulated response on the client side
```lua
local out = { -- WorkspaceEdit
  documentChanges = {
    { -- RenameFile
        kind = 'rename',
        oldUri = "file:///home/arctgx/dev/arctgx/phpactor-test-env/src/Z.php",
        newUri = "file:///home/arctgx/dev/arctgx/phpactor-test-env/src/V.php",
      },
  { -- TextDocumentEdit
      edits = {
      {
          newText = "Arctgx\\PhpactorTestEnv",
          range = {
            ['end'] = {
              character = 32,
              line = 2
            },
            start = {
              character = 10,
              line = 2
            }
          }
        },
      {
          newText = "V",
          range = {
            ['end'] = {
              character = 7,
              line = 4
            },
            start = {
              character = 6,
              line = 4
            }
          }
        }
      },
      textDocument = {
        uri = "file:///home/arctgx/dev/arctgx/phpactor-test-env/src/V.php",
        version = 7
      }
    },
  },
}
vim.lsp.util.apply_workspace_edit(out, vim.lsp.get_client_by_id(1).offset_encoding)
```
I get the expected change. It suggests that `RenameFile` should be added to `documentChanges`.

przepompownia:
Closes #1366


przepompownia:
What do you think about this approach? 

I started with yielding `RenameFile` directly but that approach required changing existing objects, creating some new complex or using `instanceof`.

No unit tests are provided yet.

lib/Extension/LanguageServerRename/Adapter/ReferenceFinder/ClassMover/ClassRenamer.php
```diff
@@ -13,18 +13,23 @@
 use Microsoft\PhpParser\ResolvedName;
 use Phpactor\ClassMover\ClassMover;
 use Phpactor\ClassMover\Domain\Name\QualifiedName;
+use Phpactor\Extension\Core\Application\Helper\ClassFileNormalizer;
```
dantleech, line 16: would be better not to couple to this class here (most anything here is legacy). Instead you can use the adapter in this package: ClassToFileUriToNameConverter

lib/Extension/LanguageServerRename/Model/LocatedTextEdit.php
```diff
@@ -8,12 +8,14 @@
...
     private TextDocumentUri $documentUri;
+    private TextDocumentUri $newDocumentUri;
 
     private TextEdit $textEdit;
 
-    public function __construct(TextDocumentUri $documentUri, TextEdit $textEdit)
+    public function __construct(TextDocumentUri $documentUri, TextEdit $textEdit, ?TextDocumentUri $newDocumentUri)
     {
         $this->documentUri = $documentUri;
+        $this->newDocumentUri = $newDocumentUri;
```
dantleech, line 18: I think this is bending the resposibility of this class a bit...

dantleech:
Some other approaches:

- Return a `RenameResult` class from `Renamer#rename`. `$result->locatedEdits(); $result->renameOperations()`
- Return a `RenameResult` from the `Renamer#rename` _Generator_. `foreach ($g as $r) { /.. } $result = $g->getReturnValue();`

Not sure if the would be cleaner or solve the problems you talked here in a better way.

przepompownia:
Indeed, we don't need to modify `LocatedTextEdit(s)`, we can pass rename object(s) directly into the converter.

Maybe `LocatedTextEditConverter` should be renamed to something more appropriate since it gets `Rename`.

lib/Extension/LanguageServerRename/Util/LocatedTextEditConverter.php
```diff
@@ -38,6 +40,11 @@ public function toWorkspaceEdit(LocatedTextEditsMap $map): WorkspaceEdit
                 )
             );
         }
+        $documentEdits[] = new RenameFile(
```
dantleech, line 43: this should be conditional i think, not all renames involve a file rename

dantleech: (if rename old URL != new URL?)

dantleech:
nice! are you happy to add tests?

lib/Extension/LanguageServerRename/Util/LocatedTextEditConverter.php
przepompownia: Looking for this case I also added quick (currently silent i.e. no exception is thrown to catch by the handler and notify the client) return when the original name equals the new.

przepompownia:
Please look at (currently anonymous) `NameToUriConverter` implementation inside `ClassRenamerTest` is enough or if it's better to use real `ClassToFileNameToUriConverter` and `Workspace`.

dantleech:
seems to work great, thanks!

przepompownia:
Thanks for support and merging!

PR #1380, opened 2022-03-25
Fix class move issues

dantleech:
This PR fixes some long-standing (5 year) bugs with the class mover.

- It does not try and re-import imported classes
- Will not re-import aliased classes


lib/ClassMover/Adapter/TolerantParser/TolerantClassReplacer.php
```diff
@@ -33,9 +33,13 @@ public function replaceReferences(
         $edits = [];
         $importClass = false;
         $addNamespace = false;
+        $replacedImports = [];
 
         foreach ($classRefList as $classRef) {
-            $importClass = $this->shouldImportClass($classRef, $originalName);
```
dantleech, line 38: no idea how this worked so "well" for 5 years.

PR #1393, opened 2022-04-02
Improved docblocks and code generation

dantleech:
- When generating properties (complete constructor, add missing properties) use the documented type if available:

```
    /**
     * @var array<string,Resolver>
     */
    private array $resolverMap;
```

- Do not HTML-escape generated code (!)
- Various refactoring

### DIFF
--- a/lib/CodeBuilder/Domain/Prototype/Type.php
+++ b/lib/CodeBuilder/Domain/Prototype/Type.php
@@ -8,13 +8,10 @@ final class Type extends Prototype
 
     private bool $none = false;
 
-    private bool $nullable = false;
-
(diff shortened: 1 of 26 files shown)


lib/CodeBuilder/Domain/Prototype/Type.php
```diff
@@ -8,26 +8,20 @@ final class Type extends Prototype
 
     private bool $none = false;
 
-    private bool $nullable = false;
-
-    public function __construct(string $type = null, bool $nullable = false)
+    public function __construct(string $type = null)
```
dantleech, line 11: the `?` nullable operator is now included in the type's own string-casting

PR #1394, opened 2022-04-02
improvement: Support "class-string" type

dantleech:
Adds docblock-level support for `class-string` and adds a dedicated type for it (not used for type inference yet though)


lib/DocblockParser/Lexer.php
```diff
@@ -25,7 +25,7 @@ final class Lexer
         '=', // equals
         '(', ')', '\{', '\}', '\[', '\]', '<', '>', // brackets
         '\$[a-zA-Z0-9_\x80-\xff]+', // variable
-        '[a-zA-Z0-9_\\\]+', // label
+        '[-a-zA-Z0-9_\\\]+', // label
```
dantleech, line 28: labels cannot start with _ or -

PR #1395, opened 2022-04-02
Short name for signature and hover

dantleech:
Use short name for signature and hover help.


lib/Extension/ObjectRenderer/ObjectRendererBuilder.php
```diff
@@ -0,0 +1,224 @@
...
+use Psr\Log\NullLogger;
+use Twig\Environment;
+use Twig\Loader\FilesystemLoader;
+
+/**
+ * Object Renderer builder.
+ *
+ * A new instance of the builder is returned for each modification.
+ */
+final class ObjectRendererBuilder
```
dantleech, line 24: copied verbatim from the object-renderer library --- allowing Twig to be configured

PR #1402, opened 2022-04-03
Assert / Type Inference Refactoring

dantleech:
This PR aims to properly support combining types. It's currently a bit of a hot-mess.

---

A better way to do this might be to introduce a way for _resolvers_ to assert a type.

```
$nodeContext = $resolver->resolve($frame, $someBinaryExpression);
foreach ($nodeContext->assertions() as $assertion) {
    // { "variable": "$foobar", "type": "Foobar" }
}
```

This would simplifiy things, but at this point it's probably worth doing some prior refactorings:

- :heavy_check_mark: Introduce literal types (E.g. `StringLiteral("Foobar")`: Remove the `value` in the node context.
- :heavy_check_mark: Remove the Types collection from Phpactor (there would then just be `type()` and `inferredType()` in the APIs)
- Remove the existing "expression resolver". The "node context resolver" will be able to resolve expressions.


lib/WorseReflection/Tests/SelfTest/assert/assert-negate-instanceof1.test
```diff
@@ -0,0 +1,13 @@
+<?php
+
+class F {
+    /** @param Foo|Bar $foo */
+    public function foo($foo) {
```
dantleech, line 5: note: create a ticket as defining a simple `function` doesn't inherit the `@param` docblocks

PR #1404, opened 2022-04-03
Fix renaming class outside the current document

### LINKED ISSUE #1401: Class + File Rename: Always renames the open file, not the one where the class is defined
Class rename renames the referencing file and not the one where the class is defined:

```
// Foobar.php
class Foobar {}
```

// Barfoo.php
```
$foobar = new Foobar();
```

When renaming `Foobar` in `Barfoo` it renames the file `Barfoo.php` instead of `Foobar.php`.

cc @przepompownia

przepompownia:
Closes #1401

### DIFF
--- a/lib/Extension/LanguageServerRename/Adapter/ReferenceFinder/ClassMover/ClassRenamer.php
+++ b/lib/Extension/LanguageServerRename/Adapter/ReferenceFinder/ClassMover/ClassRenamer.php
@@ -81,7 +81,8 @@ final class ClassRenamer implements Renamer
 
         $originalName = $this->getFullName($node);
         $newName = $this->createNewName($originalName, $newName);
-        $newUri = TextDocumentUri::fromString($this->nameToUriConverter->convert($newName));
+        $oldUri = $this->nameToUriConverter->convert($originalName);
+        $newUri = $this->nameToUriConverter->convert($newName);
 
         if ($newName === $originalName->getFullyQualifiedNameText()) {
             return;
@@ -113,7 +114,7 @@ final class ClassRenamer implements Renamer
             }
         }
 
-        return new RenameResult($textDocument->uri(), $newUri);
+        return new RenameResult($oldUri, $newUri);
     }
 
     private function rangeText(TextDocument $textDocument, ByteOffsetRange $range): string


przepompownia:
It's still a draft without tests. 

It fails when the project has classmap and the new name contains the old name (e.g. `Old` -> `Old2`). The bug is rather outside this PR. See https://github.com/phpactor/class-to-file/blob/master/lib/Adapter/Composer/ComposerClassToFile.php#L105

lib/Extension/LanguageServerRename/Adapter/ReferenceFinder/ClassMover/ClassRenamer.php
```diff
@@ -81,7 +81,8 @@ public function rename(TextDocument $textDocument, ByteOffset $offset, string $n
 
         $originalName = $this->getFullName($node);
         $newName = $this->createNewName($originalName, $newName);
-        $newUri = TextDocumentUri::fromString($this->nameToUriConverter->convert($newName));
+        $oldUri = $this->nameToUriConverter->convert($originalName);
```
dantleech, line 84: it would probably be better to use the refector? `ClassReflector` as that that's a better indicator of where the class is defined:

```
try {
    $path = $this->reflector->reflectClass($originalName)->sourceCode()->path());
} catch (NotFound $notFound) {
   
   // do something
}
if ($path === null) {
   // do somethinh
}

$oldUri = TextDocumentUri::fromString($path);
```

dantleech:
> It fails when the project has classmap

i should probably bring that lib. into the monolith too to make it easier to maintain.

lib/Extension/LanguageServerRename/Adapter/ReferenceFinder/ClassMover/ClassRenamer.php
przepompownia: This reflector seems to be free from the mentioned bug but it cannot reflect anything nonexisting yet. `$newUri` cannot be obtained this way.

przepompownia:
```php
    public function beginsWith(string $prefix): bool
    {
        if ($prefix === $this->fullyQualifiedName) {
            return true;
        }

        if (0 !== strpos($this->fullyQualifiedName, $prefix)) {
            return false;
        }

        if (mb_substr($prefix, -1, 1) === '\\') {
            return true;
        }

        return mb_substr($this->fullyQualifiedName, mb_strlen($prefix), 1) === '\\';
    }
```
works for me, what do you think? Maybe it should have tests.

lib/Extension/LanguageServerRename/Adapter/ReferenceFinder/ClassMover/ClassRenamer.php
dantleech: yeah, you can only use it for the old one, but I thought that was where the issue was?

dantleech:
> works for me, what do you think? Maybe it should have tests.

guess it would need a PR on `phpactor/class-mover`. But is the bug related to this PR?

lib/Extension/LanguageServerRename/Adapter/ReferenceFinder/ClassMover/ClassRenamer.php
```diff
@@ -81,7 +85,7 @@ public function rename(TextDocument $textDocument, ByteOffset $offset, string $n
 
         $originalName = $this->getFullName($node);
         $newName = $this->createNewName($originalName, $newName);
-        $oldUri = $this->nameToUriConverter->convert($originalName);
+        $oldUri = $this->reflector->reflectClass($originalName->getFullyQualifiedNameText())->sourceCode()->uri();
```
dantleech, line 88: Technically this should be an adapter: `Adapter/WorseReflection/ClassSourceLocator.php` or sth.

dantleech, line 88: To be consistent there should be an adapter in `Adapter/WorseReflection/WorseNameToUriConverter` which implements `Model\NameToUriConverter`. You can then inject that instead.

lib/Extension/LanguageServerRename/Adapter/ReferenceFinder/ClassMover/ClassRenamer.php
przepompownia: Yes, it was the issue.

przepompownia:
image
It's not so easy to (find a time to) prepare a good bug report in some cases. I caught this message when creating `WorseNameToUriConverter`. Phpactor crashes with such message from some days.

przepompownia:
image
Usually it looks like searching not namespaced global built-in functions.

dantleech:
> It's not so easy to (find a time to) prepare a good bug report in some cases. I caught this message when creating WorseNameToUriConverter. Phpactor crashes with such message from some days.

is that related to this ticket, if not can you create a new issue? it looks related to the LSP protocol class ContentChangeEvent

przepompownia:
No, this PR is independent and the bug is present on master too.

przepompownia:
> is that related to this ticket, if not can you create a new issue? it looks related to the LSP protocol class ContentChangeEvent

Please look at https://github.com/phpactor/phpactor/issues/1423

lib/Extension/LanguageServerRename/Handler/RenameHandler.php
```diff
@@ -84,9 +87,9 @@ public function rename(RenameParams $params): Promise
                 }
 
                 return $this->resultToWorkspaceEdit($locatedEdits, $rename->getReturn());
-            } catch (CouldNotRename $couldNotRename) {
+            } catch (CouldNotRename | SourceNotFound | CouldNotConvertClassToUri $error) {
```
dantleech, line 90: hmm. how does this work on 7.4?

dantleech:
Closing

dantleech:
Replaced with #1424

lib/Extension/LanguageServerRename/Handler/RenameHandler.php
przepompownia: https://wiki.php.net/rfc/multiple-catch

PR #1410, opened 2022-04-06
Literal Types

dantleech:
This PR starts to add support for literals, meaning:

```
$foo = 'hello'; // "hello" instead of string
$foo = 1234; // 1234 instead of int
$foo = 12.2 // 12.4 instead of float
// etc
```

These types are sub-classes of the base scalar types.

Literal types allow us to use the main "node context resolver" to resolve expressions and remove the old `NodeContext#value()` property.

A fun and useless side effect of this will be that you will be able to do math with types and get the answer by hovering over the type:

```
$x = 10 * 10 / 2;
$x; // will have type 50
```

TODO:

- [ ] Some refactorings should use the general type and not the literal type, so probably need to add a `NodeUtil::generailize($type)` or similar.
- [ ] Remove the old `ExpressionResolver` (which operated solely on values) and resolve expressions in the same way we do other nodes.
- [ ] ...


lib/CodeTransform/Tests/Adapter/WorseReflection/Transformer/AddMissingPropertiesTest.php
```diff
@@ -201,7 +201,7 @@ class Foobar
                         public $hello;
 
                         /**
-                         * @var int
+                         * @var 1234
```
dantleech, line 204: will probably regret this

PR #1411, opened 2022-04-07
Array literals

dantleech:
This PR is introducing support for array literals and will try and remove `value()` on variables and the node context (as this can now be determined from literal types)

### DIFF
--- a/lib/WorseReflection/Core/Type/ArrayLiteral.php
+++ b/lib/WorseReflection/Core/Type/ArrayLiteral.php
@@ -25,6 +25,16 @@ class ArrayLiteral extends ArrayType implements Literal, Generalizable
 
     public function __toString(): string
     {
+        if (range(0, count($this->typeMap) - 1) === array_keys($this->typeMap)) {
+            return sprintf(
+                'array{%s}',
+                implode(',', array_map(
+                    fn (Type $type) => sprintf('%s', $type->__toString()),
+                    array_values($this->typeMap),
+                ))
+            );
+        }
+
         return sprintf(
             'array{%s}',
             implode(',', array_map(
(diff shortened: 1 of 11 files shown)


lib/WorseReflection/Core/Type/ArrayLiteral.php
```diff
@@ -0,0 +1,84 @@
...
+    public function __construct(array $typeMap)
+    {
+        $this->typeMap = $typeMap;
+        $this->keyType = new MixedType();
+        $this->valueType = $this->resolveValueType($typeMap);
+    }
+
+    public function __toString(): string
+    {
+        if (range(0, count($this->typeMap) - 1) === array_keys($this->typeMap)) {
```
dantleech, line 28: should probably simplyintroduce a `List` type and resolve this dual responsiblity

lib/WorseReflection/Core/Type/ArrayLiteral.php
```diff
@@ -0,0 +1,84 @@
...
+     */
+    private array $typeMap;
+
+    /**
+     * @param array<array-key,Type> $typeMap
+     */
+    public function __construct(array $typeMap)
+    {
+        $this->typeMap = $typeMap;
+        $this->keyType = new MixedType();
```
dantleech, line 22: should support the key type

lib/WorseReflection/Core/Type/ArrayLiteral.php
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    /**
+     * @param array<array-key,Type> $typeMap
+     */
+    private function resolveValueType(array $typeMap): Type
+    {
+        $valueType = null;
+        foreach ($typeMap as $type) {
+            $type = TypeUtil::generalize($type);
```
dantleech, line 71: add a test for this, could also support resolving to a non-generalized type e.g. `[1,1,1,1,1]` would have tpe `array<array-key,1>` but not sure if that's useful.

PR #1416, opened 2022-04-10
Type Combination

dantleech:
This PR adds:

- Support for function Stubs: e.g. `array_sum` can sums types (just for a POC)
- Resolver type Assertions: `is_null($foobar)` will indicate to the caller that $foobar is of type NULL.
- Type Combination `$foo = string|int|null; assert(!is_null($foo)); // $foo = string|int`

TODO:

- [x] If statements
- [x] Assert statements
- [x] Fix the binary expressoin resolver
- [x] Type narrowing
- [x] Type restoration with non-specific types
---


lib/WorseReflection/Core/Inference/Frame.php
```diff
@@ -99,11 +115,64 @@ public function name(): string
...
+
+        foreach ($typeAssertions->properties() as $typeAssertion) {
+            $original = $this->properties()->byName($typeAssertion->name())->lastOrNull();
+            $originalType = $original ? $original->type() : TypeFactory::undefined();
+            $variable = new Variable(
+                $typeAssertion->name(),
+                TypeCombinator::applyType($originalType, $typeAssertion->type()),
+                $typeAssertion->classType(),
+            );
+            $this->properties()->add($offset, $variable);
```
dantleech, line 146: should refactor to apply the same logic on both.

lib/WorseReflection/Core/Inference/Frame.php
```diff
@@ -99,11 +115,64 @@ public function name(): string
...
+        }
+
+        foreach ($this->locals()->greaterThan($before)->lessThanOrEqualTo($after) as $extra) {
+            if (isset($locals[$extra->name()])) {
+                continue;
+            }
+
+            // if variable was not present before $before, assign as missing
+            $this->locals()->add($after, $extra->withType(TypeFactory::undefined()));
+        }
```
dantleech, line 168: same

lib/WorseReflection/Core/Inference/NodeContext.php
```diff
@@ -121,4 +140,29 @@ public function scope(): ReflectionScope
...
+        if ($subject->symbol()->symbolType() === Symbol::PROPERTY) {
+            return $this->withTypeAssertion(TypeAssertion::property(
+                $subject->symbol()->name(),
+                $subject->containerType() ?: new MissingType(),
+                $type
+            ));
+        }
+        if ($subject->symbol()->symbolType() === Symbol::UNKNOWN) {
+            dump($subject->typeAssertions());
+            dump($subject->symbol());
```
dantleech, line 163: remove

lib/WorseReflection/Core/Inference/Resolver/BinaryExpressionResolver.php
```diff
@@ -32,15 +34,31 @@ public function resolve(NodeContextResolver $resolver, Frame $frame, Node $node)
...
         if (!is_string($operator)) {
             return $context;
         }
 
-        return $context->withType($this->walkBinaryExpression($resolver, $frame, $left, $right, $operator, $node));
+        // can remove given the other logic below??
+        switch ($operator) {
+            case 'instanceof':
+                return $this->resolveInstanceOf($context, $resolver, $frame, $node->leftOperand, $right);
+        }
```
dantleech, line 48: try it

dantleech:
below `->statements` is unreachable because the type is `Expression OR CompoundStatementNode` or whatever:

image

we could:
- update the definition finder to iterate over the union
- implement type narrowing
- or both

dantleech:
Also:

image

Above it's using the first type in the union, `Type` which does not contain the method, the second type _does_ and is a subtype of the `Type`.

dantleech:
getting closer:

```
                if ($statement instanceof ExpressionStatement) {
                    if ($statement->expression instanceof ThrowExpression) {
                        return true;
                    }
                }

                if ($statement instanceof ThrowExpression) {
                    return true;
                }

                if ($statement instanceof CompoundStatementNode) {
                    foreach ($statement->statements as $statement) {

```
above `$statement->statements` inherits the type from ExpressionStatement as that branch doesn't close (`CompoundStatementNode|ThrowExpression`).

- If the assertion is _absolute_ then it should hold after the branch.
- If the assertion is not (e.g. `instanceof`) then the type should be restored after the branch.

Probably it's better to always restore.

lib/ClassMover/Adapter/WorseTolerant/WorseTolerantMemberFinder.php
```diff
@@ -377,7 +377,8 @@ private function attachClassInfoToReference(MemberReference $reference, ClassMem
             return $reference;
         }
 
-        $accepts = $type->accepts(TypeFactory::class((string) $query->class()));
+
+        $accepts = TypeFactory::reflectedClass($this->reflector, (string)$query->class())->accepts($type);
```
dantleech, line 381: index search: this was the wrong way round, the _queried_ class should accept the "found" class.

lib/WorseReferenceFinder/Tests/Unit/WorseReflectionTypeLocatorTest.php
```diff
@@ -104,6 +104,66 @@ public function bar()
         self::assertEquals(9, $location->offset()->toInt());
     }
 
+    public function testLocatesFirstUnion(): void
```
dantleech, line 107: this is a "temporary" solution, we can prompt the user to indicate which type to jump to with a union

lib/WorseReflection/Core/DefaultResolverFactory.php
```diff
@@ -112,4 +118,15 @@ public function createResolvers(): array
...
     }
+
+    private function createStubRegistry(): FunctionStubRegistry
+    {
+        return new FunctionStubRegistry([
+            'array_sum' => new ArraySumStub(),
+            'is_null' => new IsSomethingStub(TypeFactory::null()),
+            'is_float' => new IsSomethingStub(TypeFactory::float()),
+            'is_int' => new IsSomethingStub(TypeFactory::int()),
+            'is_string' => new IsSomethingStub(TypeFactory::string()),
```
dantleech, line 129: can make this an extension point in the future, although these could also be implemented with `@phpactor-type-assert` or something, aliasing to Psalm's asserts.

lib/WorseReflection/Core/Inference/Assignments.php
```diff
@@ -13,63 +13,62 @@
...
     /**
-     * @param array<array-key, array{int, Variable}> $variables
+     * @param array<int,Variable> $variables
      */
     final public function __construct(array $variables)
     {
-        foreach ($variables as [$offset, $variable]) {
-            $this->variables[] = [ $offset, $variable ];
-        }
+        $this->variables = array_map(function (Variable $v) {
```
dantleech, line 25: switched this from `[int, Variable]` to `array<int,Variable>` as variables really shouldn't be validly declared at the same offset.

lib/WorseReflection/Core/Inference/Frame.php
```diff
@@ -99,11 +115,59 @@ public function name(): string
...
+
+        $properties = [];
+        foreach ($this->properties() as $property) {
+            $properties[$property->name()] = $property;
+        }
+        foreach ($properties as $property) {
+            $this->properties()->add($after, $property);
+        }
```
dantleech, line 171: a nest of bugs

lib/WorseReflection/Core/Inference/FunctionStub/IsSomethingStub.php
```diff
@@ -0,0 +1,56 @@
...
+use Microsoft\PhpParser\Node\Expression\Variable;
+use Phpactor\WorseReflection\Core\Inference\Frame;
+use Phpactor\WorseReflection\Core\Inference\FunctionStub;
+use Phpactor\WorseReflection\Core\Inference\NodeContext;
+use Phpactor\WorseReflection\Core\Inference\NodeContextResolver;
+use Phpactor\WorseReflection\Core\Inference\TypeAssertion;
+use Phpactor\WorseReflection\Core\Type;
+use Phpactor\WorseReflection\Core\Type\BooleanLiteralType;
+
+class IsSomethingStub implements FunctionStub
```
dantleech, line 16: coudl rather be `TypeAssertion` stub

PR #1424, opened 2022-04-13
Fix renaming not current document

dantleech:
Replaces #1404


dantleech:
guess we need to tag and update class-to-file too

przepompownia:
It would be good to use the version without the mentioned error. 

Thanks for resolving the whole issue with class renaming. It revealed some side issues not only in phpactor dependencies but also on the neovim side and its extensions. Some of them are still waiting to fix.

Issue #1460, opened 2022-04-22
Request textDocument/codeAction failed


victorzhuk:
```
Message: Exception [TypeError] Phpactor\WorseReflection\Core\Inference\Resolver\BinaryExpressionResolver::applyTypeAssertions(): Argument #5 ($rightOperand) must be of type Microsoft\PhpParser\Node, Microsoft\PhpParser\MissingToken given
```

and the same with `negate(...)`

Asserted when left or right operand yet not exists, like 
```php
if ($e->getResponse()->getStatusCode() !== ) { ... }
```
or

```php
if ( !== $e->getResponse()->getStatusCode()) { ... }
```

Solution founded, PR in progress

dantleech:
should be fine now

PR #1461, opened 2022-04-22
Fix 1460 issue: request textDocument/codeAction failed

victorzhuk:
#1460


dantleech (approved):

dantleech:
you need to fix PHPStan, you can run `composer integrate` as a pre-push check

dantleech:
Looks like this was fixed in the meantime: https://github.com/phpactor/phpactor/blob/master/lib/WorseReflection/Core/Inference/Resolver/BinaryExpressionResolver.php#L62

PR #1475, opened 2022-04-24
Closure docblock

### LINKED ISSUE #1474: Union docblock
The docblock below is generated:

```
    /**
     * @return MissingType|Type|UnionType
     */
    public static function fromTypes(Type ...$types): Type
```

- If all the union members are instances of the type then it should not generate the docblock

dantleech:
Fixes #1474 and:

- Supports adding closure docblocks
- Fixes union merging (union + union = union, not union(union))
- ...


lib/WorseReflection/Core/Type/UnionType.php
```diff
@@ -18,7 +18,21 @@ final class UnionType extends Type
     public function __construct(Type ...$types)
     {
         $unique = [];
+        $toMerge = [];
+        $hasNull = false;
```
dantleech, line 22: dead codes

PR #1479, opened 2022-04-25
Improve json schema build

przepompownia:
- introduce enums
- change schema URL to working with vscode json ls
- use types
- generate default values
- add example usage

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
         "composer/package-versions-deprecated": "^1.0",
         "composer/xdebug-handler": "^2.0",
         "symfony/yaml": "^5.1",
-        "phpactor/container": "^2.0.0",
+        "phpactor/container": "dev-allow-enums",
         "phpactor/class-to-file": "~0.4.3",
         "twig/twig": "^2.4",
         "webmozart/path-util": "^2.3",
@@ -78,9 +78,25 @@
         "phpactor/worse-reference-finder": "0.2.6",
         "phpactor/worse-reflection-extension": "0.2.5"
     },
+    "repositories": [
+        {
+            "type": "vcs",
+            "no-api": true,
+            "url": "https://github.com/przepompownia/container"
+    },
+        {
+            "type": "vcs",
+            "no-api": true,
+            "url": "https://github.com/przepompownia/map-resolver"
+        }
+    ],
     "config": {
         "platform": {
             "php": "7.4.0"
+        },
+        "allow-plugins": {
+            "dantleech/what-changed": true,
+            "phpstan/extension-installer": true
         }
     },
     "autoload": {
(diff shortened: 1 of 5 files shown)


przepompownia:
Why not to generate defaults?

przepompownia:
When I opened autogenerated JSON schema to manual edition, completion did not work with the original URL so I changed it to some working.

composer.json
```diff
@@ -78,6 +78,18 @@
         "phpactor/worse-reference-finder": "0.2.6",
         "phpactor/worse-reflection-extension": "0.2.5"
     },
+    "repositories": [
+        {
+            "type": "vcs",
+            "no-api": true,
+            "url": "https://github.com/przepompownia/container"
```
dantleech, line 85: is there a PR for container also?

lib/Extension/Core/Model/JsonSchemaBuilder.php
```diff
@@ -28,7 +28,7 @@ public function __construct(string $title, array $extensions)
     public function dump(): string
     {
         $schema = [
-          '$schema' => 'https =>//json-schema.org/draft/2020-12/schema',
+          '$schema' => 'https://json-schema.org/draft-07/schema',
```
dantleech, line 31: i think the original is valid without the `=>` typo: https://json-schema.org/learn/getting-started-step-by-step

dantleech (approved):
nice. the enum is definitely something I wanted to add, i use a similar approach in PHPBench using the options Symfony resolver:

```
                    $values = $inspector->getAllowedValues($option);
                    $meta['enum'] = $values;
```

dantleech:
the map resolver PR is failing

composer.json
przepompownia: https://github.com/phpactor/container/pull/12

dantleech: cool closed that, we can now just require `map-resolver: ^1.5.0`

lib/Extension/Core/Model/JsonSchemaBuilder.php
przepompownia: Probably it's valid but it does not work for me for some reason, in opposite to `draft-07`. Maybe it's a problem of vscode-json-language-server. It reports that this draft it not fully supported yet. 

It's not important as we do not edit the schema file manually in normal case. If you prefer `2020-12` then no problem for me to change it back.

dantleech: no it's fine i guess

przepompownia:
https://github.com/phpactor/phpactor/runs/6173691607?check_suite_focus=true Could you look over there? I have no permission to run the build again.

dantleech:
i changed the settings for phpactor so that the build runs for non-new GH users

przepompownia:
Thanks!

PR #1511, opened 2022-05-01
feat: return package name from composer.json as name in capabilities-…

lalanikarim:
…>serverInfo instead of empty array.


dantleech:
thanks!

lib/Extension/LanguageServer/LanguageServerExtension.php
```diff
@@ -311,9 +311,11 @@ private function registerMiddleware(ContainerBuilder $container): void
                 $stack[] = new ErrorHandlingMiddleware($this->logger($container));
             }
 
+            $packageName = \Composer\InstalledVersions::getRootPackage()['name'];
```
dantleech, line 314: is there any reaso nto use composer for this?

Issue #1524, opened 2022-05-02
LSP: capabilities->documentHighlightProvider and capabilities->documentSymbolProvider non conformant value


lalanikarim:
The capabilities response from phpactor includes non-conformant data for "documentSymbolProvider" and "documentHighlightProvider"
See attached file:
[jsonformatter.txt](https://github.com/phpactor/phpactor/files/8605669/jsonformatter.txt)

Currently, they are being returned as empty arrays `[]` which trips certain LSP clients like [helix-editor](https://helix-editor.com/)

The issue seems to be stemming from json serialization of `DocumentSymbolOptions` and `DocumentHighlightOptions` in `DocumentSymbolProviderHandler` and `HighlightHandler` respectively. Both takes an optional argument of `$workDoneProgress` which defaults to null.

One potential fix is to supply default options for these. 
What are the correct values for $workDoneProgress for these? 

As a workaround, I am able to proceed by providing the value of `true`. However, I am not sure if that's the correct value.

dantleech:
yes, i think the solution would be to set to `true` for now

lalanikarim:
Thanks @dantleech 
Sending a PR. Hopefully, it meets the necessary requirements.

lalanikarim:
Issue resolved. Thanks.

PR #1527, opened 2022-05-02
Since workDoneProgress isn't supported for documentSymbolProvider and…

lalanikarim:
… documentHightlightProvider, removing the options and return true instead to conform it to spec

https://github.com/phpactor/phpactor/issues/1524


dantleech:
:crossed_fingers: thanks!

PR #1532, opened 2022-05-03
Fix regression

dantleech:
This change caused refactorings such as add missing properties and
complete constructor to stop working

cc @przepompownia


przepompownia:
All right, it was good to try.

PR #1558, opened 2022-05-08
Improve Completion #1

dantleech:
This PR aims to improve the context sensitivity of completion:

- Expression completion (functions + classes)
- Type completion (e.g. retun types, parameter declaration types) this includes scalar types.
- Clause completion (`implements` => interfaces, `use ` => traits, `extends` => classes
- Use statement completion => classes

Note that whereas previously we "allowed" completion by default, we now restrict more by default, so its quite likely that completion may stop working in some contexts errornously.


przepompownia:
After this change quick manual tests suggest that I can complete only keywords.

przepompownia:
Maybe false alarm - not know why initially did not work (with index built from scratch).

dantleech:
there is one migration issue which i thought i solved:

the search index now includes the class _type_ so on an old index it would not find anything if searching for `insterface` or `class`. however i put in some B/C so that it should "just" continue to work (completing `implements <>` will show classes on an old index).

but possible there are some gremlins.

przepompownia:
It seems that completion does not work directly after building the index. After starting the server again with the same index created before, completion works.

przepompownia:
Reproductible for me: https://asciinema.org/a/dne4pV2RAImoc2EaplPdvmvXI on this example I cannot complete `$this` until restart phpactor.

dantleech:
Looks like NULL was being passed https://github.com/phpactor/phpactor/pull/1561/files

przepompownia:
Now completion works just after creating index :+1:

Issue #1557, opened 2022-05-08
Virtual function are not autocompleted and are showing warnings


KasparRosin:
Hey, 

I read throw some of the issues and to my understanding, this should already be implemented in [#51](https://github.com/phpactor/worse-reflection/pull/51)?
Currently, the `@property` magic variables are auto-completed, but `@method` is not and are showing warnings `Method "name" does not exist` for me.
Working with the same [Laravel Faker](https://github.com/fzaninotto/Faker/blob/master/src/Faker/Generator.php) class mentioned in #682

Currently working with latest master commit version: `af0d2ef `.
IDE is nvim, with nvim-lspconfig and nvim-lsp-installer.

Let me know if i can provide any additional information. 

Edit:
The warnings only appear after executing write command `:w`

dantleech:
the diagnostics are executed on save only atm as they are (still) quite expensive, you can enable real time diagnostics with : https://phpactor.readthedocs.io/en/master/reference/configuration.html#language-server-diagnostics-on-update

it works for me:

![image](https://user-images.githubusercontent.com/530801/167291166-106b9c33-cded-41da-a55d-326683c4e424.png)

with the fake annotations:

![image](https://user-images.githubusercontent.com/530801/167291204-2bf830c2-b7ef-4b54-9b68-ac3df92aeaaf.png)

KasparRosin:
Ok, yes.
The @method seems to work if it's defined in the current or any of the classes being extended.
For me, the problem currently came when i tried to read a method from an instance saved to a variable.

```php
class CompanyFactory extends Factory {
// ...
    return [ 'company' => $this->faker->company() ]
}
```

```php
abstract class Factory {
 /**                                                                                                                                                                                                      
  * The current Faker instance.                                                                                                                                                                           
  *                                                                                                                                                                                                       
  * @var \Faker\Generator                                                                                                                                                                                 
  */                                                                                                                                                                                                      
  protected $faker;  
 // ...
}
```

And now inside the Faker\Generator class there is the `@method company`

KasparRosin:
Screenshots:
![image](https://user-images.githubusercontent.com/33309407/167344276-f3270a68-94a4-43ed-a1c5-2a8e94ce6415.png)
![image](https://user-images.githubusercontent.com/33309407/167344307-0fec1660-5da0-4b45-af8c-2a89410d9ca6.png)

dantleech:
[...]

PR #1568, opened 2022-05-10
Missing virtual method completion

KasparRosin:
Virtual `@method` breaking test case.
#1557

### DIFF
--- a/lib/WorseReflection/Tests/Inference/virtual_member/method.test
+++ b/lib/WorseReflection/Tests/Inference/virtual_member/method.test
@@ -10,6 +10,10 @@ class ReflectionMethod {}
 
 class MemberCollection
 {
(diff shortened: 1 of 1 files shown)


lib/WorseReflection/Tests/Inference/virtual_member/method.test
```diff
@@ -10,6 +10,10 @@ class ReflectionMethod {}
 
 class MemberCollection
 {
+    /**
+    * @method string set(string $name)
+    */
```
dantleech, line 15: this is defined in the wrong place right?

KasparRosin: It was yes, my bad. Moved it outside of class now.

PR #1582, opened 2022-05-14
Revert "Support indexing new object expressions as references to __construct"

dantleech:
Reverts phpactor/phpactor#1581


dantleech:
This needs to be done as a separate indexer and with special handling rather

PR #1604, opened 2022-05-16
Self destruct

### LINKED ISSUE #1599: LSP Process can live on after quitting neovim
Not sure if this is an issue of phpactor or https://github.com/neovim/nvim-lspconfig/. I'm only using phpactor as LSP right now.

When I close a neovim instance which created an LSP Process, this process will still live. This can stack up to an huge number of processes after a week of work. I tend to close neovim instances and start from new when switching a concrete task within a project.

Would be cool if the LSP process is being killed as soon as the neovim process is killed.

Let me know if you need further info. Using neovim
```
NVIM v0.7.0
Build type: Release
LuaJIT 2.1.0-beta3
```

and phpactor 45207d99c7db718d0d5467a0cbd1a98587ee7551 with php `7.4.29`

dantleech:
Will self destruct if no exit notification recieved after 2.5 seconds fixes #1599

### DIFF
--- a/doc/reference/configuration.rst
+++ b/doc/reference/configuration.rst
@@ -1328,6 +1328,36 @@ Log incoming and outgoing messages (needs log formatter to be set to ``json``)
 **Default**: ``false``
 
 
+.. _param_language_server.shutdown_grace_period:
+
+
+``language_server.shutdown_grace_period``
+"""""""""""""""""""""""""""""""""""""""""
+
+
+
+
+Amount of time to wait before responding to a shutdown notification
+
+
+**Default**: ``200``
+
+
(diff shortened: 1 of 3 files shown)


doc/reference/configuration.rst
```diff
@@ -1328,6 +1328,36 @@ Log incoming and outgoing messages (needs log formatter to be set to ``json``)
...
+.. _param_language_server.shutdown_grace_period:
+
+
+``language_server.shutdown_grace_period``
+"""""""""""""""""""""""""""""""""""""""""
+
+
+
+
+Amount of time to wait before responding to a shutdown notification
```
przepompownia, line 1340: It would be good to add time units to both descriptions.

PR #1615, opened 2022-05-17
Do not add $this for anonymous function

dantleech:


przepompownia:
It works on my test example :+1:

PR #1617, opened 2022-05-18
Imrove docblock performance by caching

dantleech:
- Adds a benchmark using the Yii ActiveRecord stack
- Add a cached docblock parser factory. It uses the standard WR cache (default to 5 second TTL),

```

\Phpactor\WorseReflection\Tests\Benchmarks\PhpUnitReflectClassBench

    test_case...............................I4 - [Mo0.9389μs vs. Mo0.9000μs] +4.33% (±8.13%)
    test_case_methods_and_properties........I4 ✔ [Mo10.96ms vs. Mo14.21ms] -22.83% (±10.99%)
    test_case_method_frames.................I4 ✔ [Mo0.28ms vs. Mo0.26ms] +8.16% (±5.81%)

\Phpactor\WorseReflection\Tests\Benchmarks\YiiBench

    benchMembers............................I0 - [Mo103.677ms vs. Mo193.790ms] -46.50% (±0.00%)
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@
 /**/Tests/Workspace
 /lib/Completion/cache
 /lib/Test.php
+/lib/WorseReflection/Tests/Cache
 /lib/Extension/WorseReflection/stubs
 
 # File frequently used for testing
--- a/lib/WorseReflection/Bridge/Phpactor/DocblockParser/CachedParserFactory.php
+++ b/lib/WorseReflection/Bridge/Phpactor/DocblockParser/CachedParserFactory.php
@@ -1,0 +1,28 @@
+<?php
+
+namespace Phpactor\WorseReflection\Bridge\Phpactor\DocblockParser;
+
+use Phpactor\WorseReflection\Core\Cache;
+use Phpactor\WorseReflection\Core\DocBlock\DocBlock;
+use Phpactor\WorseReflection\Core\DocBlock\DocBlockFactory;
+use Phpactor\WorseReflection\Core\TypeResolver;
+
+class CachedParserFactory implements DocBlockFactory
+{
+    private Cache $cache;
+
+    private DocBlockFactory $innerFactory;
+
+    public function __construct(DocBlockFactory $innerFactory, Cache $cache)
+    {
+        $this->cache = $cache;
+        $this->innerFactory = $innerFactory;
+    }
+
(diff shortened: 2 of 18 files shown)


dantleech:
cc @przepompownia doesn't solve the performance issue but it's almost twice as fast.

przepompownia:
https://blackfire.io/profiles/dc5aed9f-6944-44bf-b166-dee5207d59bd/graph 

@dantleech I agree that it's still far from resolving the problem but (on the example project) 3s instead of 13s is a significant change. Thank you!

dantleech:
the `ActiveRecord` has around 150 members, most of the time is taken in resolving the types for those members, which isn't strictly necessary for completion, Another option might be to either:

1) Have an option to disable "inferred" types (i.e. parsing docblocks) for completion (`[Mo868.534μs vs. Mo193.790ms]`)
2) Perhaps be smart about degrading the completion results, e.g. if time > X milliseconds stop inferring the types.

przepompownia:
Let me add another:
3. I did not check how `isIncomplete` is (and how it could be) used in practice. Sometimes getting `$this` from `$th` can take above five seconds (while such local result could be yielded before finding something heavier). 
4. what dou you think about longer cache TTL only for results from external libraries (to customize especially along with custom autoloader path)?

dantleech:
3) => this determines if the client will cache the results (were the results complete, or based on a partial search).
4) => maybe possible to have a cache based on file modification time :thinking:

przepompownia:
3. at the moment I did not checked how caching works on the built-in nvim LSP client and possibly nvim-cmp - but without checking it seems that clien-side logic should respect `isIncomplete` flag. 
4. repeating the same heavy search in a vendor space that left unchanged at regular work seems to be worth caching (and worth remembering to invalidate it on composer commands or similar actions). Independently, mtime-based caching also seems to be a good idea.

przepompownia:
>     1. Have an option to disable "inferred" types (i.e. parsing docblocks) for completion (`[Mo868.534μs vs. Mo193.790ms]`)> 
>     2. Perhaps be smart about degrading the completion results, e.g. if time > X milliseconds stop inferring the types.

 1. Disabling inferrence from docblocks can be good idea but not necessary for projects based on obsolete frameworks as Yii with lots of dependencies of Active Records at the some layers. 
 2. may be annoying if finding the expected result needs more time than X.

Issue #1603, opened 2022-05-16
No Trait Virtual Method Reflection


ging-dev:
phpactor can suggest virtual methods, but when i click save, the method doesn't exist warning.

![image](https://user-images.githubusercontent.com/42226341/168612507-f5540069-41f7-46bb-b2af-d99bdfd732be.png)

dantleech:
yes that diagnostic is notorius.

dantleech:
Looks like neither Traits or Enums support "virtual" methods or properties, bit confused how it works so far:

https://github.com/phpactor/phpactor/blob/042c2f2733189118f8fe95a03b9701e43c81c98f/lib/WorseReflection/Bridge/TolerantParser/Reflection/Collection/ReflectionClassCollection.php#L45-L53

dantleech:
should be fixed in : https://github.com/phpactor/phpactor/pull/1619

PR #1619, opened 2022-05-19
bugfix: Fix false positive for virutal method not existing #1603

dantleech:


lib/WorseReflection/Core/Inference/Resolver/MemberAccessExpressionResolver.php
```diff
@@ -104,7 +103,7 @@ public static function infoFromMemberAccess(NodeContextResolver $resolver, Frame
                     );
                 }
 
-                $types[] = $declaringClass;
+                $types[] = $subType;
```
dantleech, line 106: this change isa bit risky but I think it's the correct behavior. previously the "container type" would be resolved to the declaring type (so for example the "trait" that declared it). this caused the bug because the Trait reflection does not include virtual methods (they are grafted onto the instance class).

what this means is that if you reflect the offset for the member call, the "container type" will be the parent class, trait or interface, which might not be what you'd expect.

PR #1620, opened 2022-05-20
Adding a command for cleaning up indices

### LINKED ISSUE #1374: New command: IndexCleanup
Hi, thanks for your hard work in phpactor, it is a very useful tool and I love it!

I just found out that my phpactor cache folder was using around ~5GB of my local storage and tried to find 
an "official" way to clean it, but ended up just deleting the sub directories myself.

Is it suitable to add an "IndexCleanupCommand" to phpactor? If so, I'm happy to provide a PR implementing it.

Would also appreciate any guidance on what you'd like to see in such command (eg. parameters etc).

mamazu:
Closes #1374 

This is my shot at a cleanup command. It has an interactive version (as described in the ticket) and a command line argument. So that it can be used in scripts.

As a newbie to the phpactor development, please let me know if I have made some errors.

## Example of the UI
image

## Known issues
One thing I am not sure about and I need input on is: what is the best way to find the size of a directory. If I just use the `SplFileInfo->getSize()` function, it just returns the size of the directory entry and not the total size of the content.


dantleech:
yes, getting the size of the directory in a cross-platform way is tricky. i think the best we can do is [recursively iterate over all the files](https://stackoverflow.com/questions/478121/how-to-get-directory-size-in-php) and sum them (which is also inaccurate).

one thing I might be tempted to do here (but doesn't necessarily have to be this PR) is move the responsiblity to the `Phpactor\Indexer\Model\Index` interface. Currently we use a file index, but I'm interested in moving moving to/providing an option for sqlite.

The implementation of this class is `SerializedIndex` which uses a `FileRepository` which could have the logic for getting the size.

So then the challenge would be how to get a list of `Index` objects.

Just something to consider, it would be a fine first-step to have this command in it's current state.

dantleech:
Also you can use `composer integrate` to run the unit tests, cs and phpstan checks.

mamazu:
I can do the refactoring as well. How would I get an index from a path though? Is is through the `IndexAgentBuilder`?

mamazu:
And about the file sizes: Yes, there is a difference with du. I have run it on a few repositories and it looks like what you suspected is true. `du` returns a wildly different result than php.

image

lib/Indexer/Model/Index/IndexInfo.php
```diff
@@ -0,0 +1,58 @@
...
+        $this->absolutePath = $absolutePath;
+    }
+
+    public static function create(SplFileInfo $fileInfo): self
+    {
+        return new self(
+            $fileInfo->getRealPath(),
+            $fileInfo->getRelativePathname(),
+            $fileInfo->getSize(),
+            sprintf('%.1f days', ((time() - $fileInfo->getMTime())/(3600*24)))
```
dantleech, line 35: lets keep the formatting lazy, better to have the raw data as class properties (i.e. `getMtime()` only)

lib/Indexer/Model/Index/IndexInfo.php
```diff
@@ -0,0 +1,58 @@
...
+        $this->lastUpdated = $lastUpdated;
+        $this->absolutePath = $absolutePath;
+    }
+
+    public static function create(SplFileInfo $fileInfo): self
+    {
+        return new self(
+            $fileInfo->getRealPath(),
+            $fileInfo->getRelativePathname(),
+            $fileInfo->getSize(),
```
dantleech, line 34: i wouldn't have expected this to return the directory size? i would expect to have to traverse all the files in the index an sum the sizes?

mamazu:
I am now using the directory size. But it looks like the size in dev is not the same as in the pipeline.

Can you check out the branch and see if the tests fail on your machine as well?

dantleech:
Tests fail for me too:
```
-Total size: 78.06 K\n
+Total size: 79.32 K\n
```
it might be platform/os dependent, so maybe with stubbing out somehow? or i guess we can just test the `FileInfo` and not check the specific size.

functionally it looks pretty good, although `du` is still off (`3,9G` by DU, `4.27G` by Phpactor). So maybe worth adding some expanation there:


```
+----+---------------------------------+----------+---------------+
| #  | Directory                       | Size     | Last modified |
+----+---------------------------------+----------+---------------+
| 1  | arch-cop-5838a9                 | 67.62 M  | 34.7 days     |
| 2  | code-kata-berlin-f8f6c1         | 70.37 M  | 16.8 days     |
| 3  | deploy-b9cc03                   | 139.24 M | 25.8 days     |
| 4  | example-legacy-project-094d24   | 129.17 M | 0.4 days      |
| 5  | language-server-bbd062          | 133.65 M | 22.0 days     |
| 6  | language-server-protocol-e5762e | 109.04 M | 4.5 days      |
| 7  | php-956212                      | 69.27 M  | 16.8 days     |
| 8  | phpactor-c6a5f2                 | 278.05 M | 2.4 days      |
| 9  | phpactor2-35a1b1                | 545.20 M | 1.0 days      |
| 10 | phpactor2-9cff54                | 128.59 M | 34.7 days     |
| 11 | phpbench-f746c2                 | 112.12 M | 26.1 days     |
| 12 | phpstan-src-bae420              | 332.11 M | 35.0 days     |
| 13 | phpstorm-stubs-079d17           | 105.32 M | 42.0 days     |
| 14 | xxxxx-e5a2f7                    | 688.90 M | 19.9 days     |
| 15 | xxxxxxxxx-drupal-b78a49       | 723.97 M | 1.7 days      |
| 16 | xxxxxxxxx-sylius-515090       | 461.25 M | 1.9 days      |
| 17 | workspace-4c7620                | 158.93 M | 22.8 days     |
| 18 | yii2-basic-92244b               | 115.22 M | 3.5 days      |
+----+---------------------------------+----------+---------------+
Total size: 4.27 G
```

It also took a significant amount of time.

You could use a Symony Console `Progress` maybe to report progress (1/6 indexes calculated) f.e.

mamazu:
I have added the progress bar to it. But I am not sure that the directory size is essential. Maybe it would be enough to just count the files and really do a rough estimate on how big it is going to be.

Otherwise we could also make the file size also an optional thing, I am not sure that you need it if you just want to clear the cache of a project (no matter how big it is).

mamazu:
I have fixed the tests with the file size. And I'll fix the input next trying to centeralize the handling of the input.

dantleech:
looking good, I think we should display the filesize:

- the filesize should be evaluated lazily (otherwise the progress bar is not working as expected)
- i get a fatal error if just pressing "enter" without selecting an index

mamazu:
I implemented your suggestions. The test are now no longer checking the output of the program because constructing a regex with a table and all the `|`s is very complicated and unreadable.

What do you think?

dantleech:
yes, "smoke test" is fine here I tihnk.

mamazu:
I mean there are tests asserting that the directories are removed. Everything else is just fluff around it. :D

lib/Indexer/Extension/Command/IndexCleanCommand.php
```diff
@@ -0,0 +1,181 @@
...
+                bin/console index:clean --no-interaction
+
+            Removing an index by name
+                bin/console index:clean <index name>
+
+            Removing an index by the number in the list view:
+                bin/console index:clean <index number>
+
+            Removing all indicies
+                bin/console index:clean %s
```
dantleech, line 49: %s?

dantleech (approved):

dantleech:
thanks! i'll make a few tweaks.

lib/Indexer/Extension/Command/IndexCleanCommand.php
mamazu: Yeah this is an `sprintf` Argument, and it will be filled with the `CLEAN_ALL` option.

PR #1630, opened 2022-05-20
Do not complete constants on instance

### LINKED ISSUE #1614: Constants are added to the result of completion triggered by arrow
From [StupidController.php](https://github.com/przepompownia/example-legacy-project/blob/master/yii2-basic/modules/example/commands/StupidController.php). It's a side effect of looking at the example for #1125.

image

I haven't tested how much this affects the response time.

dantleech:
Fixes #1614


lib/Completion/Bridge/TolerantParser/CompletionContext.php
```diff
@@ -142,9 +146,32 @@ public static function classMembersBody(?Node $node): bool
...
+                }
+            }
+            if ($classLike instanceof TraitDeclaration) {
+                if (!$classLike->traitMembers->closeBrace instanceof MissingToken) {
+                    return false;
+                }
+            }
+            if ($classLike instanceof EnumDeclaration) {
+                if (!$classLike->enumMembers->closeBrace instanceof MissingToken) {
+                    return false;
```
dantleech, line 171: geez

PR #1633, opened 2022-05-21
Add PhpSpec extension

camilledejoye:
Hi,

I tried to re-introduce the PhpSpec extension.
I had a hard time figuring out how to configure the integration test to load files from the disk, hence the `createSource` method. Don't hesitate if there is a cleaner way to do.

Also I wanted to ask about the `TODO`, should this case be handled with a `Walker` or is it not the correct way to tackle it ?


lib/Extension/PhpSpec/Provider/ObjectBehaviorMemberProvider.php
```diff
@@ -0,0 +1,89 @@
...
+                $subjectClassName,
+                $e->getMessage()
+            ));
+            return VirtualReflectionMethodCollection::fromMembers([]);
+        }
+
+        $virtualMethods = [];
+        foreach ($subjectClass->methods() as $subjectMethod) {
+            $method = VirtualReflectionMethod::fromReflectionMethod($subjectMethod);
+            $subjectType = new ClassType(ClassName::fromString(self::SUBJECT_CLASS));
```
dantleech, line 58: could maybe add an intersection type here: `TypeFactory::intersection(TypeFactory::reflectedClassType($reflector, self::SUBJECT_CLASS), TypeFactory::reflectedClassType($reflector, ThePhpspecClass::class))`

(the types can be smarter if they have a Reflector instance in them)

dantleech:
nice! i forgot all about this. 

r.e. the "walker" worse reflection got refactored alot since this extension was created.

with the new type system I wonder if we could do something with Intersection types. e.g. `$this => MyFoobar&MyFoobarSpec` but maybe the MemberProvider is still useful...

We could augment the types, so `company(): Company` would become `company(): Company&ThePhpSpecObjectBehaviorThing`

dantleech:
> hence the createSource method. Don't hesitate if there is a cleaner way to do.

could also have a directory and `ReflectorBuilder::create()->withLocator(new BruteForceLocator(__DIR__ . '/stubs'))`

camilledejoye:
> with the new type system I wonder if we could do something with Intersection types. e.g. $this => MyFoobar&MyFoobarSpec but maybe the MemberProvider is still useful...

It should be a union instead of an intersection otherwise we would only have access to common methods which will often end up with no result at all.
That also made me realize that the member provider should only return public methods and if we just change the type of `$this` with a walker then we won't be able to have this distinction right ?

> We could augment the types, so company(): Company would become company(): Company&ThePhpSpecObjectBehaviorThing (Collaborator ?)

That's for what I want to use the frame walker, to change the return type of the necessary call expression to add a union with `Subject`.
`Collaborator` is their wrapper around prophecies.

> Maybe worth checking out how PHPStan or Psalm PHPSpec extnesions do it, if they exist (https://github.com/proget-hq/phpstan-phpspec wow, it's involved, maybe this is simpler :D )

Simpler ?

> could also have a directory and ReflectorBuilder::create()->withLocator(new BruteForceLocator(__DIR__ . '/stubs'))->build() to return the Reflector. You can then avoid booting the extension and have a more focused test.

I'll gave it a try, thanks.

dantleech:
> It should be a union instead of an intersection

intersection means _implementing_ both contracts. so the callable methods are all the methods on both types. the difference is that union is it can be _this_ **or** _that_. so you can't safely call methods on them until you narrow the type. with an intersection you are saying you **can** safely call the methods from all types. 

> That's for what I want to use the frame walker,

as commented, I think we can avoid the "walker" by just adding an intersection on the provided member?

camilledejoye:
> > It should be a union instead of an intersection
> 
> intersection means _implementing_ both contracts. so the callable methods are all the methods on both types. the difference is that union is it can be _this_ **or** _that_. so you can't safely call methods on them until you narrow the type. with an intersection you are saying you **can** safely call the methods from all types.

OK, I thought intersection meant only what is common to the two types. I guess I was a bit too literal in my interpretation :)

> > That's for what I want to use the frame walker,
> 
> as commented, I think we can avoid the "walker" by just adding an intersection on the provided member?

I'll try with the intersection but I'm not sure it will provide the expected result, it currently does not work with the union.  
Since the difference between union and intersection is only about "safety" I expect the same behavior.
From the example in the test when we are in the `EmployeeSpec` class:
* `$this` will provide methods from both `Employee` and `ObjectBehavior`
* `$this->company()` will return a type `Company&Subject` on which we can call either `Company::name(): string` or `Subject::shouldBe(): void`
* `$this->company()->name()` will have the type `string` because it's the return type of `Company::name(): string` when it should be `string&Subject`

This start to get weird to have intersections between literal types and objects...
This is because it should actually be a `Subject<string>`.

A few weeks ago I was actually wondering if it would be possible to define some kind of stubs with generics for the classes from PhpSpec ?
Because the use of intersection is actually a trick to obtain the expected behavior for a generic which is never defined.

dantleech:
yes, creating a generic `Subject<T>` stub sounds good, although currently there is no way for extensions to add a stub (it could be done via. configuration in the short term).

You could then do that instead of `type&intersection`?

camilledejoye:
> I'll try with the intersection but I'm not sure it will provide the expected result, it currently does not work with the union.

I confirm that intersection does not provide the "chained" behavior.

I also encounter a case which I think is invalid but since I just learned the true meaning of intersection I might be wrong :stuck_out_tongue: 
If I add the following test to `AgregateTypeTest.php`:
```php
    public function testHandleIntersectionWithNullableType(): void
    {
        self::assertEquals('(null|One)&Two', (string) TypeFactory::intersection(
            TypeFactory::nullable(TypeFactory::class('One')),
            TypeFactory::class('Two'),
        ));
    }
```
It fails:
```
Aggregate Type (Phpactor\WorseReflection\Tests\Unit\Core\Type\AggregateType)
 ✘ Handle intersection with nullable type
   ┐
   ├ Failed asserting that two strings are equal.
   ┊ ---·Expected
   ┊ +++·Actual
   ┊ @@ @@
   ┊ -'(null|One)&Two'
   ┊ +'null&One&Two'
   │
   ╵ /home/cdejoye/.local/share/nvim/site/pack/packer/start/phpactor/lib/WorseReflection/Tests/Unit/Core/Type/AggregateTypeTest.php:176
   ┴
```

I have a possible fix that I can propose as a PR if you validate the test assertion.

> ~yes, creating a generic `Subject<T>` stub sounds good, although currently there is no way for extensions to add a stub (it could be done via. configuration in the short term)~
> 
> i think the problem is how to inherit the methods, you can't do that with Generics.

Indeed it was not to add the methods, we will still need the same behavior.
But it would be more "accurate" because PhpSpec objects are decorators by nature.
It might also make things simpler as for any type `Subject<T>` we could add all the methods for the object(s) (think to handle aggregate types) with an inferred return type `Subject<U>`.

Because without the walker I don't see how to do the "chain" behavior otherwise.

dantleech:
> I also encounter a case which I think is invalid but since I just learned the true meaning of intersection I might be wrong ￼
If I add the following test to AgregateTypeTest.php

it's invalid because `null&Foo` is not possible, so migt not be worth worrying about. but maybe we could make it act in a more expected way (see the `Intersection` constructor)

dantleech:
we can add a "walker": https://github.com/phpactor/phpactor/blob/3f1b550656e1f095913040db16d7a0d23e813fc0/lib/Extension/WorseReflection/WorseReflectionExtension.php . although i'm not sure how it might help, as it is for modifying (e.g. adding/removing vars) from the `Frame`.

The other, currently not extensible, point is the `Resolver` which resolve the types of nodes (e.g.  https://github.com/phpactor/phpactor/blob/3f1b550656e1f095913040db16d7a0d23e813fc0/lib/WorseReflection/Core/Inference/Resolver/TernaryExpressionResolver.php#L13). The limitation of these is that they have a direct [1-1 mapping](https://github.com/phpactor/phpactor/blob/3f1b550656e1f095913040db16d7a0d23e813fc0/lib/WorseReflection/Core/DefaultResolverFactory.php#L117-L122) with the node class so can't register f.e. multiple resolvers for a member-access- node. 

(side note: i have been trying to remove the `Walker` concept in favor of having only the `Resolver` but there is some work to do there yet).

not much help, but if you can figure out a logical way to do this using either of the above concepts then we could figure it out.

dantleech:
fwiw the phpstan-prophecy extension might be worth a look: https://github.com/Jan0707/phpstan-prophecy/tree/master/src

DanielSiepmann:
I thought about the phpstan extensions as well. We set them up on each project, depending on the concrete project. E.g. I use TYPO3 (open source CMS) a lot of time. There is a dedicated phpstan extension for that as well.

Would it be possible to fetch them from the current project? E.g. if the composer support is enabled? Not sure about other issues. I'm currently using ale which integrates phpstan. phpactor is integrated via LSP and reports less issues.

I can also open a new issue for that, if it makes sense. Don't want to high jack this PR for something else.

dantleech:
@DanielSiepmann if you're referring to "extensions" I've disabled that possiblity for now in favour of bundling them all into the monolith. It makes maintenance easier and gives us much more flexibility (no need to worry about B/C) in addition to ensuring that they always work. Maybe once the API settles down it can be made "extensible" again by third parties. But yeah, let's raise further issues or start a Discussion.

dantleech:
nope, this is about adding a new feature

dantleech:
sorry yes, the fact that prophecy does not inherit methods from the thing it prophesises is kindof related to this issue, although it would idelaly be a separate extension.

dantleech:
closing this PR for now as stale

Issue #1634, opened 2022-05-22
Generate method provider bottleneck


dantleech:
The generate method provider takes almost 20 seconds in the Phpactor `ReflectionClassTest`
```
Diagnostic finsihed in "18.98" (Phpactor\Extension\LanguageServerCodeTransform\CodeAction\GenerateMethodProvider)
```

dantleech:
this really needs tobe solved by running the node context resolver "once" and providing diagnostics in a single pass and ideally in a separate worker process.

dantleech:
Fixed in #1673

PR #1659, opened 2022-05-26
Diagonstics

dantleech:
This PR introduces a new feature for gathering diagnostics in a single pass and introduces some performance fixes.

The aim is to:

- facilitate the improving the perforamnce of f.e. the "missing/generate methods" code action (and other code actions) and diagnostics fixing #1634 
- make it easy to add new diagnostics (provisionally looks to be 0.4s vs. 30+s)

It also introduces a command to run the diagnostics on a given path:

image

Currnetly onolly the "missing methods" diagnostic is supported but it should be easy to add an arbitrary number.

Note that the perforamance of the command is currently pretty bad and is meant as a development aid not for practice static analysis.


lib/WorseReflection/Core/Cache/TtlCache.php
```diff
@@ -32,12 +32,21 @@ public function __construct(float $lifetime = 5.0)
...
+        }
 
         if (isset($this->cache[$key]) && $this->expires[$key] > $now) {
             return $this->cache[$key];
         }
 
-        $this->purgeExpired($now);
+        $elapsed = $now - $this->lifetimeStart;
+
+        if ($elapsed >= $this->lifetime) {
```
dantleech, line 45: this coudl extend some cache entries to double the max lifetime, might be worth reducing this or doing it every X fetches

PR #1662, opened 2022-05-26
Adding configuration reference for the Console extension

mamazu:


lib/Extension/Console/ConsoleExtension.php
```diff
@@ -39,8 +41,8 @@ public function load(ContainerBuilder $container): void
 
         $container->register(self::SERVICE_OUTPUT, function (Container $container) {
             return new ConsoleOutput(
-                $container->getParameter('console.verbosity'),
-                $container->getParameter('console.decorated')
+                $container->getParameter(self::CONSOLE_VERBOSITY),
+                $container->getParameter(self::CONSOLE_DECORATED)
```
dantleech, line 45: convention is `self::PARAM_VERBOSITY`

doc/reference/configuration.rst
```diff
@@ -1088,9 +1088,15 @@ ConsoleExtension
 
 
 
+Verbosity level
+
+
 **Default**: ``32``
 
 
+**Allowed values**: 16, 32, 64, 128, 256
```
dantleech, line 1097: this might be a bit misleading, currently phpactor's verbosity is rather binary (either it shows logs or it doesn't). could be wrong though.

dantleech (approved):

dantleech:
:bow:

doc/reference/configuration.rst
mamazu: I am sure that phpactor isn't the only project who only uses verbose and non verbose logging. But from the documentation standpoint this is correct. You could add it to the description though, that only <16 and >16 are important values.

PR #1667, opened 2022-05-28
Index infos refactoring

dantleech:
I started to refactor and went down a bit of a rabbit hole and have given up:

- Ask for the index to remove in a loop and re-render the updated list
- Auto-completion for index name / number
- Show last modified and age (the previous version was incorrect)

image

Unfortunately there seems to be a bug in Symonys Section output when used with the Question and the output is corrupted:

image


dantleech:
cc @mamazu I might need to unpick this and port only the "fixes" back.

mamazu:
Did we want to refactor this to use the file index class anyways?

dantleech:
it would be good to have the abstraction, but might be premature

mamazu:
In this merge request you have removed the function to clean all indicies. Is that on purpose or was that on accident?

mamazu:
In this merge request you have removed the function to clean all indicies. Is that on purpose or was that on accident?

dantleech:
i think i was looking for another way, would have re-included it I think,

PR #1688, opened 2022-06-03
Add support for @var dockblock with variable followed by type

lumnn:
Some of the software I have to work with comes with `@var` definitions with variable name first and then type. For example:

```
/** @var $date \DateTime */
```

This pull request adds a support for that

The reason for changing `var3.test` file instead of creating a new one is that `var2.test` seems identical.

### DIFF
--- a/lib/WorseReflection/Bridge/TolerantParser/Diagnostics/WrongVarOrderProvider.php
+++ b/lib/WorseReflection/Bridge/TolerantParser/Diagnostics/WrongVarOrderProvider.php
@@ -1,0 +1,43 @@
+<?php
+
+namespace Phpactor\WorseReflection\Bridge\TolerantParser\Diagnostics;
+
+use Generator;
+use Microsoft\PhpParser\Node;
+use Phpactor\DocblockParser\Ast\Tag\InvertedVarTag;
+use Phpactor\DocblockParser\DocblockParser;
+use Phpactor\TextDocument\ByteOffsetRange;
+use Phpactor\WorseReflection\Core\DiagnosticProvider;
+use Phpactor\WorseReflection\Core\Inference\Frame;
+use Phpactor\WorseReflection\Core\Inference\NodeContextResolver;
+
+class WrongVarOrderProvider implements DiagnosticProvider
+{
+    private DocblockParser $docblockParser;
+
+    public function __construct(DocblockParser $docblockParser = null)
+    {
+        $this->docblockParser = $docblockParser ?: DocblockParser::create();
+    }
+
+    public function provide(NodeContextResolver $resolver, Frame $frame, Node $node): Generator
+    {
+        $comment = $node->getDocCommentText();
+
+        if (!$comment) {
+            return;
+        }
+
+        $parsedComment = $this->docblockParser->parse($node->getLeadingCommentAndWhitespaceText());
+        $invertedVarTags = $parsedComment->descendantElements(InvertedVarTag::class);
+
+        $offset = $node->getFullStartPosition();
+
+        foreach ($invertedVarTags as $invertedVarTag) {
+            $start = $offset + $invertedVarTag->start();
+            $end = $offset + $invertedVarTag->end();
+
+            yield new WrongVarOrderDiagnostic(ByteOffsetRange::fromInts($start, $end), '@var type must come before element name');
+        }
+    }
+}
--- a/lib/DocblockParser/Ast/Tag/InvertedVarTag.php
+++ b/lib/DocblockParser/Ast/Tag/InvertedVarTag.php
@@ -1,0 +1,12 @@
+<?php
+
+namespace Phpactor\DocblockParser\Ast\Tag;
+
+class InvertedVarTag extends VarTag
+{
+    protected const CHILD_NAMES = [
+        'tag',
+        'variable',
+        'type',
+    ];
+}
--- a/lib/DocblockParser/Parser.php
+++ b/lib/DocblockParser/Parser.php
@@ -4,27 +4,31 @@ namespace Phpactor\DocblockParser;
 
 use Phpactor\DocblockParser\Ast\ArrayKeyValueList;
 use Phpactor\DocblockParser\Ast\ArrayKeyValueNode;
-use Phpactor\DocblockParser\Ast\Tag\DeprecatedTag;
 use Phpactor\DocblockParser\Ast\Docblock;
+use Phpactor\DocblockParser\Ast\Node;
+use Phpactor\DocblockParser\Ast\ParameterList;
+use Phpactor\DocblockParser\Ast\TagNode;
+use Phpactor\DocblockParser\Ast\Tag\DeprecatedTag;
 use Phpactor\DocblockParser\Ast\Tag\ExtendsTag;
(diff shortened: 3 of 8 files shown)


dantleech:
hmm, this variation is not supported by PHPStan, not sure if it's supported by Psalm and it is not the "standard": https://docs.phpdoc.org/3.0/guide/references/phpdoc/tags/var.html

is it supported by PHPStorm?

i think it would be better to update the existing code to use the correct syntax

lumnn:
It's used in quite large project - Magento 2.
Example: https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/Catalog/view/frontend/templates/product/list.phtml#L13
_Randomly selected this example, and noticed that even they cannot be consistent in same file_

I can try PR-ing them with fix, but my last attempt at correcting simple code duplication in that repository wait's now 1.5 year to merge...

I don't know about PHPStorm support for that.

I understand Psalm or any other linter/analyser not supporting it. From LSP server perspective I think it should understand it and provide language support for that syntax.
Ideally Psalm or any analyser should pick-up the incorrect syntax and mark it with warning, so user is not getting confidence over wrong syntax.

I also believe, because of simplicity of the change it may be okay for it to be included.  
I guess test could be removed or the changes could be commented to indicate that, this is not something essential to keep working in potential future refactoring or changes as it's not standard syntax.

dantleech:
just for context: if it's a vendor file, how does it affect users of Phpactor (assuming they are not editing the vendor file)?

lumnn:
Magento allows you to override templates (the example referenced) by placing the files in correct folder outside.
In that case you're free to do whatever you want with that file including correcting the docblock statements and allowing phpactor work properly. This however gives additional things to re-do during upgrades.

Additionally, vendor templates may be used for debugging purposes and understanding the code - for me it's mostly hover definitions and goto definition shortcuts. Magento rendering engine adds a $block variable into template context - this behaviour is unknown for IDEs, but most of the time templates come with `@var` docblocks.

---

I've checked PHPStan and Psalm for support of this inverted syntax, and both of them doesn't support this neither raise any error.

I'm unable to check this on PHPStorm

lumnn:
I did create PR at source https://github.com/magento/magento2/pull/35576 to correct that problem at source. Will see how quickly this one will be merged.
It doesn't change a fact that the issue will be still revelant to pre-merge releases and probably some other code not following phpDocumentor style strictly

dantleech:
hm, we could potentially add a warning diagnostic for this invalid syntax but still support it.

can you add another test here?

```
diff --git a/lib/WorseReflection/Tests/Inference/variable/type_from_var_docblock_invalid.test b/lib/W>
new file mode 100644
index 00000000..40914ba1
--- /dev/null
+++ b/lib/WorseReflection/Tests/Inference/variable/type_from_var_docblock_invalid.test
@@ -0,0 +1,6 @@
+<?php
+
+/** @var $foo string */
+$foo;
+
+wrAssertType('string', $foo);
```

lumnn:
> add a warning diagnostic for this invalid syntax

I'll get my head around this

dantleech:
> I'll get my head around this

Diagnostics are an in-development feature, e.g. https://github.com/phpactor/phpactor/blob/97ff31d07cfbf20636023decb29bd43871dc250f/lib/WorseReflection/Bridge/TolerantParser/Diagnostics/MissingMethodProvider.php#L19 and they are not fully integrated, however you can test them with `phpactor worse:analyse src/Path/To/File.php`

But it would be enough to create an issue if you didn't want to dive into that.

dantleech:
looks like you need to add a `/** @phpstan-ignore-next-line */` annotation in the `UseStatementUpdater` (parser says nodes are not nullable but they really are). not sure why it's triggering now.

dantleech:
fixed master....

lumnn:
I have to call it a day, so I send the progress I have so far.

In order to make that working I've created `Phpactor\DocblockParser\Ast\Tag\InvertedVarTag` class as the `VarTag` didn't work as expected with `toString()` method (and probably some other methods and code which expects same order of childrens as it's in code).
It does feel little wrong as this could lead a codebase to have plenty of AST tags for wrong syntax, but I couldn't think of any better solution.

Another question would be how do I get `DocblockParser` in `WrongVarOrderProvider` class. What I've done is I created a constructor with argument, which by default is null and `DocblockParser` is instantiated if not provided. However in that case this doesn't re-use cache (not even sure if it should?)

Lastly it's not yet mergable as there are no tests for it.

lib/WorseReflection/Bridge/TolerantParser/Diagnostics/WrongVarOrderProvider.php
```diff
@@ -0,0 +1,43 @@
...
+use Phpactor\TextDocument\ByteOffsetRange;
+use Phpactor\WorseReflection\Core\DiagnosticProvider;
+use Phpactor\WorseReflection\Core\Inference\Frame;
+use Phpactor\WorseReflection\Core\Inference\NodeContextResolver;
+
+class WrongVarOrderProvider implements DiagnosticProvider
+{
+    private DocblockParser $docblockParser;
+
+    public function __construct(DocblockParser $docblockParser = null)
```
dantleech, line 18: this should be fine, we can refactor to inject the [CachedDocblockParser](https://github.com/phpactor/phpactor/blob/581bd4dc40250fcb2f95d1428ff1790e1ee74a21/lib/WorseReflection/Core/ServiceLocator.php#L98-L101)

caching is pretty crucial here

dantleech:
For testing see https://github.com/phpactor/phpactor/blob/581bd4dc40250fcb2f95d1428ff1790e1ee74a21/lib/WorseReflection/Tests/Integration/Bridge/TolerantParser/Diagonstics/MissingMethodProvider/InlinedType.test

- Create a test case: https://github.com/phpactor/phpactor/blob/581bd4dc40250fcb2f95d1428ff1790e1ee74a21/lib/WorseReflection/Tests/Integration/Bridge/TolerantParser/Diagonstics/MissingDocblockProviderTest.php#L9
- Run the test (it should fail and tell you to create a directory)
- Put some source code in the directory
- Run the test (it will prompt you to create a method in your test case)
- Run assertions on the provided diagnostics

lib/WorseReflection/Bridge/TolerantParser/Diagnostics/WrongVarOrderProvider.php
```diff
@@ -0,0 +1,43 @@
...
+        $parsedComment = $this->docblockParser->parse($node->getLeadingCommentAndWhitespaceText());
+        $invertedVarTags = $parsedComment->descendantElements(InvertedVarTag::class);
+
+        $offset = $node->getFullStartPosition();
+
+        foreach ($invertedVarTags as $invertedVarTag) {
+            $start = $offset + $invertedVarTag->start();
+            $end = $offset + $invertedVarTag->end();
+
+            yield new WrongVarOrderDiagnostic(ByteOffsetRange::fromInts($start, $end), '@var type must come before element name');
```
dantleech, line 40: s/must/should

dantleech:
closing this PR for now, i'm not incredibly enthusiastic about this change (IMO would be better to fix all the docblocks to the correct standard with a regex or similar), but feel free to comment / re-open if you want to continue to work on it.

PR #1691, opened 2022-06-03
WIP: Applying index cleaning command improvements

mamazu:
replaces #1667


mamazu:
The implementation of the "all" feature isn't the best. But I couldn't really think of a better way to do that, so feel free to nudge me in the right direction.

Otherwise this should be all there is from your merge request.

dantleech:
nice work!

note that one of the reasons I stopped is because the layout breaks after deleting an index:

```
+----+----------------------------------+----------+-----------+-----------+
| #  | Directory                        | Size     | Age       | Modified  |
+----+----------------------------------+----------+-----------+-----------+
| #  | Directory                        | Size     | Age       | Modified  |
+----+----------------------------------+----------+-----------+-----------+
| 1  | language-server-bbd062           | 36.69 M  | 35.6 days | 18.1 days |
| 2  | language-server-protocol-e5762e  | 30.46 M  | 18.1 days | 18.1 days |
| 3  | lib-e19fd4                       | 25.21 M  | 7.5 days  | 6.1 days  |
| 4  | php-5fec06                       | 19.54 M  | 3.5 days  | 3.5 days  |
```

but maybe this is fine until it's fixed upstream.

lib/Indexer/Extension/Command/IndexCleanCommand.php
```diff
@@ -52,130 +57,145 @@ protected function configure(): void
...
-                if ($indexCount >= (int) $indexString) {
-                    $indexToDelete = ((int) $indexString) - 1;
+        while (true) {
+            $this->renderIndexTable($indicies, $section);
+            if ($index) {
+                $section->writeln(sprintf('<info>Removed index "%s"</>', $index->name()));
+            }
+            try {
+                // pass $output instead of $section because the interactive input
+                // is corrupted with he Section.
```
dantleech, line 101: s/he/the

lib/Indexer/Extension/Command/IndexCleanCommand.php
```diff
@@ -52,130 +57,145 @@ protected function configure(): void
...
-                sprintf('%.1f days', $index->lastUpdatedInDays()),
+                sprintf('%.1f days', $index->ageInDays()),
+                sprintf('%.1f days', $index->lastModifiedInDays()),
             ]);
-            $progressbar->advance();
         }
-        $progressbar->finish();
-        $output->writeln('');
+        $table->addRow(new TableSeparator());
+        $table->addRow(['Σ', self::OPT_CLEAN_ALL, PhpactorFilesystem::formatSize($indexList->totalSize()), '', '']);
```
dantleech, line 144: :+1:

lib/Indexer/Extension/Command/IndexCleanCommand.php
mamazu: Why do we do the section thing in the first place? Just to reset it? I wouldn't mind just printing the table again.

dantleech: with the section it should keep the text in the same place in the console instead of scrolling off the screen, so looking for like a normal "interface". i think printing it again could look confusing ...

dantleech:
:boom:

PR #1698, opened 2022-06-04
Import name diagnostics

dantleech:
This PR moves the responsiblity for finding unresolved names to the Diagnostics


doc/reference/configuration.rst
```diff
@@ -774,6 +774,21 @@ WorseReflectionExtension
 ------------------------
 
 
+.. _param_language_server_code_transform.import_globals:
```
dantleech, line 777: will move all the diagnostics to a dedicated package and rename this accordingly. leaving as is for now for b/c

PR #1727, opened 2022-06-27
update[doc]: assign Phpactor commands to only php files

rambhosale:
The following change attaches the phpactor lsp commands to only PHP filetype


dantleech:
great, thanks!

PR #1756, opened 2022-07-09
Code Action and Diagnostic Cancellation

dantleech:
Adds cancellation handling to the aggregate providers.


BladeMF:
I am guessing `AggregateCodeActionProvider` already has it as an argument?

dantleech:
yes it's in 3.0.0 of the language server

BladeMF:
I did the token checking there - between each provider - `Delay(0)` and `$token->isRequested()`.

PR #1773, opened 2022-07-13
Improved keyword completion

dantleech:
- Support `function`, `class`, `interface`, `trait` and `enum` keywords
- Add `__construct` (not a keyword but useful, can be moved to something more suitable)


DanielSiepmann:
>     * Add `__construct` (not a keyword but useful, can be moved to something more suitable)

Wouldn't it make sense to support all magic methods then?
https://www.php.net/manual/en/language.oop5.magic.php

That would also provide a proper naming, not keyword but magic methods.
What do you think?

dantleech:
> Wouldn't it make sense to support all magic methods then?

nobody uses those right? ( :smile: ) but this was really a small optimisation because the number of times I write `__construct` per day. really there should be a dedicated completor for `__construct`, magic methods, and overridden methods etc.

DanielSiepmann:
I barely wrote those. But depends on use cases. `__invoke` becomes more popular with EventListeners, at least within TYPO3 CMS.

But I never used autocompletion for those kind of things. I've a snippet. Just noticed this PR and thought the feedback might help. I guess you expect those to work if `__construct` works.

Issue #1030, opened 2020-05-06
[RPC][documentation] Missing documentation for "language" parameter


tkotosz:
It seems some commands like `goto_definition` has a language parameter which need to be provided otherwise it defaults to php.

For example this is needed for the [behat-extension](https://github.com/phpactor/behat-extension) otherwise it will do nothing.

tkotosz:
Relevant part of the behat extension: https://github.com/phpactor/behat-extension/blob/master/lib/ReferenceFinder/StepDefinitionLocator.php#L40

dantleech:
I think ultimately we should provide generated documentation for all this: https://github.com/phpactor/phpactor/issues/1032

tkotosz:
Make sense :+1:

PR #1774, opened 2022-07-14
RPC command documentation generator

mamazu:
## What I did
* Generalized the `DocumentExtensionsCommand` to also be able to generate documentation for the commands
* Centralized / extracted the generation of the `Definition` printing logic to another class
* Added a command for generating configuration information from a `Handler`.

## What needs to be done
First of all this is only a prove of concept if the idea that I had is alright. If it is valid, then we need to add more `Handler`s to the list to document. And fill out the documentation info on all of the handlers

Relates to #1030

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,8 +27,8 @@ vimdoc: build/bin/vimdoc
 	vimdoc .
 
 configreference:
-	./bin/phpactor development:generate-documentation extension.documentor > doc/reference/configuration.rst
-	./bin/phpactor development:generate-documentation rpc.documentor > doc/reference/rpc_command.rst
+	./bin/phpactor development:generate-documentation extension > doc/reference/configuration.rst
+	./bin/phpactor development:generate-documentation rpc > doc/reference/rpc_command.rst
 
 # Put it first so that "make" without argument is like "make help".
 help:
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ build/vimdoc:
 
 build/vimdoc/build: build/vimdoc
 	cd build/vimdoc; python3 setup.py config
-	cd build/vimdoc; python3 setup.py build 
+	cd build/vimdoc; python3 setup.py build
 
 build/bin/vimdoc: build/vimdoc/build
 	cd build/vimdoc; python3 setup.py install --user
@@ -28,6 +28,7 @@ vimdoc: build/bin/vimdoc
 
 configreference:
 	./bin/phpactor development:configuration-reference > doc/reference/configuration.rst
+	./bin/phpactor development:command-reference > doc/reference/rpc_command.rst
 
 # Put it first so that "make" without argument is like "make help".
 help:
--- a/Makefile
+++ b/Makefile
@@ -27,8 +27,8 @@ vimdoc: build/bin/vimdoc
 	vimdoc .
 
 configreference:
-	./bin/phpactor development:configuration-reference > doc/reference/configuration.rst
-	./bin/phpactor development:command-reference > doc/reference/rpc_command.rst
+	./bin/phpactor development:generate-documentation extension.documentor > doc/reference/configuration.rst
+	./bin/phpactor development:generate-documentation rpc.documentor > doc/reference/rpc_command.rst
 
 # Put it first so that "make" without argument is like "make help".
 help:
--- a/doc/reference/command.rst
+++ b/doc/reference/command.rst
@@ -1,0 +1,75 @@
+Commands
+=============
+
+
+.. This document is generated via the `development:command-reference` command
+
+
+.. contents::
+   :depth: 2
+   :backlinks: none
+   :local:
+
+
+.. _GotoDefinitionHandler:
+
+
+GotoDefinitionHandler
+---------------------
+
+
+.. _param_language:
+
+
+``language``
+""""""""""""
+
+
+Type: string
+
+
+**Default**: ``"php"``
+
+
+.. _param_target:
+
+
+``target``
+""""""""""
+
+
+Type: string
+
+
+**Default**: ``"focused_window"``
+
+
+.. _param_offset:
+
+
+``offset``
+""""""""""
+
+
+**Default**: ``null``
+
+
+.. _param_source:
+
+
+``source``
+""""""""""
+
+
+**Default**: ``null``
+
+
+.. _param_path:
+
+
+``path``
+""""""""
+
+
+**Default**: ``null``
+
--- a/doc/reference/configuration.rst
+++ b/doc/reference/configuration.rst
@@ -2,7 +2,7 @@ Configuration
 =============
 
 
-.. This document is generated via the `development:configuration-reference` command
+.. This document is generated via the `development:generate-documentation` command
 
 
 .. contents::
--- a/doc/reference/configuration.rst
+++ b/doc/reference/configuration.rst
@@ -2,7 +2,7 @@ Configuration
 =============
 
 
-.. This document is generated via. the `documentation:configuration-reference` command
+.. This document is generated via the `development:configuration-reference` command
 
 
 .. contents::
@@ -25,8 +25,6 @@ CoreExtension
 """"""""""""""""""""""""""
 
 
-
-
 Name of the "dumper" (renderer) to use for some CLI commands
 
 
@@ -38,8 +36,6 @@ Name of the "dumper" (renderer) to use for some CLI commands
 
 ``xdebug_disable``
 """"""""""""""""""
-
-
 
 
 If XDebug should be automatically disabled
@@ -55,8 +51,6 @@ If XDebug should be automatically disabled
 """""""""""
 
 
-
-
 Internal use only - name of the command which was executed
 
 
@@ -68,8 +62,6 @@ Internal use only - name of the command which was executed
 
 ``core.warn_on_develop``
 """"""""""""""""""""""""
-
-
 
 
 Internal use only: if an warning will be issed when on develop, may be removed in the future
@@ -85,8 +77,6 @@ Internal use only: if an warning will be issed when on develop, may be removed i
 """""""""""""""""""""""""
 
 
-
-
 Ensure that PHP has a memory_limit of at least this amount in bytes
 
 
@@ -98,8 +88,6 @@ Ensure that PHP has a memory_limit of at least this amount in bytes
 
 ``$schema``
 """""""""""
-
-
 
 
 Path to JSON schema, which can be used for config autocompletion, use phpactor config:initialize to update
@@ -122,8 +110,6 @@ ClassToFileExtension
 """"""""""""""""""""""""""""""
 
 
-
-
 Root path of the project (e.g. where composer.json is)
 
 
@@ -135,8 +121,6 @@ Root path of the project (e.g. where composer.json is)
 
 ``class_to_file.brute_force_conversion``
 """"""""""""""""""""""""""""""""""""""""
-
-
 
 
 If composer not found, fallback to scanning all files (very time consuming depending on project size)
@@ -159,8 +143,6 @@ CodeTransformExtension
 """""""""""""""""""""""""""""""""""""
 
 
-
-
 Variants which should be suggested when class-create is invoked
 
 
@@ -172,8 +154,6 @@ Variants which should be suggested when class-create is invoked
 
 ``code_transform.template_paths``
 """""""""""""""""""""""""""""""""
-
-
 
 
 Paths in which to look for code templates
@@ -189,8 +169,6 @@ Paths in which to look for code templates
 """"""""""""""""""""""""""""""
 
 
-
-
 Indentation chars to use in code generation and transformation
 
 
@@ -202,8 +180,6 @@ Indentation chars to use in code generation and transformation
 
 ``code_transform.refactor.generate_accessor.prefix``
 """"""""""""""""""""""""""""""""""""""""""""""""""""
-
-
 
 
 Prefix to use for generated accessors
@@ -219,8 +195,6 @@ Prefix to use for generated accessors
 """"""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
 
 
-
-
 If the first letter of a generated accessor should be made uppercase
 
 
@@ -232,8 +206,6 @@ If the first letter of a generated accessor should be made uppercase
 
 ``code_transform.import_globals``
 """""""""""""""""""""""""""""""""
-
-
 
 
 Import functions even if they are in the global namespace
@@ -249,8 +221,6 @@ Import functions even if they are in the global namespace
 """"""""""""""""""""""""""""""""""""""""""""
 
 
-
-
 Object fill refactoring: show hint as a comment
 
 
@@ -262,8 +232,6 @@ Object fill refactoring: show hint as a comment
 
 ``code_transform.refactor.object_fill.named_parameters``
 """"""""""""""""""""""""""""""""""""""""""""""""""""""""
-
-
 
 
 Object fill refactoring: use named parameters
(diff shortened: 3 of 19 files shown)


doc/reference/command.rst
```diff
@@ -0,0 +1,75 @@
...
+
+
+.. _GotoDefinitionHandler:
+
+
+GotoDefinitionHandler
+---------------------
+
+
+.. _param_language:
```
dantleech, line 21: needs to include `_RpcCommand_GlobalDefinitionHandler_language` or similar as these are global references

lib/Extension/Debug/Model/CommandDocumentor.php
```diff
@@ -0,0 +1,98 @@
+<?php
+
+namespace Phpactor\Extension\Debug\Model;
+
+use Phpactor\MapResolver\Resolver;
+use ReflectionClass;
+use RuntimeException;
+use Phpactor\Extension\Rpc\Handler;
+
+class CommandDocumentor implements DocumentorInterface
```
dantleech, line 10: `RpcCommand` documentor to distinguish it from LSP commands

lib/Extension/Debug/Model/DocumentorInterface.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace Phpactor\Extension\Debug\Model;
+
+interface DocumentorInterface
```
dantleech, line 5: we don't use suffixes, this can be just `Documentor`

doc/reference/command.rst
```diff
@@ -0,0 +1,75 @@
+Commands
+=============
```
dantleech, line 2: should be same length as title text

doc/reference/command.rst
```diff
@@ -0,0 +1,75 @@
+Commands
```
dantleech, line 1: rename file to `rpc_command.rst`

Makefile
```diff
@@ -28,6 +28,7 @@ vimdoc: build/bin/vimdoc
 
 configreference:
 	./bin/phpactor development:configuration-reference > doc/reference/configuration.rst
+	./bin/phpactor development:command-reference > doc/reference/command.rst
```
dantleech, line 31: better to dump to `rpc_command.rst`

lib/Extension/Debug/Model/DocumentorInterface.php
mamazu: Fixed

Makefile
mamazu: Fixed

lib/Extension/Debug/Model/CommandDocumentor.php
mamazu: Fixed

doc/reference/command.rst
mamazu: Do we have an example of non global handlers?

dantleech:
```
/home/runner/work/phpactor/phpactor/doc/reference/rpc_command.rst:18:duplicate label gotodefinitionhandler, other instance in /home/runner/work/phpactor/phpactor/doc/reference/command.rst
```

dantleech:
Currently the handlers are aggregated here: https://github.com/phpactor/phpactor/blob/2b995dce658e5d1e1a3cf6137a66773ff077efed/lib/Extension/Rpc/RpcExtension.php#L47

it would probably make sense to add an `all()` method to the `HandlerRegistry` interface

mamazu:
For the extension it just takes the parameter from the container and as far as I can see the parameter is filled with a hard coded array somewhere.

mamazu:
But adding it through the registry should be easy and clean to do.

dantleech:
grr, the tests are trying to tell us something. DebugExtension probably shouldn't depend on RpcExtension, rather the RPC extension should probably have this responsiblity.

Let's move the documentor implementation for RPC to the RPCExtension, we can then do something like:

```
$container->register(DocumentorRegistry::class, function (Container $container) {
    $serviceMap = [];
    foreach ($container->getServicesByTag(self::TAG_DOCUMENTOR) as $serviceId => $attributes) {
        if (isset($attributes['name'])) {
            throw new \RuntimeException('... or whatever other services do in this situation');
        }

        $serviceMap[$attributes['name']] = $serviceId;
    }

    return new ContainerDocumentorRegistry($container, $serviceMap);
}
```

We can then have a single documentor command:

```
class DocumentorCommand ...
{
    public function execute(...): int
    {
         $documentor = $this->registry->get((string)$input->getArgumnet('documentor'));
         $output->write($documentor->document());
         return 0;
    }
}
```

Or whatever, if that makes sense. So:

```
phpactor development:documentation rpc_commands
```

mamazu:
Sounds like a good idea. I will do that.

mamazu:
This will still have the RPC bundle (or at least the `CommandDocumentor`) depend on the `Documentor` interface from the Debug extension. But I guess this is okay because the class is only used in the Debug Extension anyways.

mamazu:
Well, I would love to have it resolve at the container compilation stage. But I think that this isn't possible because then it would require the debug extension to be loaded last. Because getting the error at execution time (aka when generating the documentation) maybe is too late. Especially since the error will be redirected to the rst file.

dantleech:
> This will still have the RPC extension (or at least the CommandDocumentor) depend on the Documentor interface from the Debug extension

yes, but it's a step in the right direction, easy to move that interface to another extension or to a "component" later.

> Well, I would love to have it resolve all the Documentor stuff at the container compilation stage. But I think that this isn't possible because then it would require the debug extension to be loaded last.

if you collect services with tags it doesn't matter where the extension is loaded, the closure where you return the service is always executed when all the extensions are present.

Using the "container registry" pattern (as with the other registry) was necessary previously because commands were eagerly loaded for all operations, but that's no longer the case, but still might be better (e.g. doign `phpactor list` without the "reigstry" would cause all the documentors to be instantiated in order to create the command in order that Symfony can get the "description" etc)

mamazu:
Okay found my problem. Forgot a return in a bad place. It's working now. :D

But talking about the registry pattern. Wouldn't it maybe make sense to generalize it so that you can just call a method with a tag and it creates a registry with services that have that tag? Since we are using different implementations in different places.

dantleech:
In that case you would end up having a new cross-package dependency for something which is pretty trivial to implement, and then one package will need something in the "registry" that the default implementation doesn't provide, etc etc. although one of the arguments was being type safe, with generics that a bit less valid, but still I think I'd prefer to keep them package-specific for now.

lib/Extension/Debug/DebugExtension.php
```diff
@@ -27,25 +48,24 @@ public function load(ContainerBuilder $container): void
...
         }, [
             ConsoleExtension::TAG_COMMAND => [
                 'name' => 'development:configuration-reference'
             ]
         ]);
 
         $container->register('generate_documentation_command.command', function (Container $container) {
-            return new GenerateDocumentationCommand($container->get(RpcCommandDocumentor::class));
+            return new GenerateDocumentationCommand(
+                $container->get(DocumentorRegistry::class)->get(RpcExtension::RPC_DOCUMENTOR_NAME)
```
dantleech, line 67: we need to inject the registry so that the documentors are _lazily_ instantiated (this is the main reason for having a registry). As mentioned before Symfony commands are pro-actively instantiated by default, and they are all instantiated when doing something like `phpactor list`, which means that all the documentors will be instantiated too.

lib/Extension/Rpc/RpcExtension.php
```diff
@@ -61,6 +63,17 @@ public function load(ContainerBuilder $container): void
             return new LazyContainerHandlerRegistry($container, $handlers);
         });
 
+        if (class_exists(DebugExtension::class)) {
```
dantleech, line 66: I thnk it would eb fine without the class_exists?

lib/Extension/Debug/DebugExtension.php
mamazu: The phpactor list works. That's not the problem. The problem seems to be the test since the Debug extension now doesn't work in isolation anymore.

dantleech: I think we're not on the same page:

We inject the registry:

```
class FooCommand {
    public function __cosntruct(DocumentorRegistry $registry) {
        parent::__construct();
        $this->registry = $registry;
    }

    public function execite(InputArgument $input, // ...): int
    {
        $documentor = $this->registry->getByName($input->getArgument('documentor'));
        $documentor->document(); // etc
    }
}
```

the command does not know or care about what implementations it has, it just pulls a named documentor from the registry.

lib/Extension/Rpc/RpcExtension.php
mamazu: Yes, I removed it.

lib/Extension/Debug/Command/GenerateDocumentationCommand.php
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    protected function configure(): void
+    {
+        $this->setDescription('Generate configuration reference as an RST document');
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $documentor = $this->documentorRegistry->get($this->documentorName);
```
dantleech, line 30: i was rather thinking:

```
$documentor = $this->documentorRegistry->get((string)$input->getArgumnet('documentor'));
```

so
```
phpactor whateverthecommandis:document rpc_commands
```

so one command for all

Makefile
```diff
@@ -28,6 +28,7 @@ vimdoc: build/bin/vimdoc
 
 configreference:
 	./bin/phpactor development:configuration-reference > doc/reference/configuration.rst
+	./bin/phpactor development:command-reference > doc/reference/rpc_command.rst
```
dantleech, line 31: if done as i suggested this would be `development:document rpc-commands` and `development:document configuration`, if that makes sense

doc/reference/rpc_command.rst
```diff
@@ -0,0 +1,1227 @@
...
+
+
+.. _FileInfoHandler:
+
+
+FileInfoHandler
+---------------
+
+
+.. _RpcCommand_GlobalDefinitionHandler_path:
```
dantleech, line 38: this is duplicated and is the reason for the lint failure: should be `FileInfoHandler`

doc/reference/rpc_command.rst
```diff
@@ -0,0 +1,1227 @@
+Commands
```
dantleech, line 1: Legacy RPC Commands?

lib/Extension/Rpc/RpcCommandDocumentor.php
```diff
@@ -0,0 +1,85 @@
...
+            str_repeat('-', mb_strlen($documentedName)),
+            "\n",
+        ];
+
+        $resolver = new Resolver();
+        $handler->configure($resolver);
+
+        $hasDocumentation = false;
+        foreach ($resolver->definitions() as $definition) {
+            $help[] = $this->definitionDocumentor->document('RpcCommand_GlobalDefinitionHandler', $definition);
```
dantleech, line 75: need to replace GlobalDefinitionHandler with the name of the handler

lib/Extension/Debug/DebugExtension.php
```diff
@@ -7,27 +7,55 @@
...
+use Phpactor\Extension\Debug\Model\DocumentorRegistry;
+use Phpactor\Extension\Debug\Model\DefinitionDocumentor;
 use Phpactor\Extension\Core\Model\JsonSchemaBuilder;
 use Phpactor\MapResolver\Resolver;
+use RuntimeException;
 
 class DebugExtension implements Extension
 {
+    public const TAG_DOCUMENTOR = 'debug.documentor';
+    public const EXTENSION_DOCUMENTOR_NAME = 'extension.documentor';
```
dantleech, line 21: - `documentor` is superflous in this case I think, should just be `extension`

lib/Extension/Rpc/RpcExtension.php
```diff
@@ -19,12 +21,12 @@
 
 class RpcExtension implements Extension
 {
-    const TAG_RPC_HANDLER = 'rpc.handler';
+    public const TAG_RPC_HANDLER = 'rpc.handler';
     public const SERVICE_REQUEST_HANDLER = 'rpc.request_handler';
+    public const RPC_DOCUMENTOR_NAME= 'rpc.documentor';
```
dantleech, line 26: should be `rpc_commands`

mamazu:
Done.

dantleech:
doclinter still failing: `/home/runner/work/phpactor/phpactor/doc/reference/rpc_command.rst:85:Duplicate explicit target name: "rpccommand_globaldefinitionhandler_path".`

mamazu:
Could it be that two classes have the same property that they have to configure? Because currently all definitions have the same prefix.

lib/Extension/Rpc/RpcCommandDocumentor.php
dantleech: i.e. something like `document(sprintf('rpc_command_%s', $hadnler->name()), $definition)` cc @mamazu

dantleech:
yep see comment about the naming

mamazu:
Goto Type and Goto Definition have the same handler name.

mamazu:
Does that mean that this isn't possible then? I mean we could always use the classname instead of the name of the handler. That has to be unique.

dantleech:
if they have the same name then it's theoretically broken, goto type should have the name `goto_type`

mamazu:
But this seems to be an issue of the handler right?

dantleech:
sorry yes, jsut need to update the NAME. no idea why I closed this PR, think I may have got it mixed up with another

doc/reference/configuration.rst
```diff
@@ -925,9 +817,7 @@ FilePathResolverExtension
 """""""""""""""""""""""""""""""""""
 
 
-
-
-**Default**: ``"\/home\/daniel\/www\/phpactor\/phpactor"``
+**Default**: ``"\/home\/mamazu\/packages\/phpactor\/phpactor"``
```
mamazu, line 820: Well, this is also something that we might want to parameterize in the documentation.

dantleech: fine to leave it i you are, otherwise feel free to change to `daniel` :smile:

doc/reference/rpc_command.rst
```diff
@@ -0,0 +1,1227 @@
...
+
+
+``type``
+""""""""
+
+
+**Default**: ``"formatted"``
+
+
+.. _FileInfoHandler:
```
dantleech, line 31: this needs to be prefixed also ideally, `_Rpc_FileInfoHandler` or similar, but not super important.

doc/reference/configuration.rst
mamazu: Well the thing is that every time you run composer integrate it generates it again so I always have to be careful not to commit it.

dantleech: like i say, i don't mind if it's committed tbh feel free to keep it as is or change it

dantleech: it will just get reverted to my path when i push the docs next :smile:

doc/reference/rpc_command.rst
mamazu: Opted for `_Rpc_Handler` in this case.

doc/reference/configuration.rst
mamazu: Haha, yeah. All good. I am just irritated about things I didn't change. :D

dantleech (approved):

doc/reference/rpc_command.rst
```diff
@@ -0,0 +1,1227 @@
...
+
+
+.. _RpcHandler_StatusHandler:
+
+
+_RpcHandler_StatusHandler
+-------------------------
+
+
+.. _RpcCommand_status_type:
```
dantleech, line 21: not sure why this is RpcCommand but it's for `_RpcHandler_`, should probably have been `RpcHandler_StatusHandler_Param_status_typ` or something (or simply remove the reference declaration), but it's fine :+1:

dantleech:
:boom: thanks!

PR #1777, opened 2022-07-14
WIP: Fixing the OffsetInfo

mamazu:


dantleech:
what's wrong with offset info? :)

mamazu:
Forgot that I created the MR. I wanted to fix the phpdoc comment. But seeing that this would just replace one warning with 2 new warnings, I think I should just close it. :D

PR #1784, opened 2022-07-16
Fixing the typing issues in the `ClassMemberQuery`

mamazu:
## What I did
* Re-generating the baseline to remove all already fixed errors (it might be helpful to turn on `reportUnmatchedIgnoredErrors` to remove the errors when they are fixed.
* Adding return types to some of the copy methods of the `ClassMemberQuery`. This will not cause a BC break since the class is final and nobody can extend it.
* Fixing the argument type definition for the `withMember` and `withClass` methods. It now shows the correct type alongside of mixed to satisfy the phpstan compiler. When we can use PHP8 for Union request, then we can remove that all together.

### DIFF
--- a/lib/ClassMover/Domain/Model/ClassMemberQuery.php
+++ b/lib/ClassMover/Domain/Model/ClassMemberQuery.php
@@ -49,7 +49,7 @@ final class ClassMemberQuery
         return new self();
     }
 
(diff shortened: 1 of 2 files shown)


dantleech:
10 down, 1000 to go :smile:

PR #1790, opened 2022-07-17
Gh 1627 structural loops

dantleech:
Adds a "hack" to avoid infinte loops.

The better solution would be to resolve the inheritance hierarchy one time, unfortunately due to the complex mess of decorators this is non-trivial. This solution will ensure that the LS doesn't lock-up.

### DIFF
--- a/lib/WorseReflection/Core/Reflector/ClassReflector/MemonizedReflector.php
+++ b/lib/WorseReflection/Core/Reflector/ClassReflector/MemonizedReflector.php
@@ -4,6 +4,7 @@ namespace Phpactor\WorseReflection\Core\Reflector\ClassReflector;
 
 use Closure;
 use Phpactor\WorseReflection\Core\Cache;
+use Phpactor\WorseReflection\Core\Exception\ClassNotFound;
 use Phpactor\WorseReflection\Core\Exception\NotFound;
 use Phpactor\WorseReflection\Core\Reflection\ReflectionEnum;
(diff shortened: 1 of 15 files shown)


lib/WorseReflection/Core/Reflector/ClassReflector/MemonizedReflector.php
```diff
@@ -68,10 +69,16 @@ public function reflectEnum($className): ReflectionEnum
         });
     }
     
-    public function reflectClassLike($className): ReflectionClassLike
+    public function reflectClassLike($className, $visited = [], string $prefix = self::CLASS_LIKE_PREFIX): ReflectionClassLike
```
dantleech, line 72: revert

lib/WorseReflection/Core/Reflector/CoreReflector.php
```diff
@@ -135,12 +135,20 @@ public function reflectEnum($className): ReflectionEnum
      *
      * @throws ClassNotFound
      */
-    public function reflectClassLike($className): ReflectionClassLike
+    public function reflectClassLike($className, array $visited = []): ReflectionClassLike
     {
         $className = ClassName::fromUnknown($className);
 
+        if (isset($visited[$className->__toString()])) {
+            throw new ClassNotFound(sprintf(
```
dantleech, line 143: create dedicated exception

Issue #1572, opened 2022-05-11
UX: Index progress notifications intrusive


madrussa:
The notifications when indexing a large project are quite intrusive and annoying. It also slows down and can lock up VS Code.

<img width="490" alt="image" src="https://user-images.githubusercontent.com/1470257/167817577-c43bdf6e-0e75-4071-b4ae-ded4b9db4aac.png">

dantleech:
the other thing we could do is implement the LSP progress notification which I guess is less intrusive.

madrussa:
Yeah, we have a rather large monorepo. We also have a lot of duplicate libraries in our vendor directors which I could ignore I guess.

I think I ended up with over 500 notifications which VS code wasn't happy about. 

Other LSP services show a box on the footer with any indexing progress instead of it being a bunch of notifications. This is my preferred option as one: you can tell it is indexing, two: you can see any progress and three: It allows an entry point to trigger actions such as reindex without having to use the command palette. 
<img width="626" alt="image" src="https://user-images.githubusercontent.com/1470257/167830771-100eff74-f5f1-47a6-ae0f-bfd2016a2329.png">

unansweredocd:
> the other thing we could do is implement the LSP progress notification which I guess is less intrusive and/or add an option to quieten it.

Using LSP progress would be awesome.

PR #1792, opened 2022-07-17
refactor: report indexer progress through `workDoneProgress`

innocenzi:
This PR updates the indexer to not send notifications messages through the LSP, and instead use `window/workDoneProgress`. This avoids getting spammed with notifications in VSC:

**Before:**
image

**After:**
image


innocenzi:
I didn't realize there was a pinned issue about this, but this PR fixes #1572

dantleech:
ha, by complete coincidence this is what i've been working on for the past hours.

it was quite simple except for the fact that `create` is actually a request and needs to wait for the response from the client, which required some refactoring and will need a new release of the language server package,

i'll close this and maybe steal the test case, good work!

Issue #1789, opened 2022-07-17
Renaming a class to Use make LSP crash


innocenzi:
https://share.cleanshot.com/NZSuAB

In a fresh Laravel project, I wanted to try out the refactor tools, so I renamed the `App\Models\User` class to `Use` and the LSP crashed and the output window opened up.

I'm not sure what causes the crash since renaming the class to something else does not trigger the error.

```
[Error - 10:18:22 AM] Request textDocument/codeAction failed.
  Message: Exception [Error] Call to a member function getValues() on null at /Users/enzoinnocenzi/Library/Application Support/Code/User/globalStorage/dantleech.vscode-phpactor/phpactor/lib/WorseReflection/Core/Reflection/Collection/ReflectionTraitCollection.php#27
  Code: -32603 
#0 /Users/enzoinnocenzi/Library/Application Support/Code/User/globalStorage/dantleech.vscode-phpactor/phpactor/lib/WorseReflection/Bridge/TolerantParser/Reflection/ReflectionClass.php(263): Phpactor\WorseReflection\Core\Reflection\Collection\ReflectionTraitCollection::fromClassDeclaration(Object(Phpactor\WorseReflection\Core\ServiceLocator), Object(Microsoft\PhpParser\Node\Statement\ClassDeclaration))
#1 /Users/enzoinnocenzi/Library/Application Support/Code/User/globalStorage/dantleech.vscode-phpactor/phpactor/lib/WorseReflection/Bridge/TolerantParser/Reflection/ReflectionClass.php(205): Phpactor\WorseReflection\Bridge\TolerantParser\Reflection\ReflectionClass->traits()
#2 /Users/enzoinnocenzi/Library/Application Support/Code/User/globalStorage/dantleech.vscode-phpactor/phpactor/lib/WorseReflection/Core/Virtual/VirtualReflectionClassDecorator.php(80): Phpactor\WorseReflection\Bridge\TolerantParser\Reflection\ReflectionClass->methods(Object(Phpactor\WorseReflection\Bridge\TolerantParser\Reflection\ReflectionClass))
#3 /Users/enzoinnocenzi/Library/Application Support/Code/User/globalStorage/dantleech.vscode-phpactor/phpactor/lib/CodeTransform/Adapter/WorseReflection/Transformer/CompleteConstructor.php(126): Phpactor\WorseReflection\Core\Virtual\VirtualReflectionClassDecorator->methods()
[...]

PR #1794, opened 2022-07-17
fix(reflection): prevent crash when traitNameList is null

innocenzi:
Fixes https://github.com/phpactor/phpactor/issues/1789


dantleech:
:bow:

Issue #1802, opened 2022-07-18
Virtual method return types static and $this don't work


vlada-dudr:
```php
<?php

/**
 * @method static bla()
 */
class h {
}

$a = new h;
$b = $a->bla();
```
Expected result: `bla` method returns h
Actual: Method "bla" does not exist on class "h"

Changing `static` to `$this` is same. `self` makes phpactor discover the function, but doesn't mean the same thing...

ging-dev:
@vlada-dudr try `@method static void bla()`

vlada-dudr:
Well, that results in, as expected, Phpactor picking up static method on class h. But I want to describe instance method returning `static` type i.e. the class bla() was called on, as in `new static()`. If `$this` would work, it would cover most of the use cases.

ging-dev:
> Well, that results in, as expected, Phpactor picking up static method on class h. But I want to describe instance method returning `static` type i.e. the class bla() was called on, as in `new static()`. If `$this` would work, it would cover most of the use cases.

Did you mean
```php
/** @method User get() */
class h {
}

$a = new h;
$b = $a->get() // $b is User
```
Or
```php
/** @method self get() */
class h {}

$a = new h;
$b = $a->get() // $b is h
```

vlada-dudr:
`static` and `$this` are almost same, except that `$this` requires return of same instance. Compare:
```php
class a {
    public function x(): self { return new self(); }
    public function y(): static {return new static(); }
}

class b extends a {
}

$o = new b();
assert(get_class($o->x) === 'a');
assert(get_class($o->y) === 'b');
```
So there are two meanings `static` has in `@method` context. A property of a method or a return type. So, I expect that phpactor should distinguish between them. This should be legal: `@method static static make()`. Which describes static function which return instance of class it was called on as opposed to `@method static self make()`, which is static method that should return instance of class the method is defined on. In case of `@method static setWhatever()` the return type should be `static` - the class the method was called on, not a static method of such class.

Is it more understandable now?

The other part of the problem is that line `@method $this setSomething()` is ignored by phpactor, so none of them are usable.

vlada-dudr:
I think it used to work some time ago. Maybe addition of virtual static methods did break parsing of static type? Because when I add `@method static bla()` I see nameless static method in the autocomplete suggestions.

ging-dev:
[...]

PR #1806, opened 2022-07-19
Allow parsing $this return type in docblock

vlada-dudr:
This fixes part of #1802. With this little patch this is now recognized:
```php
<?php
/** @method $this bla() */
class a {}

$o = new a;
$r = $->bla(); // $r is of type a
```
I am not sure though, if it cannot have some implications for condititional types parsing.

### DIFF
--- a/lib/DocblockParser/Parser.php
+++ b/lib/DocblockParser/Parser.php
@@ -184,7 +184,7 @@ final class Parser
             }
         }
 
-        if ($this->ifType()) {
+        if ($this->ifType() || $this->tokens->if(Token::T_VARIABLE)) {
             $type = $this->parseTypes();
         }
 
@@ -266,6 +266,10 @@ final class Parser
         }
 
         if ($this->tokens->current->type === Token::T_VARIABLE) {
+            if ($this->tokens->current->value === '$this') {
+                $variable = $this->tokens->chomp(Token::T_VARIABLE);
+                return new ThisNode($variable);
+            }
             return $this->parseConditionalType();
         }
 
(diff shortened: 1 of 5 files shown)


lib/DocblockParser/Parser.php
```diff
@@ -168,7 +168,7 @@ private function parseMethod(): MethodTag
             }
         }
 
-        if ($this->ifType()) {
+        if ($this->ifType() or $this->tokens->if(Token::T_VARIABLE)) {
```
dantleech, line 171: `||` instead of `or` as convention

dantleech (changes requested):
I think this looks good, could you add some tests? a snapshot test in `lib/DocblockParser/Tests/Unit/Printer/this.test`?

```
/**
 * @method $this 
 */
---
Docblock: = 
 ElementList: = /**
 * 
    // whatever the output should be here as reported by PHPUnit
 */
```

and then, more importantly, a test for your exact use case:

in `lib/WorseReflection/Tests/Inference/types/this-return-type-in-docblock.test`

```
<?php
/** @method $this bla() */
class a {}

$o = new a;
$r = $->bla();
wrAssertType('a', $r);

vlada-dudr:
So I added some test as requested. Actually I added more and let them commented out for now, so it can be seen what more can be done.

lib/DocblockParser/Parser.php
vlada-dudr: done

dantleech:
have had a look at this -- I think it would be better for me to do https://github.com/phpactor/phpactor/issues/1823 which should fix this in a better way..

dantleech:
Will be fixed in #1855

dantleech:
Cherry picked your fix and tests, should work now!

PR #1809, opened 2022-07-20
Better Completion Documentation 

dantleech:
Use the "hover" renderer to render the completion documentation:

image

Also renders short class names (i.e. the "name" only) in rendered documentation


lib/Completion/Bridge/TolerantParser/WorseReflection/WorseClassMemberCompletor.php
```diff
@@ -144,7 +150,11 @@ private function populateSuggestions(NodeContext $symbolContext, Type $type, boo
             yield Suggestion::createWithOptions($method->name(), [
                 'type' => Suggestion::TYPE_METHOD,
                 'short_description' => $this->formatter->format($method),
-                'documentation' => $method->docblock()->formatted(),
+                'documentation' => $this->objectRenderer->render(new HoverInformation(sprintf(
```
dantleech, line 153: this shouldn't be here ...

dantleech: fixed in master

PR #1821, opened 2022-07-24
Worse Reflection Refactoring

dantleech:
This PR aims to clean up the way WR resolves class members:

- Introduce a way to resolve the class inheritance hierarchy instead of recursive calls which can result in cycles and impacts cachablity.
- Parse all the members at the same time (not iterating once for each member type)

This PR will either enable the `@mixin` cycle issue to be resolved or will facilitate that resolution.

Potentially there are also performance gains from this work, although can't guarantee it will be significant.

TODO:

- [x] Remove concept of "virtual" decorator and add virtual methods directly in the reflection classes
- [x] Ensure this is at least as fast as previous version
- [-] ~Remove previous "visited" hack?~ Can't further work required to remove this
---

We are now slightly slower when doing things with class members in isolation, but overall there is a significant speed improvement.

image

Statically analysing WR is also slightly faster ~(although there are 4 more errors on master :thinking:~ because I've gixed 4 errors :smile: 
 
image


lib/WorseReflection/Core/Reflection/Collection/ClassLikeReflectionMemberCollection.php
```diff
@@ -0,0 +1,291 @@
...
+        return $new;
+    }
+
+    public function map(Closure $closure): ReflectionCollection
+    {
+        // TODO: this isn't very efficient
+        $new = new self([]);
+        foreach (self::MEMBER_TYPES as $collection) {
+            $new->$collection = array_map($closure, $this->$collection);
+            $new->items = array_map($closure, $this->items);
```
dantleech, line 286: can refactor these to only map once?

lib/WorseReflection/Tests/Inference/reflection/mixin_static.test
```diff
@@ -19,4 +19,6 @@ class B
 }
 
 $b = new B();
-wrAssertType('B', $b->doA());
+
+// TODO: This is wrong, fix before merging
+wrAssertType('A', $b->doA());
```
dantleech, line 24: this PR should at least preserve the old `@mixin` behavior

lib/WorseReflection/Bridge/TolerantParser/Reflection/ReflectionProperty.php
```diff
@@ -65,7 +67,11 @@ public function declaringClass(): ReflectionClassLike
 
     public function name(): string
     {
-        return (string) $this->variable->getName();
+        if ($this->name) {
+            return $this->name;
+        }
+        $this->name = (string) $this->variable->getName();
```
dantleech, line 73: reading the name from the node is actually very expensive and should be avoided.

lib/WorseReflection/Core/Reflection/Collection/ClassLikeReflectionMemberCollection.php
```diff
@@ -0,0 +1,338 @@
...
+
+
+        return $new;
+    }
+
+    private function filter(Closure $closure): self
+    {
+        $new = new self([]);
+        foreach (self::MEMBER_TYPES as $collection) {
+            $new->$collection = array_filter($this->$collection, $closure);
```
dantleech, line 331: thus and `map()` get called _alot_. wonder if it's faster to use a foreach loop to filter / map?

PR #1826, opened 2022-07-26
Gh 1771: Failing test

dantleech:


dantleech:
Fixed by removing something that didn't cause any other tests to fail, hopefully just a piece of junk code.

PR #1832, opened 2022-07-30
Fix for false-positive on imported namespaces ... missing attributes

ging-dev:


lib/WorseReflection/Tests/Integration/Bridge/TolerantParser/Diagonstics/UnusedImportProvider/UsedByDoctrineAttributeAliased.test
```diff
@@ -0,0 +1,8 @@
+<?php
+
+use Doctrine\Mapping as ORM;
+
+#[ORM\Id]
```
dantleech, line 5: attributes _should_ be detected in the AST not with regex

ging-dev: oh, maybe this is the real problem:
image

ging-dev: ready to merge

lib/WorseReflection/Bridge/TolerantParser/Diagnostics/UnusedImportProvider.php
```diff
@@ -81,8 +81,11 @@ public function exit(NodeContextResolver $resolver, Frame $frame, Node $node): i
         $contents = $node->getFileContents();
 
         foreach ($this->imported as $importedFqn => $imported) {
-            if (isset($this->names[$importedFqn])) {
-                continue;
+            foreach ($this->names as $name => $_) {
+                if (strpos($name, $importedFqn) === 0) {
```
dantleech, line 85: what if we had:

```
use Foo as ORMTheSecond;
use Bar as ORM;

/** @ORM\Entity */
```

`ORMTheSecond` would not trigger a diagnostic?

I think I've done this logic somewhere else, I'll check.

dantleech:
```
<?php

use Doctrine\Mapping;
use MappingBar;

#[Mapping\Id]
#[MappingBarFoo]
function fo($foo): void
{
}
```

fails, because `MappingBarFoo` starts with `MappingBar`

dantleech: not sure if the best way to do it but:

```
                if (strpos($name . '\\', $importedFqn . '\\') === 0) {
```
seems to fix

dantleech: i think this:

```
            foreach ($this->names as $name => $_) {
                if (strpos($name . '\\', $importedFqn . '\\') === 0) {
                    unset($this->names[$name]);
                    continue 2;
                }
            }
```

although it would be good to avoid the additional loop, maybe we could store the first segment of the names in a property so that we don't need to iterate over all the names again?

ging-dev: i tried not adding loop however it doesn't work on docblock =(
```php
<?php

use App\Entity;

/**
 * @param Entity\User $foo
 */
function fo($foo): void
{
}
```

dantleech (approved):

dantleech:
Thank s!

PR #1836, opened 2022-08-03
Support lazily resolving documentation

dantleech:
This PR adds support for LSP `completion/resolve` which allows us to calculate things like documentation only when the item is selected in the user interface.

This makes completion near instantaneous in some cases where there would otherwise be a significant delay.

### DIFF
--- a/lib/Completion/Core/Suggestion.php
+++ b/lib/Completion/Core/Suggestion.php
@@ -183,7 +183,10 @@ class Suggestion
         return $this->shortDescription;
     }
 
-    public function withShortDescription(string $description): self
+    /**
+     * @param string|Closure $description
+     */
+    public function withShortDescription($description): self
     {
         $clone = clone $this;
         $clone->shortDescription = $description;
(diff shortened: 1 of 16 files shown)


lib/Completion/Core/Suggestion.php
```diff
@@ -140,16 +154,12 @@ public function toArray(): array
             'name_import' => $this->nameImport,
             'range' => $this->range ? $this->range->toArray() : null,
 
-            // deprecated: in favour of short_description, to be removed
-            // after 0.10.0
-            'info' => $this->shortDescription(),
+            // removed
+            'info' => '',
```
dantleech, line 158: this has been deprecated for many years

lib/Extension/LanguageServerCompletion/Handler/CompletionHandler.php
```diff
@@ -96,21 +110,30 @@ public function completion(CompletionParams $params, CancellationToken $token):
...
+                    'textEdit' => $this->textEdit($suggestion, $insertText, $textDocument),
+                    'additionalTextEdits' => $textEdits,
+                    'insertTextFormat' => $insertTextFormat,
+                    'data' => $index,
+                ]);
+
+                $this->resolve[$index] = function (CompletionItem $item) use ($suggestion): CompletionItem {
+                    $documentation = $suggestion->documentation();
+                    $item->documentation = $documentation ? new MarkupContent(MarkupKind::MARKDOWN, $documentation) : null;
+                    $item->detail = $this->formatShortDescription($suggestion);
```
dantleech, line 127: This removes the "eager" details, in Nvim I use this to show the FQN of the classes - but a better way would be to include the FQN (or at least an indicative abbreviation) as the `label`

Issue #1526, opened 2022-05-02
Allow to show FQCN on duplicated CompletionItems


przepompownia:
For example
![image](https://user-images.githubusercontent.com/11404453/166335193-5f671679-ca34-4d86-bdca-77bddc0472bf.png)
It requires manual scrolling until the wanted item will be selected. 

Currently in `CompletionHandler` I see that for classes FQCN is "hidden" in `'detail' => $this->formatShortDescription($suggestion)` and no phpdoc is shown. 

After duplicating `detail` to `name`:
![image](https://user-images.githubusercontent.com/11404453/166339944-898d4ce8-95ee-43c5-afeb-15192f63fa67.png)

I looked what intelephense shows:
![image](https://user-images.githubusercontent.com/11404453/166340232-52ee0e73-ba25-46c2-93bb-0f26abee7543.png)
there FQCNs are shown only on duplications. It make sense.

dantleech:
that can make sense, however is this CoC? wuth nvim LSP I get:

![image](https://user-images.githubusercontent.com/530801/166409036-26c98994-2658-4446-abcd-4a07d878aa34.png)

IIRC with CoC I had to configure something..

przepompownia:
My examples come from nvim's built-in client with cmp-nvim-lsp. 

Does your example come from master version of phpactor?

I'll be far from computer for most of this day. Maybe I'll need to look at other details returned in completion item and the ways of displaying it by the client.

dantleech:
found it, it's a configuration for `cmp`:

https://github.com/dantleech/dotfiles/blob/master/.config/nvim/lsp.cmp.lua#L5

przepompownia:
Maybe it's not the best solution (it takes lot of columns) but enough useful assuming not modifying phpactor side at the moment so thanks also for this example!

dantleech:
it goes both ways though - if I see a `Request` and there is only one in the results, it would still be useful to know where it comes from, but maybe it can be a config option.

dantleech:
Closing as duplicate of #1314

PR #1838, opened 2022-08-04
Show namespace prefix in completion item detail

dantleech:
This PR introduces a change to show the first segment of the namespace in name suggestions, and to progressively show more of the namespace segments if it is ambiguous:

image

image


dantleech:
doesn't show abbreviations for parameter completion ...

przepompownia:
#1526

przepompownia:
It mostly works nicely
image
except some `Exception`s :wink:

PR #1854, opened 2022-08-20
Splitting the Indexer Jobs into stubs and project

mamazu:
But for some reason the tests fail


lib/Indexer/Extension/Command/IndexBuildCommand.php
```diff
@@ -73,38 +73,41 @@ private function buildIndex(OutputInterface $output, ?string $subPath = null): v
...
-        $job = $this->indexer->getJob($subPath);
-        $output->writeln('done');
-        $output->writeln('<info>Building index:</info>');
-        $output->write(PHP_EOL);
-
-        if ($job->size() === 0) {
-            $output->writeln('No files found');
-            return;
-        }
+        $jobs = $this->indexer->getJobs($subPath);
```
dantleech, line 76: Bit of a smell as sub path can only apply to a single index (i.e. the "project" index)

lib/Indexer/Model/Indexer.php
```diff
@@ -15,17 +16,33 @@ class Indexer
...
-        return new IndexJob(
+        yield new IndexJob(
+            $this->builder,
+            $this->stubIndexProvider->provideFileList($this->index, $subPath)
+        );
+
+        yield new IndexJob(
             $this->builder,
             $this->provider->provideFileList($this->index, $subPath)
         );
```
dantleech, line 48: ideally the Indexer should not have direct knowledge of these two sources, but maybe it's good enough for now.

dantleech (commented):
It looks like we are still using a single index, but just have two jobs?

I think each job needs to use a separately managed index, and we need to be able to "combine" both indexes at runtime...

Using a single index doesn't add any advantage:

- It doesn't decrease the amount of of disk space used
- It doesn't increase the time required to index a project

?

mamazu:
Yeah, that all makes sense. I will look into how to create a separate index for the stubs.

dantleech:
I'd need to look into it, it might require deeper refactorings or perhaps a completely different appraoch ....

PR #1869, opened 2022-09-11
Adding command to copy the current class name to clipboard

### LINKED ISSUE #922: [vim] Copy FQN to clipboard command
Not sure how often this would be useful.

mamazu:
Fixes #922

First try at vim script, so please tell me if I made some mistakes.

This command assumes that the clipboard register is the `"+` register. Maybe we should make that configurable.

### DIFF
--- a/ftplugin/php/commands.vim
+++ b/ftplugin/php/commands.vim
@@ -35,6 +35,10 @@ command! -buffer -nargs=0 PhpactorContextMenu call phpactor#ContextMenu()
 command! -buffer -nargs=0 PhpactorCopyFile call phpactor#CopyFile()
 
 ""
+" Copy the current class FQN (based on current filename) to the clipboard
+command! -buffer -nargs=0 PhpactorCopyClassName call phpactor#CopyFullClassName()
+
+""
 " Move the current file - updating the namespace and class name according to
 " the new file location and name
 command! -buffer -nargs=0 PhpactorMoveFile call phpactor#MoveFile()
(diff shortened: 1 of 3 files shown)


ftplugin/php/commands.vim
```diff
@@ -65,3 +65,7 @@ command! -buffer -nargs=0 PhpactorGenerateAccessors call phpactor#GenerateAccess
 ""
 " Automatically add any missing properties to a class
 command! -buffer -nargs=0 PhpactorTransform call phpactor#Transform()
```
dantleech, line 67: This text is used tl generate the docs, can be changed to : Copy the current class FQN (based on current filename)  to the clipboard

ftplugin/php/commands.vim
```diff
@@ -65,3 +65,7 @@ command! -buffer -nargs=0 PhpactorGenerateAccessors call phpactor#GenerateAccess
...
+
+""
+" Getting the full name of the class
```
dantleech, line 70: This text is used tl generate the docs, can be changed to : Copy the current class FQN (based on current filename) to the clipboard

dantleech (changes requested):
Looks good to me! Need to run "make docs"  (or make vimdoc) to regenerate the vim documentation, which needs to be committed.

mamazu:
I also moved the command up to sort them alphabetically.

dantleech:
Nice, thanks. Test is failure is an unrelated, known, flakey test.

PR #1878, opened 2022-09-18
Fixing missing `false` default in json schema generation

mamazu:
Before when a command had a default value of false, it would have been not visible in the json.

### Before:
```json
         "logging.fingers_crossed": {
             "description": null,
             "type": [
                 "boolean"
             ]
         },
```

### After
```json
         "logging.fingers_crossed": {
             "description": null,
             "type": [
                 "boolean"
             ],
             "default": false
         },
```

### DIFF
--- a/lib/Extension/Core/Model/JsonSchemaBuilder.php
+++ b/lib/Extension/Core/Model/JsonSchemaBuilder.php
@@ -56,7 +56,7 @@ class JsonSchemaBuilder
                 if ($definition->types()) {
                     $meta['type'] = $this->mapTypes($definition->types());
                 }
-                if ($definition->defaultValue()) {
+                if (null !== $definition->defaultValue()) {
                     $meta['default'] = $definition->defaultValue();
                 }
                 if ([] !== $definition->enum()) {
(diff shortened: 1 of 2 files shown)


lib/Extension/Core/Model/JsonSchemaBuilder.php
```diff
@@ -53,10 +53,12 @@ public function dump(): string
...
                 ];
-                if ($definition->types()) {
-                    $meta['type'] = $this->mapTypes($definition->types());
+
+                $mappedTypes = $this->mapTypes($definition->types());
+                if ($mappedTypes) {
+                    $meta['type'] = $mappedTypes;
                 }
-                if ($definition->defaultValue()) {
+                if ($definition->defaultValue() || in_array('boolean', $mappedTypes)) {
```
dantleech, line 61: sorry to miss this PR, should this rather be `null !== $definition->defaultValue()` ? tbh there is a design flaw here as `null` could also be a default value, but this would be OK for now.

mamazu: I also forgot about it but I added a test now for it as well. One day if all extensions are documented this check isn't needed anymore.

dantleech (approved):

dantleech:
cheers!

PR #1879, opened 2022-09-18
[WIP] Adding generation for decorate.

mamazu:
Just to check that I am doing it right.

### DIFF
--- a/lib/Extension/LanguageServerCodeTransform/CodeAction/GenerateDecoratorProvider.php
+++ b/lib/Extension/LanguageServerCodeTransform/CodeAction/GenerateDecoratorProvider.php
@@ -61,11 +61,11 @@ class GenerateDecoratorProvider implements CodeActionProvider
 
             return [
                 CodeAction::fromArray([
-                    'title' => sprintf('Generate "%s"', $interfaceFQN),
+                    'title' => sprintf('Decorate "%s"', $interfaceFQN),
                     'kind' => self::KIND,
                     'command' => new Command(
                         'Generate decorator',
-                        GenerateDecoratorCommand::class,
+                        GenerateDecoratorCommand::NAME,
                         [
                             $textDocument->uri,
                             $interfaceFQN,
--- a/lib/Extension/LanguageServerCodeTransform/CodeAction/GenerateDecoratorProvider.php
+++ b/lib/Extension/LanguageServerCodeTransform/CodeAction/GenerateDecoratorProvider.php
@@ -1,0 +1,78 @@
+<?php
+
+namespace Phpactor\Extension\LanguageServerCodeTransform\CodeAction;
+
+use Amp\CancellationToken;
+use Amp\Promise;
+use Phpactor\Extension\LanguageServerCodeTransform\LspCommand\GenerateDecoratorCommand;
+use Phpactor\LanguageServerProtocol\CodeAction;
+use Phpactor\LanguageServerProtocol\Command;
+use Phpactor\LanguageServerProtocol\Range;
+use Phpactor\LanguageServerProtocol\TextDocumentItem;
+use Phpactor\LanguageServer\Core\CodeAction\CodeActionProvider;
+use Phpactor\WorseReflection\Core\Reflection\ReflectionClass;
+use Phpactor\WorseReflection\Reflector;
+use function Amp\call;
+
+class GenerateDecoratorProvider implements CodeActionProvider
+{
+    public const KIND = 'quickfix.generate_decorator';
+
+    private Reflector $reflector;
+
+    public function __construct(Reflector $reflector)
+    {
+        $this->reflector = $reflector;
+    }
+
+    public function kinds(): array
+    {
+        return [
+            self::KIND,
+        ];
+    }
+
+    public function provideActionsFor(TextDocumentItem $textDocument, Range $range, CancellationToken $cancel): Promise
+    {
+        return call(function () use ($textDocument) {
+            $classes = $this->reflector->reflectClassesIn($textDocument->text);
+            if (count($classes) !== 1) {
+                return [];
+            }
+
+            $class = $classes->first();
+            assert($class instanceof ReflectionClass);
+
+            $interfaces = $class->interfaces();
+
+            if (count($interfaces) !== 1) {
+                return [];
+            }
+
+            if (count($class->methods()) > 0) {
+                return [];
+            }
+
+            if ($class->parent()) {
+                return [];
+            }
+
+            $interfaceFQN = (string) $interfaces->first()->type();
+
+            return [
+                CodeAction::fromArray([
+                    'title' => sprintf('Generate "%s"', $interfaceFQN),
+                    'kind' => self::KIND,
+                    'command' => new Command(
+                        'Generate decorator',
+                        GenerateDecoratorCommand::class,
+                        [
+                            $textDocument->uri,
+                            $interfaceFQN,
+                        ]
+                    )
+                ]),
+            ];
+        });
+    }
+}
--- a/lib/Extension/LanguageServerCodeTransform/LanguageServerCodeTransformExtension.php
+++ b/lib/Extension/LanguageServerCodeTransform/LanguageServerCodeTransformExtension.php
@@ -187,6 +187,7 @@ class LanguageServerCodeTransformExtension implements Extension
         $container->register(GenerateDecoratorCommand::class, function (Container $container) {
             return new GenerateDecoratorCommand(
                 $container->get(ClientApi::class),
+                $container->get(LanguageServerExtension::SERVICE_SESSION_WORKSPACE),
                 $container->get(GenerateDecorator::class)
             );
         }, [
--- a/lib/Extension/LanguageServerCodeTransform/LanguageServerCodeTransformExtension.php
+++ b/lib/Extension/LanguageServerCodeTransform/LanguageServerCodeTransformExtension.php
@@ -25,6 +25,7 @@ use Phpactor\Extension\LanguageServerCodeTransform\CodeAction\ExtractMethodProvi
 use Phpactor\Extension\LanguageServerCodeTransform\CodeAction\FillObjectProvider;
 use Phpactor\Extension\LanguageServerCodeTransform\CodeAction\GenerateAccessorsProvider;
 use Phpactor\Extension\LanguageServerCodeTransform\CodeAction\GenerateConstructorProvider;
+use Phpactor\Extension\LanguageServerCodeTransform\CodeAction\GenerateDecoratorProvider;
 use Phpactor\Extension\LanguageServerCodeTransform\CodeAction\GenerateMethodProvider;
 use Phpactor\Extension\LanguageServerCodeTransform\CodeAction\ImportNameProvider;
 use Phpactor\Extension\LanguageServerCodeTransform\CodeAction\TransformerCodeActionPovider;
@@ -352,6 +353,14 @@ class LanguageServerCodeTransformExtension implements Extension
             return new GenerateConstructorProvider(
                 $container->get(GenerateConstructor::class),
                 $container->get(WorkspaceEditConverter::class),
+            );
+        }, [
+            LanguageServerExtension::TAG_CODE_ACTION_PROVIDER => []
+        ]);
+
+        $container->register(GenerateDecoratorProvider::class, function (Container $container) {
+            return new GenerateDecoratorProvider(
+                $container->get(WorseReflectionExtension::SERVICE_REFLECTOR)
             );
         }, [
             LanguageServerExtension::TAG_CODE_ACTION_PROVIDER => []
--- a/lib/Extension/LanguageServerCodeTransform/LanguageServerCodeTransformExtension.php
+++ b/lib/Extension/LanguageServerCodeTransform/LanguageServerCodeTransformExtension.php
@@ -9,6 +9,7 @@ use Phpactor\CodeTransform\Domain\Refactor\ExtractMethod;
 use Phpactor\CodeTransform\Domain\Refactor\FillObject;
 use Phpactor\CodeTransform\Domain\Refactor\GenerateAccessor;
 use Phpactor\CodeTransform\Domain\Refactor\GenerateConstructor;
+use Phpactor\CodeTransform\Domain\Refactor\GenerateDecorator;
 use Phpactor\CodeTransform\Domain\Refactor\GenerateMethod;
 use Phpactor\CodeTransform\Domain\Refactor\ImportName;
 use Phpactor\Container\Container;
@@ -34,6 +35,7 @@ use Phpactor\Extension\LanguageServerCodeTransform\LspCommand\ExtractConstantCom
 use Phpactor\Extension\LanguageServerCodeTransform\LspCommand\ExtractExpressionCommand;
 use Phpactor\Extension\LanguageServerCodeTransform\LspCommand\ExtractMethodCommand;
 use Phpactor\Extension\LanguageServerCodeTransform\LspCommand\GenerateAccessorsCommand;
+use Phpactor\Extension\LanguageServerCodeTransform\LspCommand\GenerateDecoratorCommand;
 use Phpactor\Extension\LanguageServerCodeTransform\LspCommand\GenerateMethodCommand;
 use Phpactor\Extension\LanguageServerCodeTransform\LspCommand\ImportAllUnresolvedNamesCommand;
 use Phpactor\Extension\LanguageServerCodeTransform\LspCommand\ImportNameCommand;
@@ -179,6 +181,17 @@ class LanguageServerCodeTransformExtension implements Extension
         }, [
             LanguageServerExtension::TAG_COMMAND => [
                 'name' => ExtractExpressionCommand::NAME
+            ],
+        ]);
+
+        $container->register(GenerateDecoratorCommand::class, function (Container $container) {
+            return new GenerateDecoratorCommand(
+                $container->get(ClientApi::class),
+                $container->get(GenerateDecorator::class)
+            );
+        }, [
+            LanguageServerExtension::TAG_COMMAND => [
+                'name' => GenerateDecoratorCommand::NAME
             ],
         ]);
     }
--- a/lib/Extension/LanguageServerCodeTransform/LspCommand/GenerateDecoratorCommand.php
+++ b/lib/Extension/LanguageServerCodeTransform/LspCommand/GenerateDecoratorCommand.php
@@ -1,0 +1,120 @@
+<?php
+
+namespace Phpactor\Extension\LanguageServerCodeTransform\LspCommand;
+
+use Amp\Promise;
+use Phpactor\CodeBuilder\Domain\Builder\SourceCodeBuilder;
+use Phpactor\Extension\LanguageServerBridge\Converter\TextEditConverter;
+use Phpactor\LanguageServerProtocol\Command;
+use Phpactor\LanguageServerProtocol\WorkspaceEdit;
+use Phpactor\LanguageServer\Core\Server\ClientApi;
+use Phpactor\CodeTransform\Domain\SourceCode;
+use Phpactor\LanguageServer\Core\Workspace\Workspace;
+use Phpactor\TextDocument\TextEdits;
+use Phpactor\WorseReflection\Reflector;
+use Phpactor\CodeBuilder\Domain\Builder\MethodBuilder;
+use Phpactor\WorseReflection\Core\Reflection\ReflectionMethod;
+use Phpactor\CodeBuilder\Domain\Prototype\Visibility;
+
+class GenerateDecoratorCommand extends Command
+{
+    public const NAME = 'generate_decorator';
+
+    private ClientApi $clientApi;
+
+    private Workspace $workspace;
+
+    private Reflector $reflector;
+
+    public function __construct(
+        ClientApi $clientApi,
+        Workspace $workspace,
+        Reflector $reflector
+    ) {
+        $this->clientApi = $clientApi;
+        $this->workspace = $workspace;
+        $this->reflector = $reflector;
+    }
+
+    /**
+     * @return Promise<ApplyWorkspaceEditResponse>
+     */
+    public function __invoke(string $uri, string $interfaceFQN): Promise
+    {
+        $textDocument = $this->workspace->get($uri);
+        $source = SourceCode::fromStringAndPath($textDocument->text, $textDocument->uri);
+
+        $textEdits = $this->getTextEdits($source, $interfaceFQN);
+
+        return $this->clientApi->workspace()->applyEdit(new WorkspaceEdit([
+            $uri => TextEditConverter::toLspTextEdits($textEdits, $textDocument->text)
+        ]), 'Generate decoration');
+    }
+
+    private function getTextEdits(SourceCode $source, string $interfaceFQN): TextEdits
+    {
+        $class = $this->reflector->reflectClassesIn($source)->classes()->first();
+
+        $builder = SourceCodeBuilder::create();
+        $builder->namespace($class->name()->namespace());
+        $classBuilder = $builder->class($class->name()->short());
+
+        $classBuilder->property('inner')->visibility(Visibility::PRIVATE)->type($interfaceFQN);
+
+        $constructor = $classBuilder->method('__construct');
+        $constructor->parameter('inner')->type($interfaceFQN);
+        $constructor->body()->line('$this->inner = $inner');
+
+        $interface = $this->reflector->reflectInterface($interfaceFQN);
+        foreach ($interface->methods() as $interfaceMethod) {
+            $method = $classBuilder->method($interfaceMethod->name());
+
+            $method->returnType($interfaceMethod->returnType());
+            $method->visibility($interfaceMethod->visibility());
+
+            $this->attachParameters($method, $interfaceMethod);
+
+            $method->body()->line($this->generateMethodBody($interfaceMethod));
+        }
+
+        return [];
+    }
+
+    /**
+     * Copying over the method parameters from the interface to the decoration
+     */
+    private function attachParameters(MethodBuilder $method, ReflectionMethod $interfaceMethod): void
+    {
+        foreach ($interfaceMethod->parameters() as $interfaceMethodParameter) {
+            $method->parameter($interfaceMethodParameter->name())
+                   ->type($interfaceMethodParameter->type())
+                   ->defaultValue($interfaceMethodParameter->default())
+                ;
+        }
+    }
+
+    /**
+     * This method creates the method body which means copying parameters of the interface method to the body of the function.
+     * So if the interface contains:
+     *
+     * function someFunction(string $a, int $b)
+     *
+     * then the content of the decoration method needs to be
+     *
+     * $this->inner->someFunction($a, $b);
+     */
+    private function generateMethodBody(ReflectionMethod $interfaceMethod): string
+    {
+        $code = '$this->inner->'.$interfaceMethod->name().'(';
+        foreach ($interfaceMethod->parameters() as $interfaceMethodParameter) {
+            $code .= $interfaceMethodParameter->name().','.PHP_EOL;
+        }
+        $code .= ')';
+
+        if (!$interfaceMethod->returnType()->isVoid()) {
+            $code = 'return '. $code;
+        }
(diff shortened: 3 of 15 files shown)


lib/Extension/LanguageServerCodeTransform/CodeAction/GenerateDecoratorProvider.php
```diff
@@ -0,0 +1,46 @@
...
+    {
+        return [
+            self::KIND,
+        ];
+    }
+
+    public function provideActionsFor(TextDocumentItem $textDocument, Range $range, CancellationToken $cancel): Promise
+    {
+        return call(function () use ($textDocument) {
+            $this->generateDecoration->transform($textDocument);
```
mamazu, line 29: Don't mind this line, it's not working.

If we just take the first interface then this could be gone. Otherwise, here would be the place where we extract the interfaces and make them selectable via code actions.

lib/Extension/LanguageServerCodeTransform/LspCommand/GenerateDecoratorCommand.php
```diff
@@ -0,0 +1,50 @@
...
+     * @return Promise<ApplyWorkspaceEditResponse>
+     */
+    public function __invoke(string $uri): Promise
+    {
+        $textDocument = $this->workspace->get($uri);
+
+        try {
+            $textEdits = $this->decorateClass->transform(
+                SourceCode::fromStringAndPath($textDocument->text, $textDocument->uri)
+            );
```
mamazu, line 35: Here would be the place where to implement the "generation logic"

lib/Extension/LanguageServerCodeTransform/CodeAction/GenerateDecoratorProvider.php
dantleech: yeah, here we could have:

```php
$classes = $this->reflector->reflectClassesIn($textDocument->text);
if (count($classes) !== 1) {
    return [];
}
$class = $classes->first();
$interfaces = $class->interfaces();

if (count($interfaces) !== 1) {
    return [];
}

$methods = $class->methods();

if (count($methods) > 0) {
    return [];
}

if ($class->parent()) {
    return [];
}

// otherwise return code action
```

so in order to show code action:

- file must have one class
- class must have exactly one interface
- class must not extend anything
- class must not have any methods
- ...

can improve that as required, but I think it should be ok?

lib/Extension/LanguageServerCodeTransform/CodeAction/GenerateDecoratorProvider.php
```diff
@@ -0,0 +1,46 @@
...
+
+            return [
+                CodeAction::fromArray([
+                    'title' => 'Generate decorator',
```
dantleech, line 33: You could have `Decorate "<interface name here>"` instead of `Generate decorator

lib/Extension/LanguageServerCodeTransform/LanguageServerCodeTransformExtension.php
```diff
@@ -356,5 +357,11 @@ private function registerCodeActions(ContainerBuilder $container): void
         }, [
             LanguageServerExtension::TAG_CODE_ACTION_PROVIDER => []
         ]);
+
+        $container->register(GenerateDecoratorProvider::class, function (Container $container) {
+            return new GenerateDecoratorProvider();
```
dantleech, line 362: can inject reflector with `$container->get(WorseReflectionExtension::SERVICE_REFLECTOR)`

lib/Extension/LanguageServerCodeTransform/LspCommand/GenerateDecoratorCommand.php
```diff
@@ -0,0 +1,50 @@
...
+
+    public function __construct(ClientApi $clientApi)
+    {
+        $this->clientApi = $clientApi;
+    }
+
+    /**
+     * @return Promise<ApplyWorkspaceEditResponse>
+     */
+    public function __invoke(string $uri): Promise
```
dantleech, line 28: I would possibly have the command accept an interface name `$interfaceFqn` - that way it's responsiblity may be reduced and we have a clearer distinction between the code action provider and the command.

lib/Extension/LanguageServerCodeTransform/LspCommand/GenerateDecoratorCommand.php
```diff
@@ -0,0 +1,50 @@
...
+    {
+        $textDocument = $this->workspace->get($uri);
+
+        try {
+            $textEdits = $this->decorateClass->transform(
```
dantleech, line 33: the decorate class is **not** a transformer (in the Phpactor sense).

it should accept the class name and interface name in addition to the text document.

dantleech:
cool, added some comments :+1:

lib/Extension/LanguageServerCodeTransform/LspCommand/GenerateDecoratorCommand.php
mamazu: Done that.

lib/Extension/LanguageServerCodeTransform/LanguageServerCodeTransformExtension.php
mamazu: Yep, done.

lib/Extension/LanguageServerCodeTransform/LspCommand/GenerateDecoratorCommand.php
mamazu: Yeah, I have replaced this part with the implementation of the command itself.

lib/Extension/LanguageServerCodeTransform/CodeAction/GenerateDecoratorProvider.php
mamazu: Changed that as well.

lib/Extension/LanguageServerCodeTransform/CodeAction/GenerateDecoratorProvider.php
mamazu: Looks great, I have copied that over.

lib/Extension/LanguageServerCodeTransform/LspCommand/GenerateDecoratorCommand.php
```diff
@@ -0,0 +1,120 @@
...
+
+            $method->returnType($interfaceMethod->returnType());
+            $method->visibility($interfaceMethod->visibility());
+
+            $this->attachParameters($method, $interfaceMethod);
+
+            $method->body()->line($this->generateMethodBody($interfaceMethod));
+        }
+
+        return [];
```
mamazu, line 80: I think I have done most of it but there is one question. I have generates all the methods and the new property inside the decorator class but how to I extract a `TextEdits` out of them?

dantleech: If using the "code builder" thing:

```
        return $this->updater->textEditsFor($sourceCodeBuilder->build(), Code::fromString((string) $source));
```

to transform to _phpactor_ text edits. then need to convert to LSP edits in the code action provider (the following code is in `.../LspCommand/TransformCommand` FWIW)

```
        return $this->clientApi->workspace()->applyEdit(new WorkspaceEdit([
            $uri => TextEditConverter::toLspTextEdits($textEdits, $textDocument->text)
        ]), 'Apply source code transformation');
````

Comment (mamazu) [reply lib/Extension/LanguageServerCodeTransform/LspCommand/GenerateDecoratorCommand.php:80]: Well then I should be done.

**Comment by mamazu** [comment]
So this should work. Then the next topic would be writing tests?

File: lib/Extension/LanguageServerCodeTransform/LspCommand/GenerateDecoratorCommand.php [hunk 8]
```diff
@@ -0,0 +1,126 @@
...
+
+            $method->returnType($interfaceMethod->returnType());
+            $method->visibility($interfaceMethod->visibility());
+
+            $this->attachParameters($method, $interfaceMethod);
+
+            $method->body()->line($this->generateMethodBody($interfaceMethod));
+        }
+
+        return $this->updater->textEditsFor($builder->build(), Code::fromString((string) $source));
```
dantleech, line 86: the "logic" shouldn't be in this class (think of it like an MVC controller), it should be in a plain PHP class in CodeTransform such as:

https://github.com/phpactor/phpactor/blob/9f20e64fcc828851157fc4e1896a1278b73f3f97/lib/CodeTransform/Adapter/WorseReflection/Refactor/WorseExtractConstant.php#L24

(you could add an interface as with that example too)

the class doesn't have to conform to any specific specification, just has to do this job (i.e. it returns the `TestEdits` collection)

you can delegate to that class from here,

dantleech: for testing _the logic_ you can look at any of the examples in CodeTransform, e.g. for the above:

https://github.com/phpactor/phpactor/blob/c102bef07614c48a53fb552a1fd56663b5a1949d/lib/CodeTransform/Tests/Adapter/WorseReflection/Refactor/WorseExtractConstantTest.php

mamazu:
Building the tests was super easy, and of course I found some bugs in my code.

In the generated code the types of the property are only done by annotation like this:
```php
/**
* @var Phpactor/Something
*/
private $inner;
```

Doesn't this project use a php version where this can be put in as a proper type hint.
```php
private Phpactor/Something $inner;
```

But otherwise I think I am done with the feature.

lib/Extension/LanguageServerCodeTransform/LspCommand/GenerateDecoratorCommand.php
```diff
@@ -0,0 +1,43 @@
...
+
+use Amp\Promise;
+use Phpactor\CodeTransform\Domain\Refactor\GenerateDecorator;
+use Phpactor\Extension\LanguageServerBridge\Converter\TextEditConverter;
+use Phpactor\LanguageServerProtocol\Command;
+use Phpactor\LanguageServerProtocol\WorkspaceEdit;
+use Phpactor\LanguageServer\Core\Server\ClientApi;
+use Phpactor\CodeTransform\Domain\SourceCode;
+
+class GenerateDecoratorCommand extends Command
```
dantleech, line 13: this command is not registered:

```
Exception [RuntimeException] Command "Phpactor\Extension\LanguageServerCodeTransform\LspComman
d\GenerateDecoratorCommand" not found
```

dantleech:
The implementation needs to implement the interface:

```diff
diff --git a/lib/CodeTransform/Adapter/WorseReflection/Refactor/WorseGenerateDecorator.php b/lib/CodeTransform/Adapter/WorseReflection/Refactor/WorseGenerateDecorator.php
index 01ff2c85..705f8a2b 100644
--- a/lib/CodeTransform/Adapter/WorseReflection/Refactor/WorseGenerateDecorator.php
+++ b/lib/CodeTransform/Adapter/WorseReflection/Refactor/WorseGenerateDecorator.php
@@ -5,6 +5,7 @@ namespace Phpactor\CodeTransform\Adapter\WorseReflection\Refactor;
 use Phpactor\CodeBuilder\Domain\Builder\SourceCodeBuilder;
 use Phpactor\CodeBuilder\Domain\Code;
 use Phpactor\CodeBuilder\Domain\Updater;
+use Phpactor\CodeTransform\Domain\Refactor\GenerateDecorator;
 use Phpactor\CodeTransform\Domain\SourceCode;
 use Phpactor\TextDocument\TextEdits;
 use Phpactor\CodeBuilder\Domain\Builder\MethodBuilder;
@@ -13,7 +14,7 @@ use Phpactor\CodeBuilder\Domain\Prototype\Visibility;
 use Phpactor\WorseReflection\Reflector;
 use Phpactor\WorseReflection\Core\Type\MissingType;
 
-class WorseGenerateDecorator
+class WorseGenerateDecorator implements GenerateDecorator
 {
     private Reflector $reflector;
```

The provider needs to reference the command name:
```
diff --git a/lib/Extension/LanguageServerCodeTransform/CodeAction/GenerateDecoratorProvider.php b/lib/Extension/LanguageServerCodeTransform/CodeAction/GenerateDecoratorProvider.php
index f444bf58..443d8876 100644
--- a/lib/Extension/LanguageServerCodeTransform/CodeAction/GenerateDecoratorProvider.php
+++ b/lib/Extension/LanguageServerCodeTransform/CodeAction/GenerateDecoratorProvider.php
@@ -61,11 +61,11 @@ class GenerateDecoratorProvider implements CodeActionProvider
 
             return [
                 CodeAction::fromArray([
-                    'title' => sprintf('Generate "%s"', $interfaceFQN),
+                    'title' => sprintf('Decorate "%s"', $interfaceFQN),
                     'kind' => self::KIND,
                     'command' => new Command(
                         'Generate decorator',
-                        GenerateDecoratorCommand::class,
+                        GenerateDecoratorCommand::NAME,
                         [
                             $textDocument->uri,
                             $interfaceFQN,
```

The command needs to be registered (against it's name):

```diff
diff --git a/lib/Extension/LanguageServerCodeTransform/LanguageServerCodeTransformExtension.php b/lib/Extension/LanguageServerCodeTransform/LanguageServerCodeTransformExtension.php
index 6e1f1ede..0a13f049 100644
--- a/lib/Extension/LanguageServerCodeTransform/LanguageServerCodeTransformExtension.php
+++ b/lib/Extension/LanguageServerCodeTransform/LanguageServerCodeTransformExtension.php
@@ -9,6 +9,7 @@ use Phpactor\CodeTransform\Domain\Refactor\ExtractMethod;
 use Phpactor\CodeTransform\Domain\Refactor\FillObject;
 use Phpactor\CodeTransform\Domain\Refactor\GenerateAccessor;
 use Phpactor\CodeTransform\Domain\Refactor\GenerateConstructor;
+use Phpactor\CodeTransform\Domain\Refactor\GenerateDecorator;
 use Phpactor\CodeTransform\Domain\Refactor\GenerateMethod;
 use Phpactor\CodeTransform\Domain\Refactor\ImportName;
 use Phpactor\Container\Container;
@@ -34,6 +35,7 @@ use Phpactor\Extension\LanguageServerCodeTransform\LspCommand\ExtractConstantCom
 use Phpactor\Extension\LanguageServerCodeTransform\LspCommand\ExtractExpressionCommand;
 use Phpactor\Extension\LanguageServerCodeTransform\LspCommand\ExtractMethodCommand;
 use Phpactor\Extension\LanguageServerCodeTransform\LspCommand\GenerateAccessorsCommand;
+use Phpactor\Extension\LanguageServerCodeTransform\LspCommand\GenerateDecoratorCommand;
 use Phpactor\Extension\LanguageServerCodeTransform\LspCommand\GenerateMethodCommand;
 use Phpactor\Extension\LanguageServerCodeTransform\LspCommand\ImportAllUnresolvedNamesCommand;
 use Phpactor\Extension\LanguageServerCodeTransform\LspCommand\ImportNameCommand;
@@ -181,6 +183,17 @@ class LanguageServerCodeTransformExtension implements Extension
                 'name' => ExtractExpressionCommand::NAME
             ],
         ]);
+
+        $container->register(GenerateDecoratorCommand::class, function (Container $container) {
+            return new GenerateDecoratorCommand(
+                $container->get(ClientApi::class),
+                $container->get(GenerateDecorator::class)
+            );
+        }, [
+            LanguageServerExtension::TAG_COMMAND => [
+                'name' => GenerateDecoratorCommand::NAME
+            ],
+        ]);
     }
 
     private function registerCodeActions(ContainerBuilder $container): void

```

The property is wrongly named:

```diff
diff --git a/lib/Extension/LanguageServerCodeTransform/LspCommand/GenerateDecoratorCommand.php b/lib/Extension/LanguageServerCodeTransform/LspCommand/GenerateDecoratorCommand.php
index 961bc271..f456cb76 100644
--- a/lib/Extension/LanguageServerCodeTransform/LspCommand/GenerateDecoratorCommand.php
+++ b/lib/Extension/LanguageServerCodeTransform/LspCommand/GenerateDecoratorCommand.php
@@ -5,25 +5,25 @@ namespace Phpactor\Extension\LanguageServerCodeTransform\LspCommand;
 use Amp\Promise;
 use Phpactor\CodeTransform\Domain\Refactor\GenerateDecorator;
 use Phpactor\Extension\LanguageServerBridge\Converter\TextEditConverter;
-use Phpactor\LanguageServerProtocol\Command;
+use Phpactor\LanguageServer\Core\Command\Command;
 use Phpactor\LanguageServerProtocol\WorkspaceEdit;
 use Phpactor\LanguageServer\Core\Server\ClientApi;
 use Phpactor\CodeTransform\Domain\SourceCode;
 
-class GenerateDecoratorCommand extends Command
+class GenerateDecoratorCommand implements Command
 {
     public const NAME = 'generate_decorator';
 
     private ClientApi $clientApi;
 
-    private GenerateDecorator $generateDecoration;
+    private GenerateDecorator $generateDecorator;
 
     public function __construct(
         ClientApi $clientApi,
         GenerateDecorator $generateDecorator
     ) {
         $this->clientApi = $clientApi;
-        $this->generateDecoration = $generateDecoration;
+        $this->generateDecorator = $generateDecorator;
     }
 
     /**
@@ -34,7 +34,7 @@ class GenerateDecoratorCommand extends Command
         $textDocument = $this->workspace->get($uri);
         $source = SourceCode::fromStringAndPath($textDocument->text, $textDocument->uri);
 
-        $textEdits = $this->generateDecoration->getTextEdits($source, $interfaceFQN);
+        $textEdits = $this->generateDecorator->getTextEdits($source, $interfaceFQN);
 
         return $this->clientApi->workspace()->applyEdit(new WorkspaceEdit([
             $uri => TextEditConverter::toLspTextEdits($textEdits, $textDocument->text)
```

there are further errors when trying to use the feature (e.g. provider uses the `workspace` but it's not injected).

In general there should be a test for the provider e.g.: https://github.com/mamazu/phpactor/blob/generate_decorator/lib/Extension/LanguageServerCodeTransform/Tests/Unit/CodeAction/CreateClassProviderTest.php

dantleech:
For the docblock, if using > PHP 7.4 it should only render the docblock if it's different to the AST type:

https://github.com/mamazu/phpactor/blob/42aba516ae71836e25edd92eb1730c454e000e07/templates/code/7.4/Property.php.twig#L1-L11

but if there's a bug there then not related to this PR.

mamazu:
Okay I have implemented the changes, and as for the test. I added it and also changed it to fit the scenario that should be tested. However, it fails for some reason.

dantleech:
Testing it out manually I can see that no code action is reported:

image

so the tests seem to be correct?

dantleech:
forget that last comment, try actually using the feature in vim:
image

i'm having a look now

dantleech:
For the test Phpactor is (wronlgy I think) not returning non-existing interfaces, but in this case they are necessary to exist and have at least one method:

```diff
diff --git a/lib/Extension/LanguageServerCodeTransform/Tests/Unit/CodeAction/GenerateDecoratorProviderTest.php b/lib/Extension/LanguageServerCodeTransform/Tests/Unit/CodeAction/GenerateDecoratorProviderTest.php
index def5cd38..ad4aae0d 100644
--- a/lib/Extension/LanguageServerCodeTransform/Tests/Unit/CodeAction/GenerateDecoratorProviderTest.php
+++ b/lib/Extension/LanguageServerCodeTransform/Tests/Unit/CodeAction/GenerateDecoratorProviderTest.php
@@ -20,10 +20,10 @@ class GenerateDecoratorProviderTest extends IntegrationTestCase
      * @dataProvider provideClassCreateProvider
      * @group flakey
      */
-    public function testClassCreateProvider(string $manifest, int $expectedCount): void
+    public function testClassCreateProvider(string $source, int $expectedCount): void
     {
         $this->workspace()->reset();
-        $this->workspace()->loadManifest($manifest);
+        [$source, $offset] = ExtractOffset::fromSource($source);
 
         $tester = $this->container([])->get(LanguageServerBuilder::class)->tester(
             ProtocolFactory::initializeParams($this->workspace()->path())
@@ -31,15 +31,12 @@ class GenerateDecoratorProviderTest extends IntegrationTestCase
         $tester->initialize();
         assert($tester instanceof LanguageServerTester);
 
-        $FooBar = $this->workspace()->getContents('FooBar.php');
-        [ $source, $offset ] = ExtractOffset::fromSource($FooBar);
-
         $tester->textDocument()->open('file:///foobar', $source);
 
         $result = $tester->requestAndWait(CodeActionRequest::METHOD, new CodeActionParams(
             ProtocolFactory::textDocumentIdentifier('file:///foobar'),
             new Range(
-                ProtocolFactory::position(0, 8),
+                PositionConverter::intByteOffsetToPosition((int)$offset, $source),
                 PositionConverter::intByteOffsetToPosition((int)$offset, $source)
             ),
             new CodeActionContext([])
@@ -47,7 +44,7 @@ class GenerateDecoratorProviderTest extends IntegrationTestCase
 
         $tester->assertSuccess($result);
 
-        $tester->textDocument()->save('file:///foobar', $source);
+        $tester->textDocument()->save('file:///foobar');
 
         $result = $tester->requestAndWait(CodeActionRequest::METHOD, new CodeActionParams(
             ProtocolFactory::textDocumentIdentifier('file:///foobar'),
@@ -70,9 +67,8 @@ class GenerateDecoratorProviderTest extends IntegrationTestCase
     {
         yield 'class with no interfaces' => [
             <<<'EOT'
-                // File: FooBar.php
                 <?php
-                class FooBar {}
+                class Foo<>bar {}
 
                 EOT
         , 0
@@ -80,9 +76,9 @@ class GenerateDecoratorProviderTest extends IntegrationTestCase
 
         yield 'class with one interface' => [
             <<<'EOT'
-                // File: FooBar.php
                 <?php
-                class FooBar implements SomeInterface {}
+                interface SomeInterface {public function foo(): void {}}
+                class FooBar implements So<>meInterface {}
 
                 EOT
         , 1
@@ -90,12 +86,13 @@ class GenerateDecoratorProviderTest extends IntegrationTestCase
 
         yield 'class with multiple interfaces' => [
             <<<'EOT'
-                // File: FooBar.php
                 <?php
-                class FooBar implements SomeInterface, OtherInterface {}
+                interface SomeInterface {public function foo(): void {}}
+                interface OtherInterface {public function foo(): void {}}
+                class FooBar implements SomeInterface, Ot<>herInterface {}
 
                 EOT
-        , 2
+        , 1
         ];
     }
 }
```

for the other thing, the Command needs the `$workspace` ideally that should be tested too, see the `Tests/Unit/LspCommand/*Test.php` directory for examples.

mamazu:
> For the docblock, if using > PHP 7.4 it should only render the docblock if it's different to the AST type:
> 
> https://github.com/mamazu/phpactor/blob/42aba516ae71836e25edd92eb1730c454e000e07/templates/code/7.4/Property.php.twig#L1-L11
> 
> there seems to be a bug there, as it should use `docTypeOrType` probably. regardless, you can call `->type($type)->docType($type)` in your code and it will work as expected.

If I add this to the property it still only generates the doc block but no typehint on the property.

```diff
lib/CodeTransform/Adapter/WorseReflection/Refactor/WorseGenerateDecorator.php

- $classBuilder->property('inner')->visibility(Visibility::PRIVATE)->type($interfaceFQN);
+ $classBuilder->property('inner')->visibility(Visibility::PRIVATE)->type($interfaceFQN)->docType($interfaceFQN);
```

mamazu:
I applied your changes.

I think I figured out what you meant and also added a test for the command.

dantleech:
almost works as expected:

image

are you able to test it locally?

mamazu:
Strange. The tests locally also run on php 8.1 and the tests don't fail with the var annotation.

dantleech:
There are a few things we needed to do:

- Convert the types to the "local" type (i.e. the imported name of the type as it couild have been aliased). This also is the "short" version.
- Add a class import for any types that are not already imported

The boiler plate around this makes me want to create a factory, but for now:

```diff
diff --git a/lib/CodeTransform/Adapter/WorseReflection/Refactor/WorseGenerateDecorator.php b/lib/CodeTransform/Adapter/WorseReflection/Refactor/WorseGenerateDecorator.php
index 705f8a2b..4fe1d441 100644
--- a/lib/CodeTransform/Adapter/WorseReflection/Refactor/WorseGenerateDecorator.php
+++ b/lib/CodeTransform/Adapter/WorseReflection/Refactor/WorseGenerateDecorator.php
@@ -9,8 +9,10 @@ use Phpactor\CodeTransform\Domain\Refactor\GenerateDecorator;
 use Phpactor\CodeTransform\Domain\SourceCode;
 use Phpactor\TextDocument\TextEdits;
 use Phpactor\CodeBuilder\Domain\Builder\MethodBuilder;
+use Phpactor\WorseReflection\Core\Reflection\ReflectionClassLike;
 use Phpactor\WorseReflection\Core\Reflection\ReflectionMethod;
 use Phpactor\CodeBuilder\Domain\Prototype\Visibility;
+use Phpactor\WorseReflection\Core\TypeFactory;
 use Phpactor\WorseReflection\Reflector;
 use Phpactor\WorseReflection\Core\Type\MissingType;
 
@@ -36,22 +38,32 @@ class WorseGenerateDecorator implements GenerateDecorator
         $builder->namespace($class->name()->namespace());
         $classBuilder = $builder->class($class->name()->short());
 
-        $classBuilder->property('inner')->visibility(Visibility::PRIVATE)->type($interfaceFQN);
+
+        $interfaceType = TypeFactory::class($interfaceFQN);
+        $interfaceType = $interfaceType->toLocalType($class->scope());
+
+        $property = $classBuilder->property('inner')
+            ->visibility(Visibility::PRIVATE)
+            ->type($interfaceType->toPhpString())
+            ->docType($interfaceType->__toString());
 
         $constructor = $classBuilder->method('__construct');
-        $constructor->parameter('inner')->type($interfaceFQN);
+        $constructor->parameter('inner')->type($interfaceType->toPhpString());
         $constructor->body()->line('$this->inner = $inner;');
 
         $interface = $this->reflector->reflectInterface($interfaceFQN);
         foreach ($interface->methods() as $interfaceMethod) {
             $method = $classBuilder->method($interfaceMethod->name());
 
-            if (!($interfaceMethod->returnType() instanceof MissingType)) {
-                $method->returnType($interfaceMethod->returnType());
+            if ($interfaceMethod->returnType()->isDefined()) {
+                $method->returnType($interfaceMethod->returnType()->toLocalType($class->scope())->toPhpString());
+                foreach ($interfaceMethod->returnType()->classNamedTypes() as $type) {
+                    $builder->use($type->name());
+                }
             }
             $method->visibility($interfaceMethod->visibility());
 
-            $this->attachParameters($method, $interfaceMethod);
+            $this->attachParameters($builder, $class, $method, $interfaceMethod);
 
             $method->body()->line($this->generateMethodBody($interfaceMethod));
         }
@@ -62,13 +74,18 @@ class WorseGenerateDecorator implements GenerateDecorator
     /**
      * Copying over the method parameters from the interface to the decoration
      */
-    private function attachParameters(MethodBuilder $method, ReflectionMethod $interfaceMethod): void
+    private function attachParameters(SourceCodeBuilder $builder, ReflectionClassLike $class, MethodBuilder $method, ReflectionMethod $interfaceMethod): void
     {
         foreach ($interfaceMethod->parameters() as $interfaceMethodParameter) {
             $parameter = $method->parameter($interfaceMethodParameter->name())
-                                ->type($interfaceMethodParameter->type());
+                ->type($interfaceMethodParameter->type()->toLocalType($class->scope())->toPhpString());
+
+            foreach ($interfaceMethodParameter->type()->classNamedTypes() as $type) {
+                $builder->use($type->name());
+            }
 
             $defaultValue = $interfaceMethodParameter->default();
+
             if ($defaultValue->isDefined()) {
                 $parameter ->defaultValue($interfaceMethodParameter->default()->value());
             }
```

mamazu:
Well I have just been applying your fixes. Maybe this isn't the smartest thing to do since now the code has my name on it. But I don't mind.

And on the factory idea. I think it would probably make sense to have a way to "implement methods" in a separate piece of logic. Because that's basically all this code does. "Implement methods" and add some bodies to the method.

dantleech:
works great thanks, added some documentation and  updated the changelog

mamazu:
I mean you did most of the important work. But great to see the feature added.

dantleech:
On the contrary, I spent minutes making tweeks, you did most of the work :)

PR #1886, opened 2022-09-23
Goto parent member definition

### LINKED ISSUE #1883: Jump to interface definition from implementation definition
It seems like PhpActor is not able to find the methods (type) definition from an interface.

Interface:
```php
<?php

interface FooInterface
{
    /**
     * @return string
     */
    public function getBar(): string;
}
```

Class:
```php
<?php

class Foo implements FooInterface
{
    public function getBar(): string
    {
        return 'abc';
    }
}
```
FooInterface::getBar() -> go to implementation -> works ✅
Foo::getBar() -> go to (type) definition -> does not work ❌

Let me know if you need any additional info.

dantleech:
For methods and properties, when going to definition, if the member is defined in a parent class (or interface), goto the parent class/interface.

Fixes #1883


lib/WorseReferenceFinder/WorseReflectionDefinitionLocator.php
```diff
@@ -245,4 +252,42 @@ private function gotoMember(NodeContext $symbolContext): TypeLocations
...
+        $path = $member->declaringClass()->sourceCode()->path();
+
+        if (null === $path) {
+            throw new CouldNotLocateDefinition(sprintf(
+                'The source code for class "%s" has no path associated with it.',
+                (string) $member->declaringClass()->name()
+            ));
+        }
+
+        return new TypeLocations([new TypeLocation($symbolContext->classType(), new Location(
```
dantleech, line 288: we culd quite easily return all the places where the method was "defined" here (e.g. abstract class, interface, trait...)

ckilb:
Hey, thanks for the implementation... but for me it doesn't work... 🤔
So what I did:
- Updated PHPActor:
`phpactor status
Version: "f3d15a83 (2 days ago) Gh-1898-1899 `
- Updated VS Code extension - but it's already up to date
- Opened a project and restarted indexing... waited until indexing completed.

`Go to implementation from interface` works. So it seems like PHP actor got the connection.
`Go to type definition`from class does not work ("no type definition found"). 
`Go to type definition` from class does not work (no error message, but cursor jumps to beginning of line 🤔

dantleech:
the type definition is something else, `function foobar(): Barfoo` => goto type definition will go to the class `Barfoo`.

here we implemented the "goto the interface definition on this method" which is simply `goto definition`

ckilb:
Sorry I fixed the typo - `Go to definition` neither does work. It will only move the cursor to beginning of the line.

ckilb:
What I would expect (and I thought this is the ticket here about):
If I open the file _MyClass.php_ and it contains a class like this:
`
<?php
class MyClass implements MyClassInterface 
{
   public function foo(): void {}
}
`

And I right-click on foo() and click "Go to definition" I'd expect that it opens the file 
_MyClassInterface.php_ and jumps to the line where the method is defined.

`
<?php
interface MyClassInterface
{
   public function foo(): void; // <-- jump here
}
`

dantleech:
yes exactly :thinking: 

recorded

that's in neovim, but behavior is the same with "control click" in Code

ckilb:
In this case don't think it's Spryker related (for any other issue I usually have I  - from experience - initially suspect Spryker to be the reason of failure though ;) ) 

Same same if I try with your example:
https://www.dropbox.com/s/tmqto8qt7zc0yxz/phpactor_jump2.mov?dl=0

dantleech:
maybe:

image

it's possible the binary you have in `$PATH` is not the one the extension is using

dantleech:
glad it's solved, maybe need to update some docs to point out that VS code extension installs it's own Phpactor by default

Issue #1530, opened 2022-05-03
Generics use case (partially covered)


camilledejoye:
Hi,

I'm really enjoying the new type system, great job again!

In the same line as the previous issue here is a small test scenario to show case some interesting use cases that could be added to the list of tests:
```php
<?php

namespace Test;

interface Foo {
    public function bar(): void;
}

interface Foos {
    /**
     * @return Collection<Foo>
     */
    public function all(): Collection;
}

/**
 * @template TKey of array-key
 * @template T
 * @template-extends \IteratorAggregate<TKey, T>
 * @template-extends \ArrayAccess<TKey|null, T>
 */
interface Collection extends \IteratorAggregate, \ArrayAccess
{
    /**
     * @return T
     */
    public function first();

    /**
     * @return Collection<TKey, T>
     */
    public function filter(\Closure $p);
}

function (Foos $foos) {
    foreach ($foos->all() as $foo) {
        $foo->bar(); // OK
    }

    $foos->all()->first()->bar(); // Not recognized
    $foos->all()[0]->bar(); // Not recognized

    $filtered = $foos->all()->filter(fn($foo) => $foo);
    foreach ($filtered as $foo) {
        $foo->bar(); // Not recognized
    }
}
```

dantleech:
yeah generics isn't 100% yet, i'll use this as a test case when I get back to it, thanks :)

dantleech:
Fixed with the exception of the ArrayAccess (#1593 ) in #1595 (see test case). In addition there is a follow up issue for further support in this direction #1594

PR #1905, opened 2022-10-09
Significant improvements to Generics

### LINKED ISSUE #1875: Wrong generic
```php
<?php

/**
 * @template TKey
 * @implements Iterator<TKey, string>
 */
abstract class Test1 implements Iterator
{
}

/**
 * @extends Test<int>
 */
class Test2 extends Test1
{
    public function current(): mixed
    {
    }

    public function next(): void
    {
    }

    public function key(): mixed
    {
    }

    public function valid(): bool
    {
    }

    public function rewind(): void
    {
    }

}

/** @var Test2 $a */

foreach ($a as $value) {
    /** @psalm-trace $value -> string  */
    // phpactor: $value -> int
}
```

dantleech:
Fixes #1875 

This PR aims to refactor the generic type resolution to be much saner, and potentially also support arguments.

TODO:

- [x] Replace `GenericTypeResolver` with the new "map" resolver
- [ ] Investigate passing template variables to constructor/methods
- [ ] Include class-string<t> etc
- [x] Include stubs which are fundamental to WR (e.g. enum, iterators) to the WR by default, hardcoded locations


lib/WorseReflection/Bridge/TolerantParser/Reflection/ReflectionClass.php
```diff
@@ -369,6 +370,11 @@ public function serviceLocator(): ServiceLocator
         return $this->serviceLocator;
     }
 
+    public function hierarchy(): ReflectionClassLikeCollection
+    {
+        return ReflectionClassLikeCollection::fromReflections((new ClassHierarchyResolver())->resolve($this));
```
dantleech, line 375: this should be memoized

lib/WorseReflection/Bridge/TolerantParser/Reflection/ReflectionInterface.php
```diff
@@ -158,6 +159,11 @@ public function docblock(): DocBlock
         );
     }
 
+    public function hierarchy(): ReflectionClassLikeCollection
+    {
+        return ReflectionClassLikeCollection::fromReflections((new ClassHierarchyResolver())->resolve($this));
```
dantleech, line 164: ditto

lib/WorseReflection/Core/Inference/GenericMapResolver.php
```diff
@@ -0,0 +1,64 @@
...
+            return $templateMap;
+        }
+
+        foreach (array_merge(
+            $topReflection->docblock()->implements(),
+            $topReflection->docblock()->extends()
+        ) as $genericClass
+        ) {
+            if (!$genericClass instanceof GenericClassType) {
+                // todo: trigger a diagnostic?
```
dantleech, line 41: probably not in this PR, but would be good to have a diagnostic buffer where we can inject diagonstics for stuff like this

lib/WorseReflection/Tests/Inference/generics/gh-1530-example.test
```diff
@@ -36,14 +36,4 @@ function foobar(Foos $foos) {
...
-
-    wrAssertType('Test\Collection<int,Test\Foo>', $foos->all());
-    wrAssertType('Test\Foo', $foos->all()->first());
-    wrAssertType('void', $foos->all()->first()->bar());
-    wrAssertType('Test\Collection<int,Test\Foo>', $foos->all()->filter(fn($foo)));
-
-    $filtered = $foos->all()->filter(fn($foo) => $foo);
-    foreach ($filtered as $foo) {
-        wrAssertType('void', $foo->bar());
-    }
```
dantleech, line 48: add back

PR #1908, opened 2022-10-10
Support for `class-string<T>` as an `@template`

dantleech:
This PR adds support for `class-string<T>`:

```
class Foo {
    /**
     * @template T
     * @param class-string<T> $class
     * @return T
     */
    public function foobar($class): object
    {
    }
}

$f = new Foo();
$f = $f->foobar(Foo::class);

wrAssertType('Foo', $f);
```

```
class Foo {
    /**
     * @template T
     * @param class-string<T>|null $tagFqn
     * @return ($tagFqn is string ? Generator<T> : Generator<TagNode>)
     */
    public function tags(?string $tagFqn = null): Generator
    {
    }
}

$f = new Foo();
$f = $f->tags(Foo::class);

wrAssertType('Generator<Foo>', $f);
```


lib/WorseReflection/Bridge/Phpactor/DocblockParser/TypeConverter.php
```diff
@@ -247,6 +247,13 @@ private function convertGeneric(GenericNode $type): Type
         $classType = $this->convert($type->type);
 
         if ($classType instanceof ClassStringType) {
+            if ($type->parameters()->count()) {
```
dantleech, line 250: extract `$parameters` - possibly save some iterations

lib/WorseReflection/Core/Type/StringType.php
```diff
@@ -15,4 +16,12 @@ public function emptyType(): Type
     {
         return new StringLiteralType('');
     }
+
+    public function accepts(Type $type): Trinary
+    {
+        if ($type instanceof ClassStringType) {
```
dantleech, line 22: there is a better way to do this

lib/WorseReflection/Core/Inference/GenericMapResolver.php
```diff
@@ -64,4 +69,47 @@ public function resolveClassTemplateMap(Type $topClass, ClassName $bottomClass,
...
+
+    public function mergeParameters(TemplateMap $templateMap, ReflectionParameterCollection $parameters, FunctionArguments $arguments): TemplateMap
+    {
+        foreach ($parameters as $parameter) {
+            $parameterType = $parameter->inferredType();
+            if ($parameterType instanceof ClassStringType && $parameterType->className()) {
+                $this->mapClassString($parameterType, $templateMap, $arguments, $parameter);
+                return $templateMap;
+            }
+            $parameterType->map(function (Type $type) use ($parameter, $templateMap, $arguments) {
```
dantleech, line 81: maybe to avoid this `mapIncludingSelf`?

lib/WorseReflection/Core/Inference/Resolver/MemberAccess/NodeContextFromMemberAccess.php
```diff
@@ -146,7 +148,22 @@ private function resolveMemberType(Frame $frame, ReflectionMember $member, Node
 
         $declaringClass = self::declaringClass($member);
 
-        if (count($declaringClass->templateMap())) {
+        $templateMap = $member->docblock()->templateMap();
+        if ($member instanceof ReflectionMethod && count($member->docblock()->templateMap())) {
+            $arguments = $this->resolveArguments($resolver, $frame, $node->parent);
```
dantleech, line 153: extract a method

PR #1919, opened 2022-10-17
Adding documentation to the goto commands

### LINKED ISSUE #1030: [RPC][documentation] Missing documentation for "language" parameter

mamazu:
closes #1030

### DIFF
--- a/doc/reference/rpc_command.rst
+++ b/doc/reference/rpc_command.rst
@@ -1036,6 +1036,9 @@ _RpcHandler_GotoDefinitionHandler
 Type: string
 
 
+Language of the current file
+
+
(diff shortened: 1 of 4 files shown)


mamazu:
@dantleech

dantleech:
thanks! as background info though: I'd like to quietly stop development of the legacy RPC system.

mamazu:
Yeah, that makes sense. Is there already a process to document the new commands?

PR #1922, opened 2022-10-19
Prophecy Extension

dantleech:
This is an experiment that shows how support for Prophecy can be realised:

image

- calls to `prophesize()` with a `class-string` argument resolved to `ObjectProphecy<1st-argumnet-type>`
- methods from `1st-argument-type` _added_ to the generic type as virtual methods, wrapping the types with the `MethodProphecy` generic.

I think this would also facilitate integration with PHPSpec cc @camilledejoye 

things that need to be worked on:

- allowing extensions to add stubs
- allowing the methods to be found by the reference finder (???)
- ...?

### DIFF
--- a/lib/Completion/Bridge/TolerantParser/WorseReflection/WorseClassMemberCompletor.php
+++ b/lib/Completion/Bridge/TolerantParser/WorseReflection/WorseClassMemberCompletor.php
@@ -209,7 +209,7 @@ class WorseClassMemberCompletor implements TolerantCompletor, TolerantQualifiabl
         }
 
         if ($classReflection instanceof ReflectionEnum) {
-            foreach ($members->cases() as $case) {
+            foreach ($members->enumCases() as $case) {
                 yield Suggestion::createWithOptions($case->name(), [
                     'type' => Suggestion::TYPE_ENUM,
                     'short_description' => fn () => $this->formatter->format($case),
--- a/lib/Completion/Bridge/TolerantParser/WorseReflection/WorseClassMemberCompletor.php
+++ b/lib/Completion/Bridge/TolerantParser/WorseReflection/WorseClassMemberCompletor.php
@@ -20,6 +20,7 @@ use Phpactor\WorseReflection\Core\Reflection\ReflectionClass;
 use Phpactor\WorseReflection\Core\Reflection\ReflectionEnum;
 use Phpactor\WorseReflection\Core\Reflection\ReflectionProperty;
 use Phpactor\WorseReflection\Core\Type;
+use Phpactor\WorseReflection\Core\Type\ClassType;
 use Phpactor\WorseReflection\Reflector;
 use Phpactor\Completion\Core\Suggestion;
 use Phpactor\WorseReflection\Core\Reflection\ReflectionInterface;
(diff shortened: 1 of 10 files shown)


lib/Completion/Bridge/TolerantParser/WorseReflection/WorseClassMemberCompletor.php
```diff
@@ -133,7 +134,8 @@ private function populateSuggestions(NodeContext $symbolContext, Type $type, boo
             return;
         }
 
-        foreach ($classReflection->methods() as $method) {
+        foreach ($type->members()->methods() as $method) {
```
dantleech, line 137: we enable ClassTypes to hold their own members() in addition to anything that might come from a reflected structure.

this might also provide support for anonymous classes.

lib/WorseReflection/Core/SourceCodeLocator/InternalStubs/Prophecy.php
```diff
@@ -0,0 +1,18 @@
...
+class ObjectProphecy implements ProphecyInterface
+{
+    /**
+     * @return T
+     */
+    public function reveal()
+    {
+    }
+}
+
```
dantleech, line 18: probably need to add custom stubs for Prophecy, and enable stubs to be added by extensions.

lib/Extension/Prophecy/WorseReflection/ProphecyMemberContextResolver.php
```diff
@@ -0,0 +1,61 @@
...
+        }
+
+        $innerType = TypeFactory::class($arg->className());
+        $innerReflection = $reflector->reflectClassLike($innerType->name());
+
+        $type = new GenericClassType($reflector, ClassName::fromString('Prophecy\Prophecy\ObjectProphecy'), [
+            $innerType
+        ]);
+
+        return $type->mergeMembers($innerReflection->members()->map(function (ReflectionMember $member) use ($reflector) {
```
dantleech, line 53: `map()` needs to be added to the "member collection" interface.

lib/WorseReflection/Core/SourceCodeLocator/InternalLocator.php
```diff
@@ -36,6 +36,8 @@ public static function forInternalStubs(): self
             'UnitEnum' => __DIR__ . '/InternalStubs/Enum.php',
             'BackedEnumCase' => __DIR__ . '/InternalStubs/Enum.php',
             'BackedEnum' => __DIR__ . '/InternalStubs/Enum.php',
+            'Prophecy\Prophecy\ObjectProphecy' => __DIR__ . '/InternalStubs/Prophecy.php',
+            'Prophecy\Prophecy\MethodProphecy' => __DIR__ . '/InternalStubs/Prophecy.php',
```
dantleech, line 40: remove this in favour of having extensions add their own stubs.

lib/WorseReflection/Core/Virtual/VirtualReflectionProperty.php
```diff
@@ -34,4 +34,8 @@ public function withClass(ReflectionClassLike $class): ReflectionMember
         $new->class = $class;
         return $new;
     }
+
+    public static function fromReflectionProperty(ReflectionProperty $property)
```
dantleech, line 38: rm

camilledejoye:
I'll try to take a look if I can find a bit of time.  
I dropped my first attempt to that because I feel like it's fixing something that should not even exists.

I would rather stop using phpspec all together instead of spending so much time trying to make it work with every other static analyzers.

dantleech:
> I'll try to take a look if I can find a bit of time.

no pressure, it would just be _possible_ with this PR where as before not so much I think :smile:

PR #1925, opened 2022-10-20
Updating the packages to remove webmozart/path-util

mamazu:


dantleech:
looks like conflict

composer.lock
```diff
@@ -1571,16 +1571,16 @@
         },
         {
             "name": "phpactor/class-to-file",
-            "version": "dev-master",
+            "version": "0.5.1",
```
mamazu, line 1574: I don't know what happened here. That could be a downgrade.

dantleech: yeah also not sure, but it's a downgrade indeed

dantleech (approved):

dantleech:
tagged a new class-to-file and updated master, thanks!

sitedyno:
Shouldn't this be changed now? https://github.com/phpactor/phpactor/blob/3ecebe2063f0efb875d5749d022745bd038f3652/lib/ConfigLoader/Adapter/PathCandidate/AbsolutePathCandidate.php#L7

mamazu:
Well the aim of the merge request was to remove the packages depending on it. But you are right, this needs to be fixed.

dantleech:
Oops, should really add that dep back or replace everything with the (urgh) "Filesystem" Path class.

mamazu:
Should I add the replace in the `composer.json` or are we good now?

dantleech:
the dependency is gone now or? nothing to do?

mamazu:
Yep

Issue #1258, opened 2021-04-30
Suggestions to Simplify FQN


afranioce:
Import the class when use full FQN

from
```php
<?php

namespace Foobar\Baz;

class Foo
{
    public function __construct(\Foobar\B<>ar $bar)
    {
    }
}
```

to

```php
<?php

namespace Foobar\Baz;

use Foobar\Bar;

class Foo
{
    public function __construct(Bar $bar)
    {
    }
}
```

weeman1337:
Others™ offer a code action for that

![image](https://user-images.githubusercontent.com/6216686/121145656-77014980-c83f-11eb-9ece-5c989a02b530.png)

This action can be applied to FQNs. It replaces all occurrences of the FQN in the current document by a short name with an import.

dantleech:
"classic" Phpactor also has an `:ExpandClass` action in VIM, so it would make sense to add both add code actions.

mamazu:
Would that be the same as ExpandClass but only the other way around? Then I can try to give it a shot.

dantleech:
Yes would be possible. Note that expand class doesn't have a code action ATM, maybe it could be a toggle.

mamazu:
So basically `:ExpandClass` and `:ExpandClass true` would turn the class name to an FQN and `:ExpandClass false` would turn the FQN into just the class name with an import?

dantleech:
I would just implement it as a code action.

Issue #1021, opened 2020-05-04
[RPC][class_expand] Introduce new RPC command: Class Expand


tkotosz:
In the documentation `Class Expand` mentioned as a feature but actually this command does exists. (RPC documentation is correctly not mentioning it)

I have checked the VIM plugin to see how it handles this class expand: It seem it does the following thing manually:
1. calls `class_search` RPC helper command (not user facing)
2. processes the returned `return_choice` "editor action" manually to apply the actual class expand action

I have skipped the implementation of this in the sublime plugin since I think this probably should be handled on the phpactor side by introducing a first-class command for class-expand.

mamazu:
As discussed here https://github.com/phpactor/phpactor/pull/1927#issuecomment-1287885631 Moving it to a code action wouldn't make sense since it would be limiting the usecase for it to mostly PHP files.

PR #1927, opened 2022-10-21
Adding a code action for expading classes

mamazu:
Just as discussed in #1021 and #1258 we should add a command to the language server that allows us to expand the current class name under the cursor.

### DIFF
--- a/lib/CodeTransform/Adapter/WorseReflection/Refactor/WorseExpandClass.php
+++ b/lib/CodeTransform/Adapter/WorseReflection/Refactor/WorseExpandClass.php
@@ -1,0 +1,59 @@
+<?php
+
+namespace Phpactor\CodeTransform\Adapter\WorseReflection\Refactor;
+
+use Microsoft\PhpParser\Node\QualifiedName;
+use Microsoft\PhpParser\Parser;
+use Phpactor\CodeTransform\Domain\Refactor\ExpandClass;
+use Phpactor\CodeTransform\Domain\SourceCode;
+use Phpactor\TextDocument\TextDocumentEdits;
+use Phpactor\TextDocument\TextEdit;
+use Phpactor\TextDocument\TextEdits;
+use Phpactor\WorseReflection\Core\Type\ClassType;
+use Phpactor\WorseReflection\Reflector;
+
+class WorseExpandClass implements ExpandClass
+{
+    private Reflector $reflector;
+
+    private Parser $parser;
+
+    public function __construct(
+        Reflector $reflector,
+        Parser $parser = null
+    ) {
+        $this->reflector = $reflector;
+        $this->parser = $parser ?: new Parser();
+    }
+
+    public function getTextEdits(SourceCode $sourceCode, int $offset): TextDocumentEdits
+    {
+        $symbolContext = $this->reflector
+            ->reflectOffset($sourceCode->__toString(), $offset)
+            ->symbolContext();
+        $type = $symbolContext->type();
+
+        if (!$type instanceof ClassType) {
+            return new TextDocumentEdits($sourceCode->uri(), TextEdits::none());
+        }
+
+        $newClassName = $type->name()->__toString();
+        $position = $symbolContext->symbol()->position();
+
+        return new TextDocumentEdits($sourceCode->uri(), TextEdits::fromTextEdits([
+            TextEdit::create($position->start(), $position->end() - $position->start(), $newClassName)
+        ]));
+    }
+
+    public function canExpandClassName(SourceCode $source, int $offset): bool
+    {
+        $node = $this->parser->parseSourceFile($source->__toString());
+        $targetNode = $node->getDescendantNodeAtPosition($offset);
+
+        if (!$targetNode instanceof QualifiedName) {
+            return false;
+        }
+
+        return $targetNode->isRelativeName();
+    }
+}
--- a/lib/CodeTransform/Tests/Adapter/WorseReflection/Refactor/WorseExpandClassTest.php
+++ b/lib/CodeTransform/Tests/Adapter/WorseReflection/Refactor/WorseExpandClassTest.php
@@ -1,0 +1,44 @@
+<?php
+
+namespace Phpactor\CodeTransform\Tests\Adapter\WorseReflection\Refactor;
+
+use Phpactor\CodeTransform\Adapter\WorseReflection\Refactor\WorseExpandClass;
+use Phpactor\CodeTransform\Tests\Adapter\WorseReflection\WorseTestCase;
+use Phpactor\CodeTransform\Domain\SourceCode;
+
+class WorseExpandClassTest extends WorseTestCase
+{
+    /**
+     * @dataProvider provideExpandClass
+     */
+    public function testExpandingClass(string $test): void
+    {
+        [$source, $expected, $offset] = $this->sourceExpectedAndOffset(__DIR__ . '/fixtures/' . $test);
+
+        $extractConstant = new WorseExpandClass(
+            $this->reflectorForWorkspace($source)
+        );
+        $textDocumentEdits = $extractConstant->getTextEdits(
+            SourceCode::fromStringAndPath($source, 'file:///source'),
+            $offset
+        );
+        $sourceCode = SourceCode::fromStringAndPath($source, 'file:///source');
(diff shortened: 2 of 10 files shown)


lib/CodeTransform/Adapter/WorseReflection/Refactor/WorseExpandClass.php
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    public function getTextEdits(SourceCode $sourceCode, int $offset): TextDocumentEdits
+    {
+        $symbolInformation = $this->reflector
+            ->reflectOffset($sourceCode->__toString(), $offset)
+            ->symbolContext()
+            ->symbol();
+
+        $result = $this->classSearch->classSearch('filesystem', $symbolInformation->name());
```
mamazu, line 40: This is probably not the best idea. I think we should be able to look through the use statements and see if any class name there matches. What do you think?

lib/CodeTransform/Tests/Adapter/WorseReflection/Refactor/WorseExpandClassTest.php
```diff
@@ -0,0 +1,66 @@
...
+        [$source, $expected, $offset] = $this->sourceExpectedAndOffset(__DIR__ . '/fixtures/' . $test);
+
+        $this->classSeach->classSearch('filesystem', 'Foo')->willReturn([
+            [
+                'file_path' => '/home/mamazu/phpacator/phpactor/SomeNamespace/Foo.php',
+                'class' => 'SomeNamespace\\Foo',
+                'class_name' => 'Foo',
+                'class_namespace' => 'SomeNamespace',
+            ]
+        ])->shouldBeCalled();
```
mamazu, line 37: Is that even the correct kind of return value for this class?

lib/CodeTransform/Adapter/WorseReflection/Refactor/WorseExpandClass.php
dantleech, line 40: no need to use "class search" and that class is very much legacy and should not be used here (note that it doesn't even come from a different "package" but actually from the "application".

you can just:

```
$type = $symbolInformation->type();
if (!$type instanceof ClassType) {
    return empty text edits;
}

$newClassName = $type->name()->__toString();
```

lib/CodeTransform/Tests/Adapter/WorseReflection/Refactor/WorseExpandClassTest.php
dantleech: we should not use `classSearch`, you can do everything with the `Reflector` I think.

dantleech (commented):
nice work! left some comments

lib/CodeTransform/Adapter/WorseReflection/Refactor/WorseExpandClass.php
mamazu: Yeah, the new implementation looks much cleaner now as well.

mamazu:
I replaced the class search with the code you suggested. Much more cleaner now.

lib/CodeTransform/Tests/Adapter/WorseReflection/Refactor/fixtures/expandClass1.test
```diff
@@ -0,0 +1,11 @@
...
+<?php
+use SomeNamespace\Foo;
+
+echo 'Debugging ' . F<>oo::class;
+
+// File: expected
+<?php
+use SomeNamespace\Foo;
+
+echo 'Debugging ' . SomeNamespace\Foo::class;
```
dantleech, line 11: we need to add a namespaced test:

```
namespace Foo;
```

either in this case or in both cases the expanded class should be absolute `\SomeNamespace\Foo::class`

dantleech:
hm. I've just realised that there is not really that useful. The original `:ExpandClass` would search for candidates and not require the class to already be imported. Expanding a class that's already imported isn't something you should ever need to do...

lib/CodeTransform/Tests/Adapter/WorseReflection/Refactor/fixtures/expandClass1.test
mamazu: True

mamazu:
Well, for classes that aren't imported we have the import command that does that. This would maybe help if you want to replace an import with a class reference directly like so:

```php
<?php
namespace Testing;

use Throwable;

try {
// something
} catch(Throwable) {}
```

You could convert that to:
```
<?php
namespace Testing;

try {
// something
} catch(\Throwable) {}
```

dantleech:
Right, but you could also just type backslash. In all other cases it doesn't help. Class is already imported.

dantleech:
The thing to consider is where does the value come in? I actually use `ExpandClass` quite frequently: but not in PHP files - I use it in XML and YAML files (e.g. service definitions). So I'd type `PaymentProcessor` and then expand it to `App\Payment\PaymentProcessor` or whatever.

Turning (imported) `(Foo\Bar)\Foo` to `\Foo\Bar\Foo`  in a PHP file doesn't really bring anything to the table.

mamazu:
True that makes much more sense. But then it doesn't make sense to implement that in the language server with a php parser?

dantleech:
In LS yes (needs to retrieve candidates to expand to) parser no....

mamazu:
So what would be the best course of action? Rewrite the Code Action to be compatible with XML and Yaml? How would one go about that?

dantleech:
We can certainly make a code action which acts the similar to the existing `:ExpandClass`. We can check the `$textDocument->extension() === 'xml'` for [example](https://github.com/phpactor/phpactor/blob/master/lib/Extension/Behat/ReferenceFinder/StepDefinitionLocator.php#L35), and use the `WordAtOffset` class to get the word at the given byte offset and use index search to get the class candidates and ask the user to select one if there is > 1 using the LSP client. Should it also work for PHP? Should it also "simply" the FQN?

But it' up to you, I don't think the goal is here is clearly defined. Would it make your day-to-day experience better in the editor?

mamazu:
I mean I haven't really considered the use cases for this. For us we moved our configuration to php, which is much easier to deal with than XML and Yaml. But it also makes sense to implement it for XML and Yaml. 

I mean the best way would be to fully support XML configuration like in Symfonys case, have autocomplete for the places where classes are allowed.

dantleech:
yeah, I don't really want to get into XML. Last time I checked it seemed I would have had to write an an XML parser for XML completion.

mamazu:
True, and this would maybe also make the code action more complicated. But this would mean that if we can't use the PHP reflector that we would also have to revert to using the class search functionality?

dantleech:
Well, we would use the index instead of whatever was used before, but yes, but that's necessary in anycsae (present a list of FQNs and choose one).

But I don't think it's clear what we're working or why atm. Is it an XML/HTML/Markdown/Blah plain text "fqn" expander (I'm not even sure that's a good idea, and if it were perhaps as an extension) or a PHP fqn expander (which as I said, has a limited use case)

mamazu:
Well the idea was to replace the implementation of the `:ExpandClass` function which is currently implemented in VIM. If that makes sense or not that is still up for decision.

dantleech:
What problem are we trying to solve - replacing an old command with a new one isn't solving a problem in itself. As said I find this useful primarily for expanding FQNs in XML and YAML (but also Markdown) and it doesn't really map to LSP as a code action as every single word would be eligible for expansion, and for PHP we have already established that it would bring little value. It would only be possible in it's current form as a custom Command which would then require a client-side addition (e.g. in the vscode client or as nvim configuration).

So short of opening a can of worms I would close this I think.

mamazu:
Okay, I wanted to tackle those issues:

#1021 - This is directly mentioning what I was trying to implement. But it seems like that this doesn't make sense since the base functionality is already implemented. And implementing it as a code action wouldn't make sense (as by our discussion).

#1258 - The complete opposite. That would probably make the most sense as a code action. To be able to simplify the FQN to a relative name and add an import for it.

mamazu:
Seeing that those two tickets have the complete opposite things in mind. I will just close the MR here and only to solve 1258, which I think is a cool functionality and could be useful when refactoring code.

dantleech:
cool makes sense :+1:

PR #1928, opened 2022-10-22
Replacing the class references to the webmozarts/path-util

mamazu:
Since we already remove the package dependency to the webmozarts/path-util package. We should also remove the references in code.

This isn't a big issue since the package is still in the pslam dependency. So it's still installed somehow.

### DIFF
--- a/lib/Extension/ClassMover/Application/ClassCopy.php
+++ b/lib/Extension/ClassMover/Application/ClassCopy.php
@@ -6,8 +6,8 @@ use Phpactor\ClassMover\ClassMover as ClassMoverFacade;
 use Phpactor\ClassMover\Domain\Name\FullyQualifiedName;
 use Phpactor\Filesystem\Domain\Filesystem;
 use Phpactor\Phpactor;
+use Symfony\Component\Filesystem\Path;
 use Webmozart\Glob\Glob;
-use Webmozart\PathUtil\Path;
 use Phpactor\Filesystem\Domain\CopyReport;
 use Phpactor\Extension\Core\Application\Helper\ClassFileNormalizer;
 use Phpactor\Extension\ClassMover\Application\Logger\ClassCopyLogger;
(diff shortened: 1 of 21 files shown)


lib/Indexer/Extension/IndexerExtension.php
```diff
@@ -119,17 +117,6 @@ public function load(ContainerBuilder $container): void
...
-    {
-        $builder = ReflectorBuilder::create();
-        foreach (array_keys($container->getServiceIdsForTag(WorseReflectionExtension::TAG_SOURCE_LOCATOR)) as $serviceId) {
-            $builder->addLocator($container->get($serviceId), 128);
-        }
-        $builder->enableCache();
-
-        return $builder->build();
-    }
-
```
mamazu, line 132: Phpstan has complained that this methods has no use so I removed it. And it seems like it's fine.

lib/Extension/ClassMover/Application/ClassCopy.php
```diff
@@ -72,7 +72,7 @@ public function copyFile(ClassCopyLogger $logger, string $srcPath, string $destP
             // if the src is not the same as the globbed src, then it is a wildcard
             // and we want to append the filename to the destination
             if ($srcPath !== $globPath) {
-                $globDest = Path::join($destPath, Path::getFilename($globPath));
+                $globDest = Path::join($destPath, basename($globPath));
```
dantleech, line 75: hm, which this change?

dantleech: ok, old version did basically the same thing.

dantleech (approved):

dantleech:
great, thanks!

lib/Extension/ClassMover/Application/ClassCopy.php
mamazu: Yeah, I basically replaced the function by it's implementation.

PR #1933, opened 2022-10-22
Enable class-like types to have members

dantleech:
Currently the members for a given type are determined by the static reflection of that class(like)s structure.

- The prophecy extension would use this concept to dynamically introduce virtual members to a class such as "MethodProphecy" which implements the magic `__call()`.
- Dynamic members could also be used to provide suppor for anonymous class types.

Still not 100% sure this is the right idea, 

This PR also highlights an issue in having `ClassType` and `ReflectedClassType`.

`ClassType` is a lightweight "class" which only knows the class name, whereas `ReflectedClassType` actually implements `instanceof` etc and now `members()`.

Types were intended to be "values" and not depend on services, so now to create class types you really need to have the reflector service injected. The solution would be to make types "lightweight" and put this logic in a service somewhere.


lib/WorseReflection/Core/Type/ClassType.php
```diff
@@ -78,4 +107,12 @@ public function emptyType(): Type
     {
         return $this;
     }
+
+    public function asReflectedClasssType(Reflector $reflector): ReflectedClassType
```
dantleech, line 111: :grimacing:

PR #1944, opened 2022-11-11
Add LSP Client Guides for Emacs

zonuexe:
Add a guide to migrate users using Legacy RPC in Emacs to LSP client.

I haven't written reST in a long time so sorry if the style is broken.


dantleech (approved):

dantleech:
nice addition, thanks @zonuexe

PR #1946, opened 2022-11-12
Add mutator generation

Florian-Merle:
I mostly copied and adapted what was done for the accessor generation in this pull request.


Florian-Merle:
Also, I might be missing something, but opening phpactor context menu in nvim, the new code action is missing. @dantleech you might know why ?

image

doc/reference/configuration.rst
```diff
@@ -201,6 +201,45 @@ If the first letter of a generated accessor should be made uppercase
...
+If the first letter of a generated mutator should be made uppercase
+
+
+**Default**: ``true``
+
+
+.. _param_code_transform.refactor.generate_mutator.fluant:
+
+
+``code_transform.refactor.generate_mutator.fluant``
```
dantleech, line 233: spelling mistake => fluent

lib/CodeTransform/Tests/Adapter/WorseReflection/Refactor/fixtures/generateMutator1.test
```diff
@@ -0,0 +1,20 @@
...
+    pr<>ivate $method;
+}
+// File: expected
+<?php
+
+class generateMethod
+{
+    private $method;
+
+    public function method($method): void
```
dantleech, line 16: I think it would be very rare to have a non-prefixed mutator? so not sure this should be the default

lib/Extension/LanguageServerCodeTransform/LspCommand/GenerateMutatorsCommand.php
```diff
@@ -0,0 +1,62 @@
...
+    /**
+     * @param string[] $propertyNames
+     * @return Promise<ApplyWorkspaceEditResponse|null>
+     */
+    public function __invoke(string $uri, int $startOffset, array $propertyNames): Promise
+    {
+        $textDocument = $this->workspace->get($uri);
+
+        try {
+            $textEdits = $this->generateMutator->generate(
```
dantleech, line 48: I guess the API is the same for GenerateAccessor and GenerateMutator? could we create an interface `PropertyAccessGenerator` and then we can use the same infrastructure for both cases?

lib/Extension/LanguageServerCodeTransform/CodeAction/GenerateMutatorsProvider.php
```diff
@@ -0,0 +1,90 @@
...
+                return [];
+            }
+
+            return [
+                CodeAction::fromArray([
+                    'title' => sprintf('Generate %s mutator(s)', count($propertyNames)),
+                    'kind' => self::KIND,
+                    'command' => new Command(
+                        sprintf('Generate %s mutator(s)', count($propertyNames)),
+                        GenerateMutatorsCommand::NAME,
```
dantleech, line 79: is there anything different about this command other than the command name? can we parameterise the Accessor class rather than copy / paste this one?

dantleech (commented):
nice work, I think we could possibly reduce the copy/paste here by introducing a common interface and parameterising the original code provider / command

lib/CodeTransform/Tests/Adapter/WorseReflection/Refactor/fixtures/generateMutator1.test
Florian-Merle: The default prefix is `set` see https://github.com/phpactor/phpactor/pull/1946/files#diff-0a820ddea77da22f468ce7dc570c241bd40a8b2f918314da9a1e9c36fdfe0c1cR112

I can change the tests so they all use `set` as a prefix though

dantleech: Ah OK, no worries it's fine

lib/Extension/LanguageServerCodeTransform/LspCommand/GenerateMutatorsCommand.php
Florian-Merle: Indeed they're the same. Sure, but should we keep the `GenerateMutator` interface and make it extends `PropertyAccessGenerator`, or should it be removed?

dantleech: If they are exactly the same prob fine to rename the existing one

lib/Extension/LanguageServerCodeTransform/CodeAction/GenerateMutatorsProvider.php
Florian-Merle: They're pretty much the same. I refactored so we only have one provider now. As you can see in the class's constructor, we have 3 new arguments (`kind`, `command`, and `type`). However, I'm not satisfied with the name of the `type` property, maybe you have a better suggestion?

See https://github.com/phpactor/phpactor/pull/1946/files#diff-86a88fb7dd73a00bf12d3e88a9c9bf24a70729458439176bd56afef0775dce7cR31

lib/Extension/LanguageServerCodeTransform/LspCommand/GenerateMutatorsCommand.php
Florian-Merle: It's done :)

lib/Extension/LanguageServerCodeTransform/CodeAction/PropertyAccessGeneratorProvider.php
```diff
@@ -16,23 +15,33 @@
...
+    private string $type;
 
     private Reflector $reflector;
 
-    public function __construct(Reflector $reflector)
-    {
+    public function __construct(
+        string $kind,
+        string $command,
+        string $type,
```
dantleech, line 31: generatorRole ? but `type` is also kinda ok.

lib/Extension/LanguageServerCodeTransform/Tests/Unit/CodeAction/PropertyAccessGeneratorProviderTest.php
```diff
@@ -17,7 +16,8 @@
 use Prophecy\PhpUnit\ProphecyTrait;
 use function Amp\Promise\wait;
 
-class GenerateAccessorProviderTest extends TestCase
+/** @group wip */
+class PropertyAccessGeneratorProviderTest extends TestCase
```
dantleech, line 20: remove `@group` (btw can easily run a specific test by specifying `phpunit path/to/testTest.php --filter="provide actions`)

phpstan-baseline.neon
```diff
@@ -1680,6 +1680,11 @@ parameters:
 			count: 1
 			path: lib/CodeTransform/Tests/Adapter/WorseReflection/Refactor/WorseGenerateMethodTest.php
 
+		-
+			message: "#^Method Phpactor\\\\CodeTransform\\\\Tests\\\\Adapter\\\\WorseReflection\\\\Refactor\\\\WorseGenerateMutatorTest\\:\\:provideExtractMutator\\(\\) has no return type specified\\.$#"
```
dantleech, line 1684: you can remove this and use the `generate docblock` code action to add the docblock (might need tweaking though)

dantleech (commented):
couple of minor comments, but otherwise it would be good to go

dantleech:
testing locally:

```
<?php

namespace Phpactor;

class Test {

    private $foobar;

}
```

:no_entry_sign: I no longer get the option to generate an accessor:

image

and it seems the command is not registered:

```
phpactor: -32603: Exception [RuntimeException] Command "language_server_code_transform.generate_mutators_command" not found, known commands: "name_import", "tra
nsform", "create_class", "generate_method", "extract_method", "replace_qualifier_with_import", "extract_constant", "generate_accessors", "generate_mutators", "i
mport_all_unresolved_names", "extract_expression", "generate_decorator" at /home/daniel/www/phpactor/phpactor/vendor/phpactor/language-server/lib/Core/Command/C
ommandDispatcher.php#44 
```

are you able to test it locally?

lib/Extension/LanguageServerCodeTransform/CodeAction/PropertyAccessGeneratorProvider.php
Florian-Merle: I changed it to generatorRole

lib/Extension/LanguageServerCodeTransform/Tests/Unit/CodeAction/PropertyAccessGeneratorProviderTest.php
Florian-Merle: :see_no_evil: whoops

phpstan-baseline.neon
Florian-Merle: done :)

Florian-Merle:
The command should be registered now.

As [I asked you before](https://github.com/phpactor/phpactor/pull/1946#issuecomment-1312287444), I can't use the command to generate mutators in neovim. I'm not quite sure why yet, but it's probably due to the way I set up my config.

dantleech:
>  I'm not quite sure why yet, but it's probably due to the way I set up my config.

it's a range code action, so it's only effective on ranges, with nvim LSP I do a visual range select and then invoke my code action command, the mappings are:

```
    vim.api.nvim_buf_set_keymap(bufnr, 'n', ',ca', '<cmd>lua vim.lsp.buf.code_action()<CR>', opts)
    vim.api.nvim_buf_set_keymap(bufnr, 'v', ',ca', '<cmd>lua vim.lsp.buf.range_code_action()<CR>', opts)
```

equally I don't like this interface much. When I wrote the original command CoC and Nvim did not pass the offset when invoking a single/normal code action, but would apss the entire document range instead. That behavior seems to have changed now, so we could do it as a normal code action.

Florian-Merle:
Okay, this should work for range code actions.
 
image

Also, I made it work for the context menu

image

dantleech:
fixed the docs in https://github.com/phpactor/phpactor/pull/1948 will merge that once built

dantleech:
FTR I don't really want to support the "legacy" VIM functionality going forward as it's effectively replaced with LSP which has far better client support :)

dantleech:
merged, thanks for this contribution :tada:

Florian-Merle:
Thank you for your time :)

PR #1949, opened 2022-11-13
Trying to fix import problems

### LINKED ISSUE #1941: FQN -> import can add a duplicated import
If we try to convert inline `Z\X` to import, we get a duplicate `use Z\X` if there exist another an import for `X`, e.g. `use A\X`.

Thanks for #1939 anyway!

mamazu:
I have added an aliasing mechanism in the use statement updater. This PR still has two problems though:

* missing tests
* it doesn't update the token that has been imported

I just wanted some feedback if I am on the right track.

Resolves #1941

### DIFF
--- a/lib/CodeBuilder/Adapter/TolerantParser/TolerantUpdater.php
+++ b/lib/CodeBuilder/Adapter/TolerantParser/TolerantUpdater.php
@@ -1,6 +1,7 @@
 <?php
 namespace Phpactor\CodeBuilder\Adapter\TolerantParser;
 
+use Phpactor\CodeBuilder\Adapter\TolerantParser\Util\UseStatementDeduplicator;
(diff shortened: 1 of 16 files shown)


mamazu:
The idea in this case would be to replace the `NameImporter` Functionality with the one in the updater. So that in the `NameImporter` you just add the class you want to import and the rest is handled by the updater.

mamazu:
I have just checked. The problem is if we want to use the `NameImporter` as is, this will run in an infinite loop. Because calling the `NameImporter` in the `UseStatementUpdater` would call the `UseStatementUpdater` to import the new name.

mamazu:
Okay, I think the only thing missing now is that the replace still replaces the name with the wrong class name. @dantleech could you have a look why that is?

PR #1950, opened 2022-11-13
Using array syntax to destructure

mamazu:
https://wiki.php.net/rfc/short_list_syntax

Some of the codebase already uses it, now everything uses it.

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -36,6 +36,7 @@ return (new Config())
         'no_empty_phpdoc' => true,
         'phpdoc_trim' => true,
         'array_syntax' => ['syntax' => 'short'],
+        'list_syntax' => ['syntax' => 'short'],
         'void_return' => true,
         'ordered_class_elements' => true,
         'single_quote' => true,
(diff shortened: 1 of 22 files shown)


dantleech:
Pretty sure there is a cs fixer fiz for this?

mamazu:
I tried using rector to get it up to 7.4 standards but that would be too much of a change so I split this up into smaller parts. But if you want I can check if there is a check in the ECS config.

dantleech:
we already use `php-cs-fixer`: https://mlocati.github.io/php-cs-fixer-configurator/#version:3.13|fixerset:@PHP74Migration

mamazu:
Yeah, that doesn't seem to be exhaustive though and it looks like it also only migrates from 7.3 since the feature this branch fixes is a 7.1 feature.

dantleech:
the php-cs-fixer is very minimal at the moment (we are not using many of those "migration" fixes). See https://github.com/phpactor/phpactor/pull/1900 for an example of applying more comprehensive CS rules (didn't merge for other reasons).

dantleech:
In general if we have a CS change it needs to be in cs-fixer.

mamazu:
Yeah, adding a cs rule will also prevent regression. Found the rule and added it.

PR #1957, opened 2022-11-19
Using constructor property promotion

mamazu:
Since we are using php8 now, I feel like this feature is really slimming down the code. If not that it also found 2 errors:
* In the `Prototype/MethodBody.php` the lines property should be nullable
* In the `WorseReflection/ParsedDocblock.php` the paramName can be null as well which `ltrim` doesn't allow.

## PHPCS

Further more I have also checked if there is a rule in PHPCS for this to enforce new code is written that way but only found this.
https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/issues/5956

There is a custom rule (not in the main package) that works in some cases. I don't know if it's worth adding.


lib/ClassMover/Domain/Model/ClassMemberQuery.php
```diff
@@ -17,17 +17,10 @@ final class ClassMemberQuery
...
     ];
 
-    private ?Class_ $class;
-
-    private ?MemberName $memberName;
-
     private ?string $type;
 
-    private function __construct(Class_ $class = null, MemberName $memberName = null, string $type = null)
+    private function __construct(private ?\Phpactor\ClassMover\Domain\Model\Class_ $class = null, private ?\Phpactor\ClassMover\Domain\Name\MemberName $memberName = null, string $type = null)
```
dantleech, line 22: I think we're missing another rector to convert these to imports? and/or possibly there is a cs fixer fix for that

dantleech:
>  I don't know if it's worth adding.

no, I don't think we need to enforce this at this time (can also just add it later if it's gets into mainline)

lib/ClassMover/Domain/Model/ClassMemberQuery.php
mamazu: Probably the job of the CS fixer. But I didn't found the rule for that but there weren't many instances so I manually fixed them. Luckily Simplify FQN is a thing now. :D

mamazu:
Can you see what the CS fixer is complaining about? Because under 8.1 the fixer throws no errors.

lib/ClassMover/Domain/Model/ClassMemberQuery.php
dantleech: https://mlocati.github.io/php-cs-fixer-configurator/#version:3.13%7Cfixer:fully_qualified_strict_types :)

dantleech:
did you do a composer install?

mamazu:
Yes, deleting the vendor folder and reinstalling from the `composer.lock` file yields the same results.

dantleech:
delete the `.php-cs-fixer.cache` file? works for me on 8.0 and 8.1 same results as CI

dantleech:
btw no need to squash / amend commits, better to keep them atomic. they get squashed on merge anyway.

mamazu:
Yeah, removing the cache. That's what it was.

Ah yeah, makes sense. I'll stop using them.

PR #1959, opened 2022-11-21
Using php8 string functions

mamazu:
In php8 you can use `str_contains`, `str_ends_with` and `str_starts_with` to replace calls to the `str_pos` function and make the code much more readable.

Sadly there isn't a rule for that in phpcs.


mamazu:
Is there anything wrong with this PR? Or are you holding off on the new string functions?

dantleech (approved):

dantleech:
no all good :+1:

Issue #1961, opened 2022-11-21
Type information not used in foreach (type is union of empty array and typed array)


dantleech:
```
        $offsets = [];
        $highlights = [];
        foreach ($this->generate($source, $offset) as $highlight) {
            $offsets[] = $highlight->start;
            $offsets[] = $highlight->end;
            $highlights[] = $highlight;
        }

        $highlights; // <- has type array{}/Highlight[]

        $lineCols = EfficientLineCols::fromByteOffsetInts($source, $offsets, true);
        $lspHighlights = [];

        foreach ($highlights as $highlight) { // <- highlight has no type
            $startPos = $lineCols->get($highlight->start);
            $endPos = $lineCols->get($highlight->end);
```

PR #1962, opened 2022-11-21
Fix #1960 document highlight (regression?)

### LINKED ISSUE #1960: Document highlight does not work with recent changes
I switched back to 707e7d90 (before `constructor_property_promotion`) to make it working for me.

Can you confirm this bug? I get it on two separate environments.

dantleech:
Seems to work again after fixing the types... maybe an issue rather with PHP 8 fixes #1960

### DIFF
--- a/lib/Extension/LanguageServerReferenceFinder/Model/Highlighter.php
+++ b/lib/Extension/LanguageServerReferenceFinder/Model/Highlighter.php
@@ -45,6 +45,7 @@ class Highlighter
         $lspHighlights = [];
 
         foreach ($highlights as $highlight) {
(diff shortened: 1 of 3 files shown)


lib/Extension/LanguageServerReferenceFinder/Model/Highlighter.php
```diff
@@ -45,6 +45,7 @@ public function highlightsFor(string $source, ByteOffset $offset): Highlights
         $lspHighlights = [];
 
         foreach ($highlights as $highlight) {
+            assert($highlight instanceof Highlight);
```
dantleech, line 48: added because there's a bug with type infernce #1961

PR #1971, opened 2022-11-27
Add support for array union operator

wouterj:


lib/WorseReflection/Tests/Inference/binary-expression/array-union.test
```diff
@@ -0,0 +1,3 @@
+<?php
+
+wrAssertType('array{}&array{a:1}', [] + ['a' => 1]);
```
dantleech, line 3: _technically_ this should be generalized to `array{a:1}` as the evaluation no longer has `[]` as a possiblity. There is `$type->generalize()` although this is could be some work for little gain.

dantleech (approved):

PR #1972, opened 2022-11-29
Add LSP diagnostics and quickfix code actions for PHP CS Fixer

lumnn:
WIP: I still haven't fully finished, there is not tests, and I'm not 100% sure about this working reliably. Adding PR through now to see whether you're interested in these changes and to get some feedback on what needs to be changed.

This PR changes almost entirely how PHP CS Fixer integration works.

So far integration only allowed for formatting document. The produced text edit was a whole file replacement.

This PR keeps formatter with few advantages:

- it only produces TextEdits for changed places
- doesn't create any temporary files

Additionally it adds diagnostics and quickfix actions for them.
Everything above works by sending code to php-cs-fixer through stdin and reading it's `--dry-run --diff`.

LSP Diagnostics works similarly, but executes php-cs-fixer few times in order to know where specific rules apply:

1. Runs php-cs-fixer with `--verbose` to know applied rules.
2. Loops for each applied rule and gets diff for that specific rule only.
3. Runs php-cs-fixer describe, to get human readable descriptions of errors

This is how diagnostics presents in Sublime Text:

image


lib/Extension/LanguageServerPhpCsFixer/Provider/PhpCsFixerDiagnosticsProvider.php
```diff
@@ -0,0 +1,231 @@
...
+use Phpactor\LanguageServer\Core\CodeAction\CodeActionProvider;
+use Phpactor\LanguageServer\Core\Diagnostics\DiagnosticsProvider;
+use Psr\Log\LoggerInterface;
+use SebastianBergmann\Diff\Line;
+use SebastianBergmann\Diff\Parser;
+use LogicException;
+
+class PhpCsFixerDiagnosticsProvider implements DiagnosticsProvider, CodeActionProvider
+{
+    private array $ruleDescriptions = [];
```
lumnn, line 26: Rule descriptions are kept in memory after once fetched, hope that's okay

lib/Extension/LanguageServerPhpCsFixer/Provider/PhpCsFixerDiagnosticsProvider.php
```diff
@@ -0,0 +1,231 @@
...
+        }
+
+        return \Amp\call(function () use ($textDocument) {
+            $title = 'Format with PHP CS Fixer';
+
+            $actions = [
+                CodeAction::fromArray([
+                    'title' => $title,
+                    'kind' => 'quickfix',
+                    'diagnostics' => $this->lastPublishedDiagnostics,
```
lumnn, line 187: CodeAction is added to all last published diagnostics. I don't think this should done this way. I'm not sure how else could I do this yet, but I haven't got deep into this yet.

lib/Extension/LanguageServerPhpCsFixer/Exception/CliException.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+namespace Phpactor\Extension\LanguageServerPhpCsFixer\Exception;
+
+use RuntimeException;
+
+class CliException extends RuntimeException
```
dantleech, line 7: could be `PhpCsFixerError`

lib/Extension/LanguageServerPhpCsFixer/Provider/PhpCsFixerDiagnosticsProvider.php
dantleech: code action can simply recalculate :shrug:

dantleech (commented):
hm, I tried to create a text-edit diff between old and new when I originall implemented this, firstly using the existing:

https://github.com/phpactor/phpactor/blob/16c6f106281d5ac469a26d671972822e21bc3f1e/lib/Extension/Rpc/Diff/TextEditBuilder.php#L16

and then by implementing an own diff algorithm using https://en.wikipedia.org/wiki/Longest_common_subsequence_problem

I gave up finally as I was burning too much time, and in anycase, I wasn't providing diagnostics and so didn't care about it actually.  Looks like you have another solution.

The existing method above I mentioned is used for RPC currently, but it's broken, it just happens to work well enough.

Having a generic method to create a text-edit diff between two documents would be generally useful, but I think it's some work and could be tackled in a separate PR even.

dantleech:
R.e. the text edit diff bit, I made a "proper" way to get the text edits to convert one document to another:

https://github.com/phpactor/phpactor/blob/master/lib/TextDocument/TextEditDiff.php

(i created it to solve a problem that I didn't have, so slughtly untested in real-life atm :D )

lumnn:
I've corrected the issues, and resolved appropriate review comments.

Regarding the text edits. As far as I understand the class you provided creates `TextEdits` by providing 2 versions of a string.
This would be rather useful with existing implementation of php-cs-fixer, which, if I remember correctly, outputs fixed file to temporary folder, therefore I'd have 2 strings I could pass into the `TextEditDiff::diff` method.

This PR changes that behaviour, so `php-cs-fixer` gives me a udiff formatted diff between file directly in the command output. ([--diff option documentation](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/master/doc/usage.rst))

The reason I chosen diff format was that I wanted to avoid unnecessary file writes and reads as well as to create a TextEdits that changes only affected parts (not the whole file as per original implementation)

Given above, what do you think should be done?

lumnn:
Before merging this, I want to also have a look into adding a source action prefixed `source.fixAll.` or something like this, as I believe it may be used as auto-action on save in some LSP client implementation - [stylelint example](https://github.com/stylelint/vscode-stylelint#editorcodeactionsonsave)

lib/Extension/LanguageServerPhpCsFixer/Model/PhpCsFixerProcess.php
```diff
@@ -0,0 +1,114 @@
...
+                ->onResolve(function (?Throwable $error, $data) use ($process): void {
+                    $this->logger->log(
+                        $error ? 'warning' : 'debug',
+                        sprintf(
+                            'Executed %s, which exited with %s',
+                            $process->getCommand(),
+                            $data
+                        )
+                    );
+                });
```
dantleech, line 109: any reason to use `onResolve`?

can do:

```
yield $process->join();
```
?

lumnn: I'm not going to lie, but I did forgot a bit of a implementation details and reasoning, so the explanation may be wrong. I'm about to leave now, can do little testing tommorow to find out whether what I write below is true.

However, I believe the reason is in lines 39-41, that write to `stdin` of the process. `onResolve` is used to add a listener for a logging purposes.
I believe if I used `yield $process->join();` with the process that expects `stdin` then that Promise would get stuck waiting for input. Input from lines 39-41 wouldn't happen, because we'd be still on line 37 waiting for run to complete.

lib/Extension/LanguageServerPhpCsFixer/Provider/PhpCsFixerDiagnosticsProvider.php
```diff
@@ -0,0 +1,230 @@
...
+            }
+
+            return $diagnostics;
+        });
+    }
+
+    public function provideActionsFor(TextDocumentItem $textDocument, Range $range, CancellationToken $cancel): Promise
+    {
+        return \Amp\call(function () use ($textDocument, $cancel) {
+            $title = 'Format with PHP CS Fixer';
```
dantleech, line 174: I'm not sure this should be a code action. There is it's (already?) available as the document formatting request: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_formatting

lib/Extension/LanguageServerPhpCsFixer/Util/DiffLinesConsumer.php
```diff
@@ -0,0 +1,97 @@
...
+    {
+        return $this->eatWhileType(Line::ADDED);
+    }
+
+    public function current(): ?Line
+    {
+        return $this->chunk->getLines()[$this->position] ?? null;
+    }
+
+    private function revind(): void
```
dantleech, line 84: typo?

dantleech:
> something like this, as I believe it may be used as auto-action on save in some LSP client implementation - [stylelint example](https://github.com/stylelint/vscode-stylelint#editorcodeactionsonsave).

in that example it looks like an arbitrarily named code action / command, so I'd guess you could already use that feature in VS code without further modification

---

The PR looks fantastic, although there seem to be no tests?

lib/Extension/LanguageServerPhpCsFixer/LanguageServerPhpCsFixerExtension.php
```diff
@@ -17,20 +22,56 @@ class LanguageServerPhpCsFixerExtension implements OptionalExtension
...
+                'name' => 'php_cs_fixer.fix'
+            ],
+        ]);
     }
 
     public function configure(Resolver $schema): void
     {
         $schema->setDefaults([
             self::PARAM_PHP_CS_FIXER_BIN => '%project_root%/vendor/bin/php-cs-fixer',
             self::PARAM_ENV => [],
```
dantleech, line 74: maybe an option to disable diagnostics? I would prefer them to be disabled but still have the option to format the code.

lib/Extension/LanguageServerPhpCsFixer/Model/PhpCsFixerProcess.php
```diff
@@ -0,0 +1,114 @@
...
+    /**
+     * @param string[] $options
+     *
+     * @return Promise<string>
+     */
+    public function describe(string $rule, array $options = []): Promise
+    {
+        return call(function () use ($rule, $options) {
+            /** @var Process */
+            $process = yield $this->run('describe', ...[...$options, $rule]);
```
dantleech, line 72: image

:point_up: this is a bit "meta", would it be possible to parse out the first sentence?

dantleech: it might also be good to have just `rule_name` (or identifier) as the diagnostic title, then (perhaps optionally?) the rule description.

lumnn: I think the "Description of braces rule" part was not intended in first place, and shouldn't appear any more.

I did amend the code to remove that part, as well as I did remove few extra content for things as fixing configuration (that shows some of configuration options of php-cs-fixer related to the rule) and fixing example (code snippets from before and after fixing)

lumnn: If you look on the screenshot provided in first comment the rule name in my editor displays after the rule description.

I did briefly checked other LSP servers and most of them don't provide rule name in content, but when they do, then they append it on the end. I left it for now without the name

lib/Extension/LanguageServerPhpCsFixer/LanguageServerPhpCsFixerExtension.php
lumnn: Added in one of the commits

lib/Extension/LanguageServerPhpCsFixer/Provider/PhpCsFixerDiagnosticsProvider.php
lumnn: I think it was driven by the editor in my scenario.

For diagnostics, my editor is able to provide quickfix link that show just below the diagnostic content, which seems a bit convenient.
However, thinking further, that link would fix entire file which may get confusing why other problems are fixed, while user may have expected to fix only that specific problem.

I also revieved other LSP servers, and some of them do provide quickfix, and as expected that quickfix resolves only that specific problem.

Therefore I've removed it (well replaced with `source.fixAll` command name, more on that within other comment)

lib/Extension/LanguageServerPhpCsFixer/Util/DiffLinesConsumer.php
lumnn: I saw that word for about 5 years on a daily basis in the town I used to live and I yet I still didn't know it's spelling. Sorted!

dantleech:
Am I right in assuming there are no tests? If so do you need a hand with those?

lumnn:
> in that example it looks like an arbitrarily named code action / command, so I'd guess you could already use that feature in VS code without further modification

I end-up on adding a code action named `source.fixAll.phpactor.phpCsFixer`

I couldn't find much info about that, but this [vscode release notes](https://github.com/microsoft/vscode-docs/blob/main/release-notes/v1_31.md#fix-all-source-actions
) mentions about "convention" of source actions capable to fix all problems having this prefix.
Another ref on vscode docs: https://code.visualstudio.com/api/references/vscode-api#CodeActionKind

vscode is not alone about this, as I did successfully got that working with Sublime Text, which also mentions that source code action within configuration examples: https://github.com/sublimelsp/LSP/blob/main/LSP.sublime-settings#L25

Emacs is also aware of this convention - https://github.com/emacs-lsp/lsp-mode/issues/1842

Looking further I do find also many LSP language servers providing action with that prefix
https://sourcegraph.com/search?q=context%3Aglobal+source.fixAll&patternType=standard&sm=1&groupBy=repo

Therefore I added `source.fixAll.phpactor.phpCsFixer` that would allow other fixAll commands, and also let LSP clients to control which of them should be applied (for example on save)

lumnn:
I'll be looking into adding tests and updating documentation to add extra parameter, and update description of phpactor integration with php-cs-fixer

dantleech:
For documentation just run composer integrate (it's auto generated)

lib/Extension/LanguageServerPhpCsFixer/Provider/PhpCsFixerDiagnosticsProvider.php
```diff
@@ -0,0 +1,274 @@
...
+                ])
+            ];
+
+            return $actions;
+        });
+    }
+
+    public function kinds(): array
+    {
+        return ['source.fixAll.phpactor.phpCsFixer'];
```
dantleech, line 82: Is there some standard here? AFAIK this is completely arbitrary, and existing things are `<refactor|quickfix>.<thing>`

lib/Extension/LanguageServerPhpCsFixer/Provider/PhpCsFixerDiagnosticsProvider.php
```diff
@@ -0,0 +1,274 @@
...
+
+                                $startPos = new Position($startLineNo, $startChars);
+                                $endPos = new Position($lineNo - 1, $endChars);
+                            }
+
+                            $diagnostics[] = Diagnostic::fromArray([
+                                'message' => yield $this->explainRule($rule),
+                                'range' => new Range($startPos, $endPos),
+                                'severity' => DiagnosticSeverity::WARNING,
+                                'source' => $this->name().':'.$rule
```
dantleech, line 191: should we be "contaminating" the source with the rule name? what if it's used to surpress diagnostics on the client side?

lib/Extension/LanguageServerPhpCsFixer/Provider/PhpCsFixerDiagnosticsProvider.php
```diff
@@ -0,0 +1,274 @@
...
+                            }
+
+                            $startPos = new Position($startLineNo, 0);
+                            $lineLength = strlen($lastChangedLine->getContent());
+                            $endPos = $lineLength
+                                ? new Position($lineNo - 1, $lineLength)
+                                : new Position($lineNo, 0);
+
+                            if ($replacedLines) {
+                                $firstLineA = $replacedLines[0]->getContent();
```
dantleech, line 175: this is a lot of indentation, can each block be separated into separate methods so that:

```
foreach ($foos as $bar) {
     $bars = $this->fooBars($bar, $boo, $baz); // can pass lines arrays by references
}
```

and make the variable scopes more explicit and perhaps make the code easier to understand/

lib/Extension/LanguageServerPhpCsFixer/Provider/PhpCsFixerDiagnosticsProvider.php
```diff
@@ -0,0 +1,274 @@
...
+                                'severity' => DiagnosticSeverity::WARNING,
+                                'source' => $this->name().':'.$rule
+                            ]);
+
+                            $contextLines = [];
+
+                            continue;
+                        }
+
+                        throw new LogicException('This is not intended to be reached');
```
dantleech, line 232: minor thing -- i would make the success action the _last thing_. The condition above can be `if ($line->getType() != Line::ADDED || $prevLine->getType() !== Line::UNCHANGED) { // throw exception }`. So all the conditions before it are "guards" including the "this should never happen"

lib/Extension/LanguageServerPhpCsFixer/Util/DiffToTextEditsConverter.php
```diff
@@ -0,0 +1,79 @@
+<?php
+
+namespace Phpactor\Extension\LanguageServerPhpCsFixer\Util;
+
+use Phpactor\LanguageServerProtocol\Range;
+use Phpactor\LanguageServerProtocol\Position;
+use Phpactor\LanguageServerProtocol\TextEdit;
+use SebastianBergmann\Diff\Line;
+use SebastianBergmann\Diff\Parser;
```
dantleech, line 9: depending on another package here is probably OK, although in core Phpactor packages I would make this an interface and put the implementation in `Adapter/SebastianDiff` or similar.

dantleech (approved):
Left some more or less annoying comments, but looks good :+1:

lib/Extension/LanguageServerPhpCsFixer/Provider/PhpCsFixerDiagnosticsProvider.php
lumnn: image

This is how that could look with client I use. For this PR this looks exactly same, but without the link

lib/Extension/LanguageServerPhpCsFixer/Provider/PhpCsFixerDiagnosticsProvider.php
lumnn: Extracted few methods which broke code into smaller readable or not repetitive tasks. Is that enough or is there a need for more?

lib/Extension/LanguageServerPhpCsFixer/Provider/PhpCsFixerDiagnosticsProvider.php
lumnn: I'd do throw/exit early possible, but in this scenario I think it won't make sense.

The whole loop works by finding an appropriate case and continue on the loop and it's dependent on previous entries of the looped array.
Meaning that all successful scenarios are early the block, while potential incorrect behaviour is left as the last one - Making this change would mean we have few correct cases, an incorrect one followed by one more correct case

That's my reasoning behind it, but not too attached to it. If you prefer to reverse the if, then it's fine and I'll update that

lib/Extension/LanguageServerPhpCsFixer/Util/DiffToTextEditsConverter.php
lumnn: I've left it without interface and adapter. I'd rather expect php-cs-fixer changes to output than the changes in that `diff` package, which is one of most popular ones on packagist.org 

Unless there are other reasons for that?

lumnn:
I updated PR based on your input and updated below your comments.
I do consider PR beeing ready to merge now.

lib/Extension/LanguageServerPhpCsFixer/Util/DiffToTextEditsConverter.php
dantleech: As said, it's probably OK. It's just hygienic, we shouldn't depend on random packages (regardless of how popular they are) in "core" code. Also ... is the Diff package a non-dev dependency in composer?

lib/Extension/LanguageServerPhpCsFixer/Model/PhpCsFixerProcess.php
```diff
@@ -0,0 +1,114 @@
...
+        });
+    }
+
+    /**
+     * @return Promise<Process>
+     */
+    public function run(string ...$args): Promise
+    {
+        return call(function () use ($args) {
+            $process = new Process([$this->binPath, ...$args], null, $this->env);
```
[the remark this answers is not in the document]
lumnn: Found that. The reason is xdebug being enabled. Then, I believe php-cs-fixer restarts and somehow `PHP_CS_FIXER_IGNORE_ENV` is not working as expected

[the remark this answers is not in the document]
lumnn: Potentially I could change default for `language_server_php_cs_fixer.env` to include `XDEBUG_MODE=off` to ensure php-cs-fixer will always start without xdebug?

lib/Extension/LanguageServerPhpCsFixer/Provider/PhpCsFixerDiagnosticsProvider.php
dantleech: What I'm saying is this is an arbitrary identifier. AFAIK it has no semantic meaning? All the existing kinds are `refactor.blah`, or `quickfix.blah`. But maybe you know more than I do here?

dantleech:
> Therefore I added source.fixAll.phpactor.phpCsFixer that would allow other fixAll commands, and also let LSP clients to control which of them should be applied (for example on save)

hm, thanks for looking into that. in this case we should probably update all the other code actions baesd on the VS Code conventions: https://code.visualstudio.com/api/references/vscode-api#CodeActionKind

I created https://github.com/phpactor/phpactor/issues/2063

Are you using VS code? As I see this:
image
which is too verbose really (and it shows all the examples)

dantleech:
ok, updating it and it looks better, but no qucik fixes?

image

lumnn:
> no qucik fixes?

I did remove them as I got influenced by Sublime Text integration that shows duplicates available code actions (in context menu) for me (well, they're showing different symbols for source action and quickfix, but I think it's a LSP client implementation to handle that better, not LSP server to workaround the client problems)

Looking into VSCode it's probably better to include them along `source.fixAll` as over there that doesn't seem to be duplicated.
Possibly worth checking in vim/nvim?

I also thought that `quickfix` may not be appropriate as it fixes whole document, not just the problem that it shows against. To work arround that I can put a title like "Fix this and all other PHP CS Fixer problems"

In that scenario, there will be 3 different ways of executing formatting:

1. Document Formatting Request 
  - I think we can provide use other formatters along php-cs-fixer? Not sure though how Phpactor handles multiple formatters
2. `quickfix` code action that shows against diagnostics
3. `source.fixAll` code action (that shows on demand or is used via "on save" trigger configuration in certain cases). `source.fixAll.phpactor.phpCsFixer` is exactly used this means:
  - We can add more fixAll providers under `source.fixAll.phpactor` namespace
  - User can use any part of namespace as an action on save, which will trigger all or selected phpactor formatters on save

It does feel quite a few options to get the fixing/formatting done, but as long as LSP clients we can test against are showing them in an unobtrusive manner, then it should be fine.

I'll implement those changes today, unless you think of any different approach?

dantleech:
yeah, NVIM doesn't care about these "kinds" at all as far as I can tell, hence my confusion.

In that case `source.fixAll` and the document formatting routes are fine, let's not have `quickfix`

lib/Extension/LanguageServerPhpCsFixer/Model/PhpCsFixerProcess.php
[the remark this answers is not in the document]
lumnn: Sorted

dantleech:
The LSP protocol changed under your feet: `ApplyWorkspaceEditResponse` was renamed to `ApplyWorkspaceEditResult` otherwise there are some more failures, I guess all quite easy to address though.

dantleech:
you can run `composer integrate` before pushing to ensure the tests (should pass). I think you still need to `composer update`?

lib/Extension/LanguageServerPhpCsFixer/Util/DiffToTextEditsConverter.php
dantleech: it's a direct dependency of Phpactor :+1: in general though I'd rather put things that use 3rd parties behind an interface, so `Adapter/SebastianDiff/DiffToTextEditConverter` and `DiffToTextEditsConverter` is then the interface.

but :shrug:

dantleech (approved):

dantleech:
Happy to merge if you're happy.

lumnn:
I believe it's all is good for merge.

dantleech:
Great addition, thanks

PR #1977, opened 2022-12-04
Fixing the string resolver

### LINKED ISSUE #1969: the quick fix 'Extract constant' does not support heredoc strings
Hi,

It looks like the quick fix action 'Extract constant' does not support heredoc strings properly.

Here is a reproducer:

```php
<?php

declare(strict_types=1);

final class MyClass
{
    public function __invoke(): void
    {
        $string = <<<EOT
        hello
        EOT;

        echo $string;
    }
}
```

Hit the 'Extract constant' quick fix, and you get this code:

```php
<?php

declare(strict_types=1);

final class MyClass
{
    const NEW_CONSTANT = '';

    public function __invoke(): void
    {
        $string = self::NEW_CONSTANT;

        echo $string;
    }
}
```

Context: branch 'master' of phpactor, used by vscode-phpactor, PHP 8.1.11.

mamazu:
The string resolver didn't handle the case where the string literal was a heredoc string.

Fixes #1969

### DIFF
--- a/lib/WorseReflection/Core/Inference/Resolver/StringLiteralResolver.php
+++ b/lib/WorseReflection/Core/Inference/Resolver/StringLiteralResolver.php
@@ -36,17 +36,20 @@ class StringLiteralResolver implements Resolver
     private function getStringContentsText(StringLiteral $node): string
     {
         $children = $node->children;
+        if (is_array($children)) {
+            $children = $children[0];
+        }
+
         if ($children instanceof Token) {
             $value = (string)$children->getText($node->getFileContents());
             $startQuote = substr($node, 0, 1);
 
-            if ($startQuote === '\'') {
-                return rtrim(substr($value, 1), '\'');
-            }
-
-            if ($startQuote === '"') {
-                return rtrim(substr($value, 1), '"');
-            }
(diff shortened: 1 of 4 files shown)


lib/WorseReflection/Core/Inference/Resolver/StringLiteralResolver.php
```diff
@@ -36,17 +36,20 @@ public function resolve(NodeContextResolver $resolver, Frame $frame, Node $node)
     private function getStringContentsText(StringLiteral $node): string
     {
         $children = $node->children;
+        if (is_array($children)) {
+            $children = $children[0];
+        }
+
```
mamazu, line 42: For some reason the node's children is an array for when it's a here doc comment. I don't know why though.

dantleech: yes it's a bit broken at the parser level. can we also check that `[0]` exists?

dantleech:
nice, thanks :) could you also update the CHANGELOG with the fix?

dantleech (approved):

dantleech:
thanks!

PR #1980, opened 2022-12-05
Adding more types to phpactor

mamazu:


lib/WorseReflection/Core/Type/IntLiteralType.php
```diff
@@ -41,7 +39,7 @@ public function negative(): NumericType
         return new self(-$this->value());
     }
 
-    public function withValue($value)
+    public function withValue($value): IntLiteralType
```
dantleech, line 42: this is safe as long as the class is final, and it's not

lib/WorseReflection/Core/Type/IntLiteralType.php
```diff
@@ -18,10 +18,8 @@ public function __toString(): string
         return (string)$this->value;
     }
 
-    /**
-     * @return int
-     */
-    public function value()
+
+    public function value(): int
```
dantleech, line 22: i'm confused by both of htese changes can child classes narrow the type proided in an interface (both of these methods come from the interface)

dantleech: hmm, ok, since 7,4 and co-variance this is fine :shrug: https://wiki.php.net/rfc/covariant-returns-and-contravariant-parameters

dantleech (approved):

dantleech:
:bow:

Issue #1931, opened 2022-10-22
Cannot rename interface


dantleech:
When trying to rename an `interface Foo` there is an error "Foo is not a class, it is an interface". Probably something is doing `reflectClass` insteadof `reflectClassLike`

dantleech:
Was fixed.

PR #1982, opened 2022-12-05
fix: renaming interfaces

nataneb32:
[fixing this issue](https://github.com/phpactor/phpactor/issues/1931)

### DIFF
--- a/lib/Rename/Adapter/WorseReflection/WorseNameToUriConverter.php
+++ b/lib/Rename/Adapter/WorseReflection/WorseNameToUriConverter.php
@@ -18,7 +18,7 @@ class WorseNameToUriConverter implements NameToUriConverter
     public function convert(string $className): TextDocumentUri
     {
         try {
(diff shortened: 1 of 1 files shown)


lib/Rename/Adapter/WorseReflection/WorseNameToUriConverter.php
```diff
@@ -18,7 +18,7 @@ public function __construct(private Reflector $reflector)
     public function convert(string $className): TextDocumentUri
     {
         try {
-            $uri = $this->reflector->reflectClass($className)->sourceCode()->uri();
+            $uri = $this->reflector->reflectClassLike($className)->sourceCode()->uri();
```
dantleech, line 21: would be great to add a test for this if you can

dantleech:
thanks, feel free to PR a  test, but as there are no existing tests for this :open_mouth: fine if you don't

PR #1985, opened 2022-12-09
Complete inside class constant declaration

### LINKED ISSUE #1978: Class completion does not work inside a constant value
image

przepompownia:
Closes #1978

### DIFF
--- a/lib/Completion/Bridge/TolerantParser/CompletionContext.php
+++ b/lib/Completion/Bridge/TolerantParser/CompletionContext.php
@@ -43,10 +43,6 @@ class CompletionContext
             return false;
         }
 
-        if ($parent instanceof ConstElement) {
-            return true;
-        }
-
         if (self::classMembersBody($node)) {
             return false;
         }
@@ -54,6 +50,7 @@ class CompletionContext
         return
             $parent instanceof Expression ||
             $parent instanceof StatementNode ||
+            $parent instanceof ConstElement ||
             $parent instanceof ArrayElement // yield;
         ;
     }
@@ -140,6 +137,10 @@ class CompletionContext
         }
 
         if (null === $node->parent) {
+            return false;
+        }
+
+        if ($node->parent instanceof ConstElement) {
             return false;
         }
 
--- a/lib/Completion/Bridge/TolerantParser/CompletionContext.php
+++ b/lib/Completion/Bridge/TolerantParser/CompletionContext.php
@@ -184,13 +184,13 @@ class CompletionContext
             return false;
         }
 
-        $methodDeclaration = $nodeBeforeOffset->getFirstAncestor(MethodDeclaration::class);
+        $memberDeclaration = $nodeBeforeOffset->getFirstAncestor(MethodDeclaration::class, ConstElement::class);
 
-        if (!$methodDeclaration) {
+        if (!$memberDeclaration) {
             return true;
         }
 
-        if ($methodDeclaration->getEndPosition() < $node->getStartPosition()) {
+        if ($memberDeclaration->getEndPosition() < $node->getStartPosition()) {
             return true;
         }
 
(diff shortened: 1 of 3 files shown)


przepompownia:
I'm not sure if it's the correct logic but first tests suggest that it works.

lib/Completion/Tests/Integration/Bridge/TolerantParser/ReferenceFinder/ExpressionNameCompletorTest.php
```diff
@@ -96,6 +96,16 @@ public function provideComplete(): Generator
                 ]
             ]
         ];
+
+        yield 'inside class constant declaration' => [
+            '<?php class Foobar { private const FOO; private const BAR = self::F<>  }', [
+                [
+                    'type'              => Suggestion::TYPE_CONSTANT,
+                    'name'              => 'self::FOO',
```
przepompownia, line 104: Please check if the expectation is correct.

lib/Completion/Bridge/TolerantParser/CompletionContext.php
```diff
@@ -42,6 +43,10 @@ public static function expression(?Node $node): bool
             return false;
         }
 
+        if ($parent instanceof ConstElement) {
```
dantleech, line 46: think we can move this to the similar expressions in `return` below?

dantleech: or is it triggered by classMembersBody somehow?

lib/Completion/Tests/Integration/Bridge/TolerantParser/ReferenceFinder/ExpressionNameCompletorTest.php
dantleech: why do you suspect it _not_ to be correct?

przepompownia: At the first I started with `name => 'FOO'` - the whole `'self::FOO'` caught my attention.

dantleech: hmm, yeah that doesn't seem quite right, did you test it in the editor?

przepompownia: Yes,

https://user-images.githubusercontent.com/11404453/206718275-a8e0bc8d-6d79-4524-80ba-277634802213.mp4

but I found that the new test case checks something that works already on master :grin: so I need to add the proper case before refactoring discussed in the other comment.

przepompownia:
It's still not what I need:
image

dantleech:
What are you missing ?

dantleech:
it's a mine field, good luck!

przepompownia:
https://user-images.githubusercontent.com/11404453/206793817-7a3d8915-d842-46bc-ac12-c3c973875c5a.mp4

Now it's time to return to the question where `$parent instanceof ConstElement` should be placed. Including this condition in `classMembersBody` requires repeat the query again in `expression()`.

dantleech:
I think it's fine, its not a "query" but just an `instanceof` and it's true, it's not a direct child of the class members _bod but we should complete expressions on ConstElements.

I've just spent the past 30m trying to stop `function` and `const` being completed:

image

but I think we can merge this in it's current state as an improvement?

przepompownia:
I agree that at the moment almost accurate suggestion list is better than none and we could merge this PR. 

Independently it would be good to keep in `master` PHPUnit test cases that fails (but should/could pass, like that shown by you on the picture below) but marked to exclude them with `--exclude-group` by CI. What do you think about such approach?

lib/Completion/Tests/Integration/Bridge/TolerantParser/ReferenceFinder/ExpressionNameCompletorTest.php
dantleech: it is an artificial suggestion:

image

dantleech (approved):

dantleech:
thanks,I updated the changelog and changed the test to simulate something more realistic...

PR #1807, opened 2022-07-20
[WIP] Laravel Blade

haringsrob:
This pr is just to show the concept.

It uses the output of code written in https://github.com/haringsrob/laravel-dev-tools (a lsp on its own).

PR #1991, opened 2022-12-20
[WIP] Add laravel extension

haringsrob:
Todo:

- [ ] Route autocomplete `route('<>')`
- [ ] Service container make
- [ ] View autocomplete `view('<>')`

### DIFF
--- a/lib/Extension/Laravel/Adapter/Laravel/LaravelContainerInspector.php
+++ b/lib/Extension/Laravel/Adapter/Laravel/LaravelContainerInspector.php
@@ -1,0 +1,37 @@
+<?php
+
+namespace Phpactor\Extension\Laravel\Adapter\Laravel;
+
+use Phpactor\WorseReflection\Core\TypeFactory;
+use Phpactor\WorseReflection\Core\Type\ClassType;
+
+class LaravelContainerInspector
+{
+    public function __construct(private string $xmlPath)
+    {
+    }
+
+    public function service(string $id): ?ClassType
+    {
+        foreach ($this->services() as $short => $service) {
+            if ($short === $id || $service === $id) {
+                return TypeFactory::fromString($service);
+            }
+        }
+        return null;
+    }
+
+
+    public function services(): array
+    {
+        $serviceCache = require $this->xmlPath;
+
+        $services = [];
+
+        foreach ($serviceCache['providers'] as $provider) {
+            $services[$provider] = $provider;
+        }
+
+        return $services;
+    }
+}
--- a/lib/Extension/Laravel/Completor/LaravelContainerCompletor.php
+++ b/lib/Extension/Laravel/Completor/LaravelContainerCompletor.php
@@ -1,0 +1,105 @@
+<?php
+
+namespace Phpactor\Extension\Laravel\Completor;
+
+use Generator;
+use Microsoft\PhpParser\Node;
+use Microsoft\PhpParser\Node\Expression\ArgumentExpression;
+use Microsoft\PhpParser\Node\Expression\CallExpression;
+use Microsoft\PhpParser\Node\Expression\ScopedPropertyAccessExpression;
+use Microsoft\PhpParser\Node\QualifiedName;
+use Microsoft\PhpParser\Node\StringLiteral;
+use Phpactor\Completion\Bridge\TolerantParser\TolerantCompletor;
+use Phpactor\Completion\Core\Suggestion;
+use Phpactor\Extension\Laravel\Adapter\Laravel\LaravelContainerInspector;
+use Phpactor\TextDocument\ByteOffset;
+use Phpactor\TextDocument\TextDocument;
+use Phpactor\WorseReflection\Core\TypeFactory;
+use Phpactor\WorseReflection\Core\Util\NodeUtil;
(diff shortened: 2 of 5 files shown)


lib/Completion/Bridge/TolerantParser/WorseReflection/WorseClassMemberCompletor.php
```diff
@@ -76,10 +76,10 @@ public function complete(Node $node, TextDocument $source, ByteOffset $offset):
         $reflectionOffset = $this->reflector->reflectOffset($source, $memberStartOffset);
 
         $symbolContext = $reflectionOffset->symbolContext();
-        $type = $symbolContext->type();
+        $symbolType = $symbolContext->type();
```
dantleech, line 79: would be idimatic to just call this `type` i think is it ambiguous?

lib/Extension/Laravel/Adapter/Laravel/LaravelContainerInspector.php
```diff
@@ -0,0 +1,37 @@
...
+            }
+        }
+        return null;
+    }
+
+
+    public function services(): array
+    {
+        if ([] === $this->services) {
+            $output = shell_exec("php $this->executablePath container $this->projectRoot");
```
dantleech, line 29: preferred to use the Symfony `Process` class

dantleech:
nice, feel free to concentrate on one feature at a time, that way we can ensure we're adding the necessary tests etc.

lib/Extension/Laravel/Adapter/Laravel/LaravelContainerInspector.php
dantleech: also wonder if there is a way to inspect the container that doesn't involve an external binary?

dantleech: or if something like [this](https://github.com/barryvdh/laravel-ide-helper#phpstorm-meta-for-container-instances) is more idiomatic? do most devs use those generators?

haringsrob: I intentionally did not go for that direction as it enforces a project dependency.

My package does use that generator and can generate it in the vendor folder.

With a lot of code I think we can figure it out. However, I personally would like to not focus on that for now.

lib/Completion/Bridge/TolerantParser/WorseReflection/WorseClassMemberCompletor.php
haringsrob: I changed this because $type is used in the loop `as`. It could be a bug if it at some point is being used after the foreach.

haringsrob:
hey @dantleech I am about to pick this back up.

Regarding the external binary. It is for sure doable without I guess, but for me it would be more difficult at this point as I simply am not known enough to the phpactor codebase. Can keep using it for now?

Also, maybe I better split this work into a separate extension if that is still a possibility? Some thing I for sure need to do in the future is to have the ability to inject types into a file (see https://github.com/phpactor/phpactor/pull/1807)

Waiting for your advice.

lib/Extension/Laravel/Adapter/Laravel/LaravelContainerInspector.php
haringsrob: Switched to Process

haringsrob:
Lot's of progress was made, may want to add some tests as well so I can avoid myself from breaking but also add failing tests for things I expect to work :), this however, would need a full laravel installation to work with, any suggestions?

composer.json
```diff
@@ -30,7 +30,8 @@
         "phpactor/tolerant-php-parser": "dev-main",
         "phpactor/map-resolver": "^1.5.0",
         "webmozart/assert": "^1.11",
-        "webmozart/path-util": "^2.3"
+        "webmozart/path-util": "^2.3",
+        "illuminate/view": "^8.83"
```
haringsrob, line 34: - [ ] I refactored this out already so no need anymore I think.

haringsrob: It is in fact needed as I managed to use fully compiled blade for lsp resolving using a slightly modified compiler.

dantleech:
please note that this branch should be treated as an experiment, it will be very difficult to review / merge this as it is, we should start creating PRs offering one feature at a time (or at least PRs that are of a small enough size)

composer.json
dantleech: i don't really feel comfortable depending on having a hard dependency on a framework component

haringsrob: Pretty sure i can copy just the needed parts over without all the rest

haringsrob:
> impressive work so far :tada: but please note that this branch should be treated as an experiment, it will be very difficult to review / merge this as it is, we should start creating PRs offering one feature at a time (or at least PRs that are of a small enough size). Any change to the Phpactor internals should be offered separately in individual PRs.

Yes totally agree. It is however very hard to see of things work if I would have contributed smaller parts. I think one I have worked with it for a few weeks and things are stable I can start making pr's with tests of smaller chunks

vivere-dally:
Hello! Sorry for bothering, but are there any updates on this issue?

PR #1998, opened 2022-12-27
Complete any class name inside attribute

przepompownia:
It only allows what the title says. Classes that are not attributes are also suggested with this change.

### DIFF
--- a/lib/Completion/Bridge/TolerantParser/CompletionContext.php
+++ b/lib/Completion/Bridge/TolerantParser/CompletionContext.php
@@ -6,6 +6,7 @@ use Microsoft\PhpParser\ClassLike;
 use Microsoft\PhpParser\MissingToken;
 use Microsoft\PhpParser\Node;
 use Microsoft\PhpParser\Node\ArrayElement;
+use Microsoft\PhpParser\Node\Attribute;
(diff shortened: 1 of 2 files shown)


przepompownia:
@dantleech what do you think about merging it before working on #1999?

lib/Completion/Bridge/TolerantParser/CompletionContext.php
```diff
@@ -51,6 +52,7 @@ public static function expression(?Node $node): bool
             $parent instanceof Expression ||
             $parent instanceof StatementNode ||
             $parent instanceof ConstElement ||
+            $parent instanceof Attribute ||
```
dantleech, line 55: it's not valid to have an expression in an attribute though... !?

przepompownia: Maybe it's a hack.  Would that mean that it should be checked separately, especially in some new Completor mentioned in #1999?

dantleech: yes, probably there should be a `attributeClass(Node $node)` method here - but equally can be done in the new `Completor`

przepompownia:
Let's focus on #2001 instead.

PR #2001, opened 2022-12-27
Complete attribute names

### LINKED ISSUE #1999: Complete attribute classes
Given #1998 and previous examples with class constant completion (#1979,  #1985) I started to find how to restrict the suggestion list to `Attribute`s only. This led me to the question which [Completor](https://github.com/phpactor/phpactor/blob/master/lib/Completion/Core/Completor.php) should be used to it: any existing or some new?

przepompownia:
#1999


przepompownia:
At the moment it seems to do the same as #1998 

Do we need to change `NameSearcher` implementations to filter attribute classes?

lib/Completion/Bridge/TolerantParser/ReferenceFinder/AttributeCompletor.php
```diff
@@ -0,0 +1,57 @@
...
+{
+    public function __construct(
+        private NameSearcher $nameSearcher,
+        private DocumentPrioritizer $prioritizer
+    ) {
+    }
+
+    public function complete(Node $node, TextDocument $source, ByteOffset $offset): Generator
+    {
+        if (!CompletionContext::attribute($node)) {
```
przepompownia, line 29: Is it needed if `AttributeQualifier` checked it before?

dantleech:
> Is it needed if AttributeQualifier checked

you can probably ignore the AttributeQualifier

> Do we need to change NameSearcher implementations to filter attribute classes?

:thinking: it would be quite expensive to parse all the results to see if they are attributes, so we might want to index them. 

How we do that is a different question - currently `NameSearcherType::` is used which is `class|interface|trait|enum`, and attributes are already a `class`, but I guess at this rough level of abstraction having `attribute` as a type might be OK.

Then the indexer would either need to store a boolean flag `is_attribute` or the set of attributes on a given class.

lib/Completion/Bridge/TolerantParser/ReferenceFinder/AttributeCompletor.php
dantleech: you could use the "qualifier" or remove it and use this only. I don't think there is any advantage in using the `Qualfiier` (IIRC it was to allow the "qualifiying" logic to be shared, but that's better done with `CompletionContext` anyway).

lib/Completion/Bridge/TolerantParser/ReferenceFinder/AttributeCompletor.php
```diff
@@ -0,0 +1,57 @@
...
+            return true;
+        }
+
+        $search = $node->getText();
+        $type = NameSearcherType::CLASS_;
+
+        foreach ($this->nameSearcher->search($search, $type) as $result) {
+            if (!$result->type()->isClass()) {
```
dantleech, line 37: it's already a "class" as you specified that as the "type"

lib/Completion/Bridge/TolerantParser/ReferenceFinder/AttributeCompletor.php
przepompownia: It's nicer for me to work separated pieces than on one bigger class. If it were up to me, I'd leave `AttributeQualifier`. Let me know which way to get rid of.

dantleech: i would prefer not using the "qualifier" - the code is still "segregated", and it's more explicit with a guard condition.

przepompownia:
Similarly like in #1999 what do you think about merging it as is (with some better title) and working on indexing and filtering separately?

przepompownia:
Thank you. It's hard to say how long the next step will take, especially for me.

dantleech:
I'll see if I have time to look at it

PR #2026, opened 2023-01-03
perf(psalm): improve psalm diagnostics performances

gbprod:
I've lately moved from [null-ls](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/b287e9d124037a7ea5de8e3368a3b75274cec8b4/lua/null-ls/builtins/diagnostics/psalm.lua) to phpactor for psalm diagnostics and I found it slower...

After investigating, I've found that the `PsalmProcess` class doesn't pass the filename to the command, then, psalm always parse all the project instead of just the file it should analyse.

In my case, It reduces the execution time from 5 to 1.8 seconds.

Then, I've also noticed that `--no-cache` directive is passed to the command, and it's not present in the  [null-ls](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/b287e9d124037a7ea5de8e3368a3b75274cec8b4/lua/null-ls/builtins/diagnostics/psalm.lua) command. I've removed it and it reduces the execution time from 1.8 to 0.9 seconds.

### DIFF
--- a/lib/Extension/LanguageServerPsalm/LanguageServerPsalmExtension.php
+++ b/lib/Extension/LanguageServerPsalm/LanguageServerPsalmExtension.php
@@ -19,6 +19,7 @@ class LanguageServerPsalmExtension implements OptionalExtension
 {
     public const PARAM_PSALM_BIN = 'language_server_psalm.bin';
     public const PARAM_PSALM_SHOW_INFO = 'language_server_psalm.show_info';
+    public const PARAM_PSALM_USE_CACHE = 'language_server_psalm.use_cache';
 
     public function load(ContainerBuilder $container): void
     {
@@ -38,10 +39,11 @@ class LanguageServerPsalmExtension implements OptionalExtension
             $binPath = $container->get(FilePathResolverExtension::SERVICE_FILE_PATH_RESOLVER)->resolve($container->getParameter(self::PARAM_PSALM_BIN));
             $root = $container->get(FilePathResolverExtension::SERVICE_FILE_PATH_RESOLVER)->resolve('%project_root%');
             $shouldShowInfo = $container->getParameter(self::PARAM_PSALM_SHOW_INFO);
+            $useCache = $container->getParameter(self::PARAM_PSALM_USE_CACHE);
 
             return new PsalmProcess(
                 $root,
-                new PsalmConfig($binPath, $shouldShowInfo),
+                new PsalmConfig($binPath, $shouldShowInfo, $useCache),
                 LoggingExtension::channelLogger($container, 'PSALM')
             );
         });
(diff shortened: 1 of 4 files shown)


dantleech:
> Then, I've also noticed that --no-cache directive is passed to the command

This was necessary at the time as the cache didn't invalidate properly. That may have improved.

Possibly not passing the filename was necessary too.

Can you let me know if it works consistently well?

gbprod:
> This was necessary at the time as the cache didn't invalidate properly. That may have improved.

I don't know if it have been improved, I will try to work without `--no-cache` and look for inconsistencies, but I've used for a long time [null-ls](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/b287e9d124037a7ea5de8e3368a3b75274cec8b4/lua/null-ls/builtins/diagnostics/psalm.lua) and I didn't have any problems...

> Possibly not passing the filename was necessary too.

I don't understand why not passing the filename should be necessary, the results are the same and performances are greatly improved. 
 
I will try to use psalm like this for some times and I'll give you some feedbacks.

PS: I'm running the latest version of psalm.

dantleech:
> I don't understand why not passing the filename should be necessary

There were lots of issues when I implemented this years ago regarding the cache, it's probably better now.

gbprod:
> > I don't understand why not passing the filename should be necessary
> 
> There were lots of issues when I implemented this years ago regarding the cache, it's probably better now.

I'm using this branch for 7 days without detecting any issue with psalm diagnostics. 
I don't know if you want to merge like this or if it will be safer to keep the `--no-cache` option ?

dantleech:
If you say it's good, let's do it.

dantleech:
but maybe as a configuration option where the default is "cache enabled" ?

gbprod:
I've updated the PR to introduce a configuration option to enable/disable cache and it works :)

lib/Extension/LanguageServerPsalm/LanguageServerPsalmExtension.php
```diff
@@ -53,10 +55,12 @@ public function configure(Resolver $schema): void
         $schema->setDefaults([
             self::PARAM_PSALM_BIN => '%project_root%/vendor/bin/psalm',
             self::PARAM_PSALM_SHOW_INFO => true,
+            self::PARAM_PSALM_USE_CACHE => true,
         ]);
         $schema->setDescriptions([
             self::PARAM_PSALM_BIN => 'Path to pslam if different from vendor/bin/psalm',
             self::PARAM_PSALM_SHOW_INFO => 'If infos from psalm should be displayed',
+            self::PARAM_PSALM_USE_CACHE => 'Does psalm should use cache (see `--no-cache` option)',
```
dantleech, line 63: Change this to something like `If Psalm's cache should be used`

PR #2036, opened 2023-01-06
Added ext-posix extension as required

dacianb:
Added ext-posix as required in the composer.json


dantleech:
not sure why the other test runs were ok, but you will need to do `composer update --lock` to update the `composer.lock` file.

PR #2038, opened 2023-01-07
Adding more type information to the UseStatementUpdater

mamazu:


dantleech:
Already being taken care of here: https://github.com/phpactor/phpactor/pull/2033/files#diff-536119380389e51b201271f6de1590a99451560e834e1110d15248a1001e0122R168

PR #2051, opened 2023-01-20
Complete inside `match` expression

przepompownia:


przepompownia:
I wanted to add also
```php
        yield 'expression inside match arm' => [
            '<?php class Foobar { public function x() { match (true) { 1 => $th<> }; }}', [
                [
                    'type'              => Suggestion::TYPE_VARIABLE,
                    'name'              => '$this',
                    'short_description' => '$this',
                ]
            ],
        ];
```
but have no idea what to yield in `createTolerantCompletor` instead of `NameSearchResult` (which cannot be created with `variable` type).

dantleech:
cool, thanks

PR #2068, opened 2023-01-31
Improved variable name by correcting typo

still-dreaming-1:
This change could be helpful if anyone is searching for the text "psalm".


dantleech:
Thanks!

PR #2070, opened 2023-02-01
Relative name completion

dantleech:
This PR will try and solve the completion of `Relative\Names`, this is especically useful for attributes/annotations such as `@ORM\Foobar()`

image

- Makes `\\` a trigger char

TODO:

- [x] Support searching and completion of relative names for expressions.
- [x]  Support namespace completion
- [x] Support only completing the children of the relative name
- [x] Support attributes (why doesn't this just work already?)


lib/Completion/Bridge/TolerantParser/ReferenceFinder/ExpressionNameCompletor.php
```diff
@@ -35,7 +38,12 @@ public function complete(Node $node, TextDocument $source, ByteOffset $offset):
             return true;
         }
 
-        $suggestions = $this->completeName($node, $source->uri(), $node);
+        $name = $node->__toString();
+        if ($node instanceof QualifiedName && str_contains($name, '\\')) {
+            $name = '\\' . $node->getResolvedName()->__toString();
```
dantleech, line 43: is this really necessary, also use the new `NameUtil` here

Issue #2063, opened 2023-01-30
Update Phpactor code action "kinds"


dantleech:
Update code actions to use the `kinds` defined here: https://code.visualstudio.com/api/references/vscode-api#CodeActionKind

PR #2074, opened 2023-02-05
Adding new CodeActionKinds to the language server

mamazu:


lib/Extension/LanguageServerCodeTransform/CodeAction/ExtractConstantProvider.php
```diff
@@ -4,6 +4,7 @@
 
 use Amp\CancellationToken;
 use Amp\Promise;
+use LanguageServerProtocol\CodeActionKind;
```
dantleech, line 7: this is the "wrong" the one phpactor uses is `phpactor/language-server-protocol` this one is an incidental dependency from Psalm.

AFAIK the code action "kinds" are not standardised and that class is perhaps specific to the felixbecker language server. We couild perhaps add a similar class in `lib/Extension/LanguageServer` if it adds benefit.

dantleech:
As commented we should not use the LanguageServerProtocol package as it is not "ours". Also - I'm not sure if you are aware how IDEs filter these actions, as I am unaware how the filtering is done or what the optimal "kinds" are. Maybe @lumnn can point out how they are used in Code (if he didn't already).

mamazu:
You are right that those code actions are from an external dependency. I could also duplicate the class into the phpactor/language-server-protocol if need be. The reason I used the constants from felixbecker was that those are exactly the classes that you pointed out in the documentation here: #2063

lumnn:
Based on [spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeActionKind) and testing with both VSCode and Sublime Text

**`quickfix`** - Shows instantly as a lightbulb (vscode) or button with action name (st) when caret gets into a region where action is available. They do also show in hover popups.

VSCode
Screenshot from 2023-02-06 14-22-22
Screenshot from 2023-02-06 13-41-42

ST (both visible)
Screenshot from 2023-02-06 13-38-58

**`refactor`** - They show on demand in context menu and shows refactor actions related to where carret is. In vscode they seem to grouped in 4 sections (Rewrite, Inline, Extract, More actions), so I believe that's how `refactor.rewrite`, `inline` and `extract` are used. ST has no grouping

VSCode
Screenshot from 2023-02-06 13-58-37
Screenshot from 2023-02-06 14-18-54

ST
image

**`source`** - Actions for whole source file. Both VSCode and ST allow to get list of them on demand. Some of them seem to have special extra meaning:

- `source.organizeImports` - VSCode shows keyboard shortcut for that
- `source.fixAll` - VSCode and ST allows to specify code actions to execute when file is saved. This seems to be sometimes used as actions to do on file save. As opposed to [Document Formatting Request](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_formatting) it seems like you can get finer control over what actions are executed. In VSCode Document Formatting also requires you to choose the desired formatter (if there are multiple language servers with this capability for same language)

VScode
Screenshot from 2023-02-06 14-30-55

ST
Similar to refactorings

dantleech:
> Based on spec

:facepalm: not sure how I missed that

we even have it here:

image

mamazu:
Where did you find it. It's not part of the `phpactor/language-server-protocol` package.

dantleech:
> Where did you find it? It's not part of the phpactor/language-server-protocol package.

oh fun, it's in the latest version https://github.com/phpactor/language-server-protocol/blob/3.17/src/CodeActionKind.php

I have a feeling there are lots of B/C breaks

mamazu:
Should I try to update the protocol package in this MR or make a new one?

dantleech:
New one, it will likely need to be updated in the language server package first

mamazu:
Oddly the `phpactor/language-server` and `phpactor/language-server-protocol` don't have an update when running composer update on those packages. Even `composer outdated` don't show that there are updates for this package.

dantleech:
It's not merged yet: https://github.com/phpactor/language-server-protocol/pull/16

dantleech:
I'll see if I can regenerate the LSP protocol based on the latest VS code version, and then make it work with Phpactor. Until then feel free to copy it from that branch and put it in `Phpactor\LanguageServerProtcol\CodeActionKinds.php`

dantleech:
I've regenerated the protocol (looks like 3.17 is the latest version) and made a PR for the language server which updates it there: https://github.com/phpactor/language-server/pull/53

the next step wil be a PR using `phpactor/language-server@dev-protocol-3.17` and updating Phpactor to use it.

dantleech:
phpactor is using 3.17 now, so this should be good to rebase :)

mamazu:
Well we have to remove the old kinds in this merge request now since they aren't valid anymore. But if it wasn't part of the standard anyways, I guess nobody should have been relying on that anyways.

lib/Extension/LanguageServerCodeTransform/CodeAction/CreateClassProvider.php
```diff
@@ -32,7 +30,6 @@ public function __construct(private Generators $generators, private Parser $pars
     public function kinds(): array
     {
         return [
-            self::KIND,
             CodeActionKind::QUICK_FIX,
```
dantleech, line 33: @lumnn should this be `quickfix.createclass` or just `quickfix`?

lib/Extension/LanguageServerCodeTransform/CodeAction/CreateClassProvider.php
```diff
@@ -19,11 +18,10 @@
 use Phpactor\LanguageServer\Core\CodeAction\CodeActionProvider;
 use Phpactor\LanguageServer\Core\Diagnostics\DiagnosticsProvider;
 use function Amp\call;
+use Phpactor\LanguageServerProtocol\CodeActionKind;
 
 class CreateClassProvider implements DiagnosticsProvider, CodeActionProvider
 {
-    public const KIND = 'quickfix.create_class';
```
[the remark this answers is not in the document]
mamazu: So just keeping the constant but removing it from the `kinds()` function? Because the kinds function can only return language server kinds.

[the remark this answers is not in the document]
dantleech: Referencing it from kinds. The protocol changed, it allows any arbitrary string now as per the spec (can still use codeactionkinds though)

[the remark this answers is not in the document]
mamazu: I don't know what you mean. Could you try to elaborate?

lib/Extension/LanguageServerCodeTransform/CodeAction/CreateClassProvider.php
dantleech: the `kind` is `<level1>.<level2>.<level3>` where `<level1>` is f.e. the value of `CodeActionKind::` - they define 1st level menus if you like.

just 

```
const KIND = CodeActionKind::QUICK_FIX;

// ...return [self::KIND];
```

and I don't know if we need more than that without actually seeing this in action.

dantleech: so this would be a menu like:

```
- quickfix:
    - create class
    - whatever
- refactor
    - extract method
    - whatever
   
```

dantleech: but without actually knowing how this information is used it's a bit like the blind leading the blind.

mamazu:
I have added the old code actions back. And by convention the "protocol based" code action strings are the first in the kind array so that anybody just looking at the first gets the right one.

dantleech:
Do we know what effect this configuration will have on the end user? 

- Should it be `CodeActionKind::REFACTOR_EXTRACT` or `CodeActionKind::REFACTOR_EXTRACT . '.method'` ?
- Do we need to kinds or one? Why do we need two?

I'd like to sit down with VS code and find the best "kinds".

mamazu:
Feel free to do that and report back what the best way to go about it. I don't know what to look for.

Issue #2071, opened 2023-02-05
Option to switch from complete constructor to make promoted properties?


dantleech:
e.g. in PHP 8:

```
    public function __construct(UserRepository $userRepository)
    {
    }
```

we currently provide code actions to "complete constructor" but, for me at least, it's _far_ more common that I have to go back and prefix `private` or `public`

provide a code action to "promote" constructor properties.

mamazu:
Could it be that we don't have a way to build promoted properties in the code right now? If not, I could try to create it.

dantleech:
Yes for sure, which is triggered exactly the same as `complete constructor`  but will promote properties instead of adding the properties + assignments.

dantleech:
Done

PR #2078, opened 2023-02-07
Adding option to add visibility to constructor properties

mamazu:
I am trying my hand at #2071 but I have found no way of setting the visibility of a property in the source code builder so I added this. However this doesn't work since it now doesn't create the constructor method anymore. I am probably missing something.


lib/CodeBuilder/Domain/Builder/ConstructorBuilder.php
```diff
@@ -0,0 +1,26 @@
...
+namespace Phpactor\CodeBuilder\Domain\Builder;
+
+class ConstructorBuilder extends MethodBuilder
+{
+    /**
+     * @param mixed $originalType
+     */
+    public function returnType(string $returnType, $originalType = null): MethodBuilder
+    {
+        trigger_error('You can not have a return type on a constructor', E_USER_NOTICE);
```
dantleech, line 12: throw an exception here - it's a developer error not a user error?

dantleech: but equally better to not have this method at all. 

for the sake of simplicity we could simply forget the semantics of `__construct` and add visibility on the `parameter`.

dantleech: (for example, look at `tolerant-parser/Node/Parameter` which has a visiblity token.

dantleech:
As noted above I think we can keep this simple and add the visiblity property to the "parameter builder"

lib/CodeBuilder/Tests/Adapter/UpdaterTestCase.php
```diff
@@ -1545,6 +1596,33 @@ public function hello(
...
+                    ->build()
+                ,
+                <<<'EOT'
+                    class Foo {
+                        public function changeMe(SomeOtherClass $dependency)
+                        {
+                        }
+                    }
+                    EOT
+        ];
```
mamazu, line 1624: I have no idea why this test is failing. Should I try to fix it or just leave it be for now?

dantleech:
```diff
--- a/lib/CodeBuilder/Adapter/TolerantParser/Updater/AbstractMethodUpdater.php
+++ b/lib/CodeBuilder/Adapter/TolerantParser/Updater/AbstractMethodUpdater.php
@@ -162,9 +162,7 @@ abstract class AbstractMethodUpdater
         // Adding new parameters to the mix
         foreach ($parameters as $parameter) {
             assert($parameter instanceof PhpactorParameter);
-            if (!isset($renderedParameters[$parameter->name()])) {
-                $renderedParameters[$parameter->name()] = $this->renderer->render($parameter>
-            }
+            $renderedParameters[$parameter->name()] = $this->renderer->render($parameter);
         }
 
         $startPosition = $methodDeclaration->openParen->getStartPosition();
```

makes it pass - it simply was not updating parameters. the question is do we feel confident that it won't mess up existing parameters by unconditionally replacing them?

mamazu: Since phpactor still doesn't know all the ins and outs of constuctor parameters (eg. `readonly` keyword) I wouldn't recommend that. But I have to admit that would also be one of the next things I will be working on.

dantleech: yep, so we don't support modifying parameters currently then, and this test can be removed :+1:

dantleech:
great, thanks @mamazu

PR #2079, opened 2023-02-11
Adding a code action to "Promote property"

mamazu:
References #2071

### DIFF
--- a/lib/CodeTransform/Adapter/WorseReflection/Refactor/WorsePromoteProperty.php
+++ b/lib/CodeTransform/Adapter/WorseReflection/Refactor/WorsePromoteProperty.php
@@ -1,0 +1,59 @@
+<?php
+
+namespace Phpactor\CodeTransform\Adapter\WorseReflection\Refactor;
+
+use Microsoft\PhpParser\Node\MethodDeclaration;
+use Microsoft\PhpParser\Node\Parameter;
+use Microsoft\PhpParser\Parser;
+use Phpactor\CodeBuilder\Domain\Builder\SourceCodeBuilder;
+use Phpactor\CodeBuilder\Domain\Code;
+use Phpactor\CodeBuilder\Domain\Prototype\Visibility;
+use Phpactor\CodeBuilder\Domain\Updater;
+use Phpactor\CodeTransform\Domain\Refactor\PromoteProperty;
+use Phpactor\TextDocument\ByteOffset;
+use Phpactor\TextDocument\TextDocument;
+use Phpactor\TextDocument\TextEdits;
+use Phpactor\WorseReflection\Bridge\TolerantParser\Reflection\ReflectionMethod;
+use Phpactor\WorseReflection\Core\Exception\NotFound;
+use Phpactor\WorseReflection\Core\SourceCode;
+use Phpactor\WorseReflection\Reflector;
+
+class WorsePromoteProperty implements PromoteProperty
+{
+    public function __construct(
+        private Reflector $reflector,
+        private Parser $parser,
+        private Updater $updater,
+    ) {
+    }
+
+    public function promoteProperty(TextDocument $document, ByteOffset $offset): TextEdits
+    {
+        $node = $this->parser->parseSourceFile($document->__toString())->getDescendantNodeAtPosition($offset->toInt());
+        $methodNode = $node->getFirstAncestor(MethodDeclaration::class);
+
+        if (!$methodNode instanceof MethodDeclaration) {
+            return TextEdits::none();
+        }
+
+        if ($methodNode->getName() !== '__construct') {
+            return TextEdits::none();
+        }
+
+        if(count($methodNode->parameters->children) === 0) {
+            return TextEdits::none();
+        }
+
+        $parameterNode = $node->getFirstAncestor(Parameter::class);
+        if (!$parameterNode instanceof Parameter::class) {
+            return TextEdits::none();
+        }
+
+        $parameter = $this->reflector->reflectNode($parameterNode, $methodNode->getStartPosition());
+        if (!$parameter instanceof ReflectionParameter) {
+            return TextEdits::none();
+        }
+
+        // Todo stuff
+    }
+}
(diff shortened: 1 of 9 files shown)


lib/CodeTransform/Adapter/WorseReflection/Refactor/WorsePromoteProperty.php
```diff
@@ -0,0 +1,54 @@
...
+            return TextEdits::none();
+        }
+
+        $constructor = $this->reflector->reflectNode($document, $node->getStartPosition());
+        if (!$constructor instanceof ReflectionMethod) {
+            return TextEdits::none();
+        }
+
+        // Todo stuff
+    }
```
mamazu, line 53: I need help with this function I am not sure what the best way is. Ideas that I have:

Get the `Parameter` node and then use the `getFirstAncestor` again to get the `MethodDeclaration` for name checking. And then I can still use the original node I got for all the manipulation. The problem is that we need the class name to let the source code updater know which class to update the constructor on. Any ideas?

dantleech: Can you not copy the existing transformer for complete constructor? 

https://github.com/phpactor/phpactor/blob/474eb3ccd7c58af5b7f83466dadaa88ce0eec6e6/lib/CodeTransform/Adapter/WorseReflection/Transformer/CompleteConstructor.php#L36-L40

This is very similar

mamazu: I just had a look at this. The thing is I would recommend to change the `CompleteConstructor`. It would make the command easier, generates less code and would also do the job.

The thing that I found with the `PromoteProperty` action is that if you already have a non promoted property with the property separately you also have to modify the method's body to remove the assignment of the property, which the `SourcecodeBuilder` doesn't support yet.

dantleech: In care it has an assignment we should automatically leave it alone, less WTF :)

mamazu: I mean you could make the case of both, because leaving it is just asking for trouble if the promoted argument is getting renamed.

lib/Extension/LanguageServerCodeTransform/CodeAction/PromotePropertyProvider.php
```diff
@@ -36,7 +36,7 @@ public function provideActionsFor(TextDocumentItem $textDocument, Range $range,
 
         return new Success([
             new CodeAction(
```
dantleech, line 38: Use named parameters for protocol classes as argument order can change when it's regenerated

mamazu:
Well I think I found a problem:
https://github.com/phpactor/phpactor/blob/fc7c4a3fd41a669f302143a40b9794080dd2aa1d/lib/CodeBuilder/Adapter/TolerantParser/Updater/AbstractMethodUpdater.php#L163

Remember this block of code? Where we decided to leave it and not touch parameters? Since phpactor would discard readonly status and other things when updating parameters? Well, now we need to update parameters. So there isn't a good way around it. Right?

lib/Extension/LanguageServerCodeTransform/CodeAction/PromotePropertyProvider.php
mamazu: Will do in the future, would you be opposed to abstracting this away and giving the `promoteProperty` Function on the handler an attribute that automatically registers it as a `CodeAction` provider?

lib/CodeTransform/Adapter/WorseReflection/Refactor/WorsePromoteProperty.php
dantleech: not sure I understand: I think for the sake of simplicity, this action should only act on parameters without visibility which have no corresponding property, as (for me at least) that is 99% of the use case, There are things like PHP-CS-Fixer and Rector which can migrate assigned properties to promoted ones.

lib/Extension/LanguageServerCodeTransform/CodeAction/PromotePropertyProvider.php
dantleech: no - at least not without some thought. I think the current way is explicit, adding an attribute would be "implicit" while not actually adding on LOC - because assuming we would not automatically scan all the classes in the project to find out which ones have attributes, you'd still need to register the class in the DI with a tag anyway...

dantleech:
> So there isn't a good way around it. Right?

:grimacing:

lib/Extension/LanguageServerCodeTransform/CodeAction/PromotePropertyProvider.php
mamazu: I think you kinda get the point but my idea on the implementation would be a little different. I'll open up an issue explaining it further.

mamazu:
> Of course another method would be to simply use the AST and insert `private` or `public` as a text edit and avoid the code builder entirely - and, thinking about it, this would really be simpler and less prone to disaster.

Well, that would indeed be much better. Though I like the interface of the source code builder but you are completely right. It would make the whole project easier to understand and easier to work with if we only needed to manipulate the AST.

lib/CodeTransform/Adapter/WorseReflection/Refactor/WorsePromoteProperty.php
mamazu: Yeah, as you figured out below. Technically if that's the only use case you could just add a shortcut to vim like `Iprivate ` or `Ipublic ` if you are already on the parameter. :D But having it in phpactor would be much nicer, I agree.

dantleech:
That's not manipulating the AST though, it's using the AST to determine where to insert a text edit. The AST is immutable.

mamazu:
True, I mean for the end user the result is the same, if we are re-printing the entire ast with one more token or if we instruct the editor to just update one line to add the token.

dantleech:
I will update the complete construtor transformer to handle promoting properties.

mamazu:
Sounds good, then I can close this PR.

PR #2093, opened 2023-02-13
[WIP] Making the current phpstan level better to prepare for phpstan level 9

mamazu:


lib/WorseReflection/Core/SourceCodeLocator/NativeReflectionFunctionSourceLocator.php
```diff
@@ -22,17 +23,22 @@ public function locate(Name $name): SourceCode
...
 
+        $fileName = $function->getFileName();
         if ($function->isInternal()) {
             throw new SourceNotFound(sprintf(
                 'Function "%s" is an internal function, there is another locator for that',
                 $name->__toString()
             ));
         }
 
-        return SourceCode::fromPathAndString($function->getFileName(), file_get_contents($function->getFileName()));
```
dantleech, line 36: can use `SourceCode::fromPath($function>getFilename())` instead

phpstan.neon
```diff
@@ -13,11 +13,8 @@ parameters:
         - lib/Test.php
         - lib/WorseReflection/Core/SourceCodeLocator/InternalStubs/*
         - lib/Extension/LanguageServerCodeTransform/Tests/Stub/*
-    reportUnmatchedIgnoredErrors: false
+    reportUnmatchedIgnoredErrors: true
```
dantleech, line 16: i set this to `false` because it happens on almost change which is rather unnecessary. It's easier to just regenerate the baseline every now and then. so would prefer to levae this as `false`

phpstan.neon
```diff
@@ -13,11 +13,8 @@ parameters:
...
     inferPrivatePropertyTypeFromConstructor: true
-    ignoreErrors:
-        - '#Yield can be used only with these return types: Generator, Iterator, Traversable, iterable.#'
```
dantleech, line 19: why removing this?

mamazu: Because we fixed all the errors.

phpstan.neon
mamazu: Yeah, I have only enabled that because I wanted to see how many errors I can remove.

phpstan.neon
dantleech: hm if that's the case then :+1: although I thought this was a specific issue with Amphp, but if it passes it passes :taco:

phpstan.neon
dantleech: I find it better to just `generate-baseline` and then do a diff to see how much was removed, as long as PHPStan is passing it's safe to do that.

mamazu: Yeah, I do the same. Just forgot to revert the change.

lib/WorseReflection/Core/SourceCodeLocator/NativeReflectionFunctionSourceLocator.php
mamazu: Yep, I have checked the whole code base and used it wherever possible.

lib/WorseReflection/Core/SourceCodeLocator/NativeReflectionFunctionSourceLocator.php
```diff
@@ -22,17 +23,22 @@ public function locate(Name $name): SourceCode
...
+        if (!$fileName || !file_exists($fileName)) {
+            throw new SourceNotFound(sprintf('Unable to locate file for function: "%s"', (string) $name));
+        }
```
dantleech, line 40: can remove this condition now or?

dantleech:
Happy to merge this when the CS fix is applied.

lib/WorseReflection/Core/SourceCodeLocator/NativeReflectionFunctionSourceLocator.php
mamazu: The second part yes. But the `$fileName` can still be null. #InconsitentReturnTypes

dantleech:
ready to review/merge?

mamazu:
Yes

dantleech:
thanks :tada:

PR #2094, opened 2023-02-15
Fixing phpstan to max in the class mover directory

mamazu:


mamazu:
Not the smartest move since we probably can't keep it up until it's done but it's still better than nothing. 

One thing that I noticed though is that the container could become a problem. `$this->get('some_service')` returns `mixed`.

lib/ClassMover/Domain/Model/ClassMemberQuery.php
```diff
@@ -11,7 +11,8 @@ final class ClassMemberQuery
     const TYPE_METHOD = 'method';
     const TYPE_PROPERTY = 'property';
 
-    private $validTypes = [
+    /** @var array<string> */
```
dantleech, line 14: not important but `list<string>` is better IMO

dantleech:
> Not the smartest move since we probably can't keep it up until it's done but it's still better than nothing.

Not sure what you mean

> One thing that I noticed though is that the container could become a problem. $this->get('some_service') returns mixed.

`$foo = $this->get(SomeService::class)` will return `SomeService`, but there is no way to know what `get('some_service')` will return some type hints would be required.

mamazu:
> Not sure what you mean.

It would have been smarter to raise the level and not make one directory level 9 compliant. Since now if someone would commit code to that directory it could remove the level 9 compliance. But raising the level creates problems everywhere and I like to deal with them one by one.

> `$foo = $this->get(SomeService::class)` will return `SomeService`, but there is no way to know what `get('some_service')` will return some type hints would be required.

That's good to know. So you are proposing that for all services to use the class or interface name for all of our services? That would make it easier for phpstan. And easier to navigate with phpactor as well.

dantleech:
> So you are proposing that for all services to use the class or interface name for all of our services?

Where there is a 1-1 mapping yes, but that's not always the case.

> But raising the level creates problems everywhere and I like to deal with them one by one.

It would probably take months to fix all the PHPStan issues, I'm also happy to raise it to level 9 and increase the baseline. It's also possible to run PHPStan on a specific directory without the baseline.

lib/ClassMover/Adapter/WorseTolerant/WorseTolerantMemberFinder.php
```diff
@@ -395,7 +412,9 @@ private function attachClassInfoToReference(MemberReference $reference, ClassMem
     private function memberStartPosition(Node $memberNode): int
     {
         assert($memberNode instanceof MethodDeclaration || $memberNode instanceof ConstElement || $memberNode instanceof Variable);
-        $start = $memberNode->name->start;
+        $name = $memberNode->name;
+        assert($name !== null);
```
dantleech, line 416: this is dangerous code, but I guess it's worked for 8 years or so.

dantleech (approved):

dantleech:
nice work and refactoring, not sure I looked closely enough, hope the tests are good :crossed_fingers:

PR #2095, opened 2023-02-15
Phpstan Level 9 with a larger baseline

mamazu:


dantleech:
what could go wrong, thanks!

PR #2096, opened 2023-02-16
Removing Phpstan warnings

mamazu:
Slowly and one by one. This should be ready to be merged at any point in time when the pipeline is green.


dantleech:
needs rebase

mamazu:
Well that looks like a bug from Github since the php level 9 commit is from the master. Very strange.

mamazu:
Continued in #2097

Issue #2060, opened 2023-01-29
Only suggest Attributes (i.e. classes with the Attribute attribute) for Attribute completion.


dantleech:
It "works" but only for searching all classes and only on the "short name". There are two things to make it more convenient:

- :heavy_check_mark:  Index "flags" associated with the class - while we probably don't want to index all the attributes a class has, it would be useful to know if it _is_ an attribute so we can efficiently filter it in the index. So we can add an "ATTRIBUTE" flag.
- :heavy_check_mark:  ~We can separately support completion on namespace segments.~ :point_left:  implemented.
- ... profit?

mamazu:
Is there anything I can do to help with this issue. At my company we are just getting started to work with attributes but trying to find the right class in a Symfony framework called `Route` is always guesswork.

mamazu:
Question is how do I need to change the query to only return "attribute classes" when completing an attribute

przepompownia:
I started to play with #2100. No test case was touched yet. Feel free to make changes if you think this might be a good start.

przepompownia:
Is there anything else missing to close this issue?

dantleech:
Nope I think that's it :)

PR #2100, opened 2023-02-18
Implement indexing classes with type attributes

mamazu:
As mentioned here: #2060 for an efficient suggestion when it comes to attributes we need to index which classes are attributes and which aren't. This PR adds a type `TYPE_ATTRIBUTE` which can be used for filtering later.

### DIFF
--- a/lib/Indexer/Adapter/Tolerant/Indexer/ClassDeclarationIndexer.php
+++ b/lib/Indexer/Adapter/Tolerant/Indexer/ClassDeclarationIndexer.php
@@ -52,7 +52,8 @@ class ClassDeclarationIndexer extends AbstractClassLikeIndexer
             foreach ($attributeGroup->attributes->children as $attribute) {
                 /** @var Attribute $attribute */
                 if ($attribute->name->getText() === 'Attribute') {
-                    $record->setType(ClassRecord::TYPE_ATTRIBUTE);
+                    $record->addFlag(ClassRecord::FLAG_ATTRIBUTE);
+                    return;
                 }
             }
         }
--- a/lib/Indexer/Adapter/Tolerant/Indexer/ClassDeclarationIndexer.php
+++ b/lib/Indexer/Adapter/Tolerant/Indexer/ClassDeclarationIndexer.php
@@ -4,12 +4,13 @@ namespace Phpactor\Indexer\Adapter\Tolerant\Indexer;
 
 use Microsoft\PhpParser\MissingToken;
 use Microsoft\PhpParser\Node;
+use Microsoft\PhpParser\Node\Attribute;
+use Microsoft\PhpParser\Node\Statement\ClassDeclaration;
 use Phpactor\Indexer\Model\Exception\CannotIndexNode;
+use Phpactor\Indexer\Model\Index;
 use Phpactor\Indexer\Model\Name\FullyQualifiedName;
 use Phpactor\Indexer\Model\Record\ClassRecord;
-use Microsoft\PhpParser\Node\Statement\ClassDeclaration;
 use Phpactor\TextDocument\TextDocument;
-use Phpactor\Indexer\Model\Index;
 
 class ClassDeclarationIndexer extends AbstractClassLikeIndexer
 {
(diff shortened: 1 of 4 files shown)


mamazu:
@dantleech I am trying to figure out what github wants from me and found out that you have a lot of stale branches and some even beginning with `master` like `master-list-system`. Would you mind removing them? Otherwise Just review the last commit and cherry-pick it on the master. I am out of ideas as to what I am doing wrong.

dantleech:
Hmm. I didn't take this approach because attributes _are classes_. They are _also_ attributes.

mamazu:
Yes, this would mean that you might need to change the query. But I don't think that this will slow down the query since if you put the "real" classes up front it will only be marginally slower for attributes.

lib/Indexer/Model/Record/ClassRecord.php
```diff
@@ -15,6 +15,7 @@ final class ClassRecord implements Record, HasFileReferences, HasPath, HasFullyQ
     public const TYPE_INTERFACE = 'interface';
     public const TYPE_TRAIT = 'trait';
     public const TYPE_ENUM = 'enum';
+    public const TYPE_ATTRIBUTE = 'attribute';
```
mamazu, line 18: Wouldn't this be better off as an enum?

dantleech:
So, "attribute" isn't a type of class, in the way that interface, trait and enum are.

So we definitely should not do this as it breaks categorisation.

Probably we should use a bitmask to add additional flags to indexed records (f.e. if it's abstract, an attribute, etc).

mamazu:
Sounds good to me.

Is there already an example on this? Or should I try to come up with my own implementation?

dantleech:
> Is there already an example on this? Or should I try to come up with my own implementation?

it's just an idea at this point, so not sure what's required. The basic thing is:

```
const ATTRIBUTE = 1;
// const ABSTRACT = 2;
// const DEPRECATED = 4;
// const WHATEVER = 8;
```

and you can then do bitwise operations: `$isAttribute = $record->flags() & ClassFlags::ATTRIBUTE` or something.

mamazu:
In this case I would add the Changelog entry when the autocompletion also picks that up.

lib/Indexer/Adapter/Tolerant/Indexer/ClassDeclarationIndexer.php
```diff
@@ -35,9 +37,28 @@ public function index(Index $index, TextDocument $document, Node $node): void
...
+        $attributes = $node->attributes ?? [];
+        if (count($attributes) === 0) {
+            return;
+        }
+
+        foreach ($attributes as $attributeGroup) {
+            foreach ($attributeGroup->attributes->children as $attribute) {
+                /** @var Attribute $attribute */
+                if ($attribute->name->getText() === 'Attribute') {
+                    $record->addFlag(HasFlags::ATTRIBUTE);
```
dantleech, line 56: can continue here, no need to go through all the attributes.

lib/Indexer/Model/Record/HasFlags.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace Phpactor\Indexer\Model\Record;
+
+interface HasFlags
```
dantleech, line 5: this can simply be a method on `ClassRecord` - the flags apply only to class record and could have a specific return type based on that information. If other things need "flags" they can be implemented separately.

lib/Indexer/Model/Record/HasFlagsTrait.php
```diff
@@ -0,0 +1,25 @@
...
+    {
+        $this->flags = $this->flags | $flag;
+        return $this;
+    }
+
+    public function flags(): int
+    {
+        return $this->flags;
+    }
+}
```
dantleech, line 25: as said, let's move this to class record for now. the valid flags are dependent on the tpye of record, and we may never add more.

lib/Indexer/Tests/Adapter/IndexBuilderTestCase.php
```diff
@@ -38,6 +39,24 @@ public function testIndexClass(string $source, string $name, Closure $assertions
...
+                    #[Attribute]
+                    class IamAnAttribute {}
+                DOC,
+            'IamAnAttribute',
+            function (ClassRecord $record): void {
+                self::assertInstanceOf(ClassRecord::class, $record);
+                self::assertEquals($this->workspace()->path('project/attribute.php'), $record->filePath());
+                self::assertEquals('IamAnAttribute', $record->fqn());
+                self::assertEquals(15, $record->start()->toInt());
+                self::assertEquals(HasFlags::ATTRIBUTE, $record->flags());
```
dantleech, line 56: `self::assertTrue($record->flags() & HasFlags::ATTRIBUTE)` would be more fitting for the usage.

lib/Indexer/Adapter/Tolerant/Indexer/ClassDeclarationIndexer.php
mamazu: True, one other thing I also wanted to bring up is, this only checks the name of the attribtue not the class. Which means that it returns true on any class called Attribute no matter the namespace.

dantleech: `$attribute->getResolvedName()` should return the FQN;

dantleech (approved):

dantleech:
thanks!

PR #2101, opened 2023-02-18
Making extractMethod from static function also generate static function

### LINKED ISSUE #1285: Generate method does not detect static contexts
That setup:
```php
	public static function setUpBeforeClass(): void
	{
		<>
		... some code ...
		<>
	}
```
should generate a static method, not an instance one.

mamazu:
Closes #1285


lib/CodeTransform/Tests/Adapter/WorseReflection/Refactor/WorseExtractMethodTest.php
```diff
@@ -39,154 +35,69 @@ public function testExtractMethod(string $test, string $name, ?string $expectedE
...
-            'extract method from trait' => [
-                'extractMethod33.test',
-                'newMethod'
-            ],
+        yield['extractMethod24.test'];
+        yield['extractMethod25.test'];
+        yield['extractMethod26.test'];
+        yield 'empty selection 2' => [
+            'extractMethod28.test',
         ];
```
dantleech, line 93: not sure i like this. the description has been lost, I'd rather either:

1) Keep the old test style
2) Refactor to `yield 'union argument => [ '...' ]`

I don't really mind if we keep or remove the method name.

dantleech (approved):

dantleech:
thanks! would be great in PRs that fix bugs or add features if you could also update the CHANGELOG.md

Issue #52, opened 2017-07-06
Blow up early if neither composer nor GIT are available


dantleech:

dantleech:
This is the current behavior.

PR #2105, opened 2023-02-18
Outsourcing Diagnostics to external process

dantleech:
This PR experiments with creating a new process to resolve diagnostics.

This has a big effect on peformance in large files (e.g. the toelrant-php-parser `Parser.php`) as diagnostics have been blocking all other LS functionality:

- `"PROF 7.2593 << request #52 [textDocument/completion]"` <- BEFORE
- `"PROF 0.8991 << request #30 [textDocument/completion]"` <- AFTER

The only real disadvantage is that unsaved files (other than the one being edited) will not be taken into account during the analysis - but this is a reasonable trade off I think.

(I originally envisioned worker "language server(s)" running in the background and syncing files to them, but meh, this delivers most of the value and doesn't suffer from stale caches or having to manage long-running processes)

TODO:

- [x] Configuration flag needs to enable / disable calling out to the worker process.
- [x] Code action providers not showing diagnostics, why?
- [ ] 
---

Additionally, I wonder if the command could be generalized to allow outsourcing things like `documentHighlight`, `codeActions` etc.

### DIFF
--- a/lib/Extension/LanguageServer/LanguageServerExtension.php
+++ b/lib/Extension/LanguageServer/LanguageServerExtension.php
@@ -482,7 +482,10 @@ class LanguageServerExtension implements Extension
         });
 
         $container->register(AggregateDiagnosticsProvider::class, function (Container $container) {
-            $providers = $this->collectDiagnosticProviders($container, false);
+            $providers = $this->collectDiagnosticProviders(
+                $container,
+                outsourced: $container->getParameter(self::PARAM_DIAGNOSTIC_OUTSOURCE) ? false : true,
+            );
 
             return new AggregateDiagnosticsProvider(
                 $this->logger($container, 'LSPDIAG'),
@@ -508,7 +511,8 @@ class LanguageServerExtension implements Extension
         ]);
 
         $container->register(OutsourcedDiagnosticsProvider::class, function (Container $container) {
-            if (false === $container->getParameter(self::PARAM_DIAGNOSTIC_OUTSOURCE)) {
+            // only register this if we should call out to an external process for diagnostics
+            if (!$container->getParameter(self::PARAM_DIAGNOSTIC_OUTSOURCE)) {
                 return null;
             }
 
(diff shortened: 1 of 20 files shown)


lib/Extension/LanguageServerCodeTransform/LanguageServerCodeTransformExtension.php
```diff
@@ -320,7 +321,9 @@ private function registerCodeActions(ContainerBuilder $container): void
                 'Fix PSR namespace and class name'
             );
         }, [
-            LanguageServerExtension::TAG_DIAGNOSTICS_PROVIDER => [],
+            LanguageServerExtension::TAG_DIAGNOSTICS_PROVIDER => [
+                'channel' => 'phpactor',
+            ],
```
dantleech, line 326: use DiagnosticTagProvider here and should be "outsourced" too

lib/Extension/LanguageServer/LanguageServerExtension.php
```diff
@@ -506,9 +509,24 @@ private function registerDiagnostics(ContainerBuilder $container): void
...
+            /** @var PathResolver $resolver */
+            $resolver = $container->get(FilePathResolverExtension::SERVICE_FILE_PATH_RESOLVER);
+            $projectPath = $resolver->resolve('%project_root%');
+            // only register this if we should call out to an external process for diagnostics
+            if (!$container->getParameter(self::PARAM_DIAGNOSTIC_OUTSOURCE)) {
+                return null;
+            }
+
+            return new OutsourcedDiagnosticsProvider([
+                __DIR__ . '/../../../bin/phpactor',
```
dantleech, line 525: this should be fine, could also use the file path resolver to get the application root

lib/WorseReflection/Core/Type/IntType.php
```diff
@@ -29,15 +29,15 @@ public function shiftLeft(Type $right): Type
...
-
-        return new BooleanType();
+public function bitwiseXor(Type $right): Type
+{
+    if ($right instanceof IntType && $right instanceof Literal && $this instanceof Literal) {
+        return $this->withValue($this->value() ^ $right->value());
     }
 
+    return new BooleanType();
+}
```
dantleech, line 39: ?

PR #2110, opened 2023-02-22
Adding more types

mamazu:


lib/ClassMover/Domain/Name/QualifiedName.php
```diff
@@ -7,7 +7,7 @@
 class QualifiedName
 {
     /**
-     * @param string[] $parts
+     * @param non-empty-array<string> $parts
```
mamazu, line 10: Is this correct or can the list of parts also be empty. Then functions like end and begin sometimes return false which we need to handle.

dantleech: i think this is fine

dantleech: but would be good to handle the error scenarios, ideally the class should be closed and it should not be possible for those error conditions to happen, but in practice it probably is.

lib/ClassMover/Tests/Adapter/TolerantParser/TolerantClassReplacerTest.php
```diff
@@ -38,7 +38,8 @@ public function testFind($fileName, $classFqn, $replaceWithFqn, $expectedSource)
         self::assertStringContainsString($stripEmptyLines($expectedSource), $stripEmptyLines($edits->apply($source->__toString())));
     }
 
-    public function provideTestFind()
+    /** @return array<array<string>> */
```
dantleech, line 41: not for future, prefer `list<string>` where it makes sense.

lib/ClassMover/Tests/Adapter/WorseTolerant/WorseTolerantMemberFinderTest.php
```diff
@@ -19,7 +19,8 @@ public function testFindMember(string $source, ClassMemberQuery $classMember, in
         $this->assertCount($expectedRiskyCount, $members->withoutClasses());
     }
 
-    public function provideFindMember()
+    /** @return array<array{0: string, 1: ClassMemberQuery, 2: int, 3?:int}> */
+    public function provideFindMember(): array
```
dantleech, line 23: note you can use the Phpactor code action "generate docblock return types" here to get 80% of the way, I would prefer multi line PHPDocs but won't stop a PR onit.

lib/ClassMover/Tests/Adapter/TolerantParser/TolerantClassReplacerTest.php
mamazu: I have never understood what the difference between those is. I mean the most generic would be `iterable<string>` what would be the difference there?

lib/CodeBuilder/Adapter/TolerantParser/Updater/AbstractMethodUpdater.php
```diff
@@ -141,19 +146,21 @@ private function updateOrAddParameters(Edits $edits, Parameters $parameters, Met
 
         $renderedParameters = [];
 
+        /** @var ParameterDeclarationList|null $existingParameterDeclaration */
+        $existingParameterDeclaration = $methodDeclaration->parameters;
```
dantleech, line 150: upstream TP has wrong declaration here.

lib/ClassMover/Tests/Adapter/WorseTolerant/WorseTolerantMemberFinderTest.php
mamazu: Fair enough I'll keep that in mind. I like single line ones since they are less visually distracting. But shouldn't be a problem to switch. Or we add a code formatter rule for it, since the project uses both.

lib/CodeBuilder/Adapter/TolerantParser/Updater/ClassMethodUpdater.php
```diff
@@ -2,16 +2,20 @@
...
+use Microsoft\PhpParser\Node\TraitMembers;
 use Phpactor\CodeBuilder\Domain\Renderer;
 use Phpactor\CodeBuilder\Domain\Prototype\Method;
 use RuntimeException;
+use Microsoft\PhpParser\Node;
 
 class ClassMethodUpdater extends AbstractMethodUpdater
 {
-    public function memberDeclarationsNode(ClassLike $classNode)
+    public function memberDeclarationsNode(ClassLike $classNode): ClassMembersNode|TraitMembers|InterfaceMembers
```
dantleech, line 18: what about EnumMembers?

lib/CodeBuilder/Adapter/TolerantParser/Updater/AbstractMethodUpdater.php
mamazu: Exactly, can we change that somehow?

lib/CodeBuilder/Adapter/TolerantParser/Updater/ClassMethodUpdater.php
dantleech: but this cannot return anything other then ClassMembersNode? same for the other "updaters" I guess, they should only return the corresponding `*MembersNode`

dantleech (changes requested):
left a few comments, some of which might need to be actioned.

lib/CodeBuilder/Adapter/TolerantParser/Updater/AbstractMethodUpdater.php
dantleech: It would be fixing a needle in a haystack of wrong types. I've started pushing PHPStan fixes up stram: https://github.com/microsoft/tolerant-php-parser/pull/396

but upstream is sometimes slow, and I wanted to add some changes which I didn't have the capacity to justify, so we use a fork: https://github.com/phpactor/tolerant-php-parser

lib/CodeBuilder/Adapter/TolerantParser/Updater/ClassMethodUpdater.php
mamazu: Correct EnumMembers are missing. I will add them but we don't have an enum updater that's why I forgot / left it out.

dantleech: we should use a template type here.

```
/**
 * @template T
 * @extends AbstractMethodUpdate<ClassMembersNode>
 */
class ClassMethodUpdater ...

/**
 * @template TMembersNodeType of Node
 */
class AbstractMethodUpdater ...
{
   /** @return TMembersNodeType */
    abstract public function memberDeclrationsNode(...)
}
```

lib/ClassMover/Tests/Adapter/TolerantParser/TolerantClassReplacerTest.php
dantleech: list is an array, but when used like a vector: `[1,2,3,4]`, i.e. the keys are implicit and correspond to the index of the value.

lib/ClassMover/Domain/Name/QualifiedName.php
mamazu: Yeah, as far as I know there is only the factory methods that make this object.

dantleech:
does it make sense to use a generic type for AbstractMethodUpdater?

lib/ClassMover/Tests/Adapter/TolerantParser/TolerantClassReplacerTest.php
mamazu: But in this case we have explicit string keys.

mamazu:
Yes, it would make sense to use a template type there. I'll add that to the code. But I'd still leave the php return type there since phpdoc comment types easily become outdated method signatures are a little harder to mess up especially with tests.

lib/CodeBuilder/Adapter/TolerantParser/Updater/AbstractMethodUpdater.php
mamazu: I'll make a note in the other package. Should I also fix it here?

dantleech:
> I'd still leave the php return type

let's remove it (assuming we're talking about `ClassMembers|EnumMembers|TraitMembers|etc`. it effectively couples the abstract class to all it's implementations and would be incompatible with a generic type. tests cover the functionality.

lib/CodeBuilder/Adapter/TolerantParser/Updater/AbstractMethodUpdater.php
dantleech: i would rather we updated all PHPStan level than try and fix each individual error. so fine to leave this.

dantleech:
> In my opinion the problem is that there is no Memberinterface that is implemented by ClassMembers, EnumMEmbers and Traitmembers

doesn't really solve the problem though, you'd still need generics.

mamazu:
Yeah, I have removed the typehints but something with the doc comments is off.

dantleech:
is this PR ready for review?

mamazu:
Yep should be good now.

.phpactor.json
```diff
@@ -0,0 +1,7 @@
+{
+    "$schema": "/home/mamazu/packages/phpactor/phpactor/phpactor.schema.json",
+    "language_server_phpstan.enabled": true,
+    "language_server_psalm.enabled": false,
+    "language_server_php_cs_fixer.enabled": true,
+    "prophecy.enabled": true
+}
```
dantleech, line 7: remove this :)

dantleech (approved):

dantleech:
thanks!

PR #2118, opened 2023-02-26
Moving constructor to the top when creating methods

### LINKED ISSUE #2113: Generator constructor: generate at top of file, not bottom...
```
class TestChangeSuggestor implements ChangeSuggestor
{
    public function suggestChanges(): Changes
    {
    }

    /**
     * @param Closure(): Changes $closure
     */
    public function __construct(Closure $closure)
    {
    }
}

```

mamazu:
Closes #2113

### DIFF
--- a/lib/CodeBuilder/Adapter/TolerantParser/Updater/AbstractMethodUpdater.php
+++ b/lib/CodeBuilder/Adapter/TolerantParser/Updater/AbstractMethodUpdater.php
@@ -36,24 +36,26 @@ abstract class AbstractMethodUpdater
         }
 
         $lastMember = $this->memberDeclarationsNode($classNode)->openBrace;
+        $lastNonMethodMember = null;
+        $methodHasBeenEncountered = false;
         $newLine = false;
         $existingMethodNames = [];
         $existingMethods = [];
-        $firstMethod = null;
         foreach ($this->memberDeclarations($classNode) as $memberNode) {
             if ($memberNode instanceof PropertyDeclaration) {
                 $lastMember = $memberNode;
                 $newLine = true;
+                if (!$methodHasBeenEncountered) {
+                    $lastNonMethodMember = $memberNode;
+                }
             }
 
             if ($memberNode instanceof MethodDeclaration) {
-                if ($firstMethod === null) {
-                    $firstMethod = $memberNode;
-                }
                 $lastMember = $memberNode;
                 $existingMethodNames[] = $memberNode->getName();
                 $existingMethods[$memberNode->getName()] = $memberNode;
                 $newLine = true;
+                $methodHasBeenEncountered = true;
             }
         }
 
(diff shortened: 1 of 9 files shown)


lib/CodeBuilder/Adapter/TolerantParser/Updater/AbstractMethodUpdater.php
```diff
@@ -91,6 +95,19 @@ public function updateMethods(Edits $edits, ClassLikePrototype $classPrototype,
         }
 
         foreach ($methodPrototypes as $methodPrototype) {
+            // If class has methods add the constructor before it.
+            if ($methodPrototype->name() === '__construct' && $firstMethod !== null) {
+                // We can't use the first method's starting location because this is already indented to we unindent the code
+                $edits->add(
+                    TextEdit::create(
+                        $firstMethod->getStartPosition() - 4,
```
mamazu, line 103: Maybe there is a better text token I could use?

dantleech: the indentation is configurable, so we can't depend on -4. the current code always starts _after_ and adds a new line. If we could determine the last non-method node then we can do it the same way.

mamazu:
I found a way around it. But now it has an extra new line at the end which if we want to remove it, could be a little tricky.

dantleech:
:wave: are the test failures expected?

mamazu:
@dantleech If you have time, could you give feedback on this please?

dantleech:
in the following case the last member was `const FOO`:

```
class Test
{
    private $fo;


    public function __construct(string $string)
    {
    }
    const FOO = 'BAR';
}
```

in this case I'd expect it to be above `bar()`, but it's fine:
```
class Test
{
    const FOO = 'BAR';
    public function bar() {}
    private $fo;


    public function __construct(string $string)
    {
    }

}
```

dantleech:
nice - but still a failing test:

```

1) Phpactor\CodeBuilder\Tests\Adapter\TolerantParser\TolerantUpdaterTest::testMethods with data set "It generates a constructor with promoted properties" ('class Foo\n{\n}', Phpactor\CodeBuilder\Domain\Prototype\SourceCode Object (...), 'class Foo\n{\n    public func...  }\n}')
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
     public function __construct(private int $config)\n
     {\n
     }\n
+\n
 }'
```

dantleech:
if possible don't force push after a review as it makes seeing what changed harder

mamazu:
I will try to remember that. And as for rebasing? Also rather merging the changes?

dantleech:
either way, it's all squashed in the end, not sure if merging would make it easier to review.

dantleech:
constructor is placed before constants:
```
/**
 * @method static Bar foobar(string $baz)
 */
class Test
{
    public function __construct(string $string)
    {
    }

    const  BAR ='';
}
```

dantleech:
but merging now, feel free to fix in a follow up if you want :tada:

PR #2120, opened 2023-02-27
Deprecation diagnostic

### LINKED ISSUE #2104: Provide deprecation information
Currently phpactor doesn't report when a method is deprecated. It would be nice if there would be a way to do that probably best with a diagnostic?

dantleech:
Would fix #2104

Code:

image

Neovim:

image

Should this be a WARNING, INFO or HINT?

TODO:

- [x] Support properties and constants
- [x] Support classes and class-likes


mamazu:
I would say hint, because it still works but shouldn't be used when rewritten.

dantleech:
hint is not shown on NVIM or VS Code :/ INFO could be OK, but actually it _is_ a warning.

lib/Extension/LanguageServerPhpCsFixer/Provider/PhpCsFixerDiagnosticsProvider.php
```diff
@@ -243,8 +243,6 @@ private function createRangesForDiff(Diff $fileDiff): array
 
                     continue;
                 }
-
-                throw new LogicException('This is not intended to be reached');
```
dantleech, line 247: cc @lumnn this _was_ reached of course and crashes the diagnostic engine :smile:

dantleech:
test failing for the VIM tests because of a weird state issue involving the index in CI. maybe need to clean the index / disable indexing -- or simply rename the test class in VimPlugin to something unique and not `Foobar`.

dantleech:
works now, but diagnostics performance takes a big hit for some reason.

lib/Extension/LanguageServerPhpCsFixer/Provider/PhpCsFixerDiagnosticsProvider.php
lumnn: Bit late to answer, any chances to share a piece of code on which this would be reached?

lumnn: No worries, will have a look into that then

PR #2127, opened 2023-03-03
Gh 2122: Fix prophecy not returning correct type for method prophecy methods

dantleech:
Currently just a proof of concept, should be generalized maybe.

### DIFF
--- a/lib/Extension/Prophecy/WorseReflection/ProphecyMemberContextResolver.php
+++ b/lib/Extension/Prophecy/WorseReflection/ProphecyMemberContextResolver.php
@@ -29,6 +29,9 @@ class ProphecyMemberContextResolver implements MemberContextResolver
     ): ?Type {
         $memberType = $member->inferredType();
         if ($memberType instanceof GenericClassType && $memberType->instanceof(TypeFactory::reflectedClass($reflector, 'Prophecy\Prophecy\ObjectProphecy'))->isTrue()) {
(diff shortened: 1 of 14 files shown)


lib/WorseReflection/Tests/Inference/generics/method_returns_templated_generic.test
```diff
@@ -0,0 +1,37 @@
...
+    /**
+     * @return Collection<T>
+     */
+    public function foobar(): Collection
+    {
+    }
+}
+
+$foo = new Test(new \stdClass());
+wrAssertType('stdClass', $foo->foobar()->get());
```
dantleech, line 37: this test worked as expected first time, keeping though as the test that covers this isn't so obvious.

lib/Extension/Prophecy/WorseReflection/ProphecyMemberContextResolver.php
```diff
@@ -26,6 +29,9 @@ public function resolveMemberContext(
     ): ?Type {
         $memberType = $member->inferredType();
         if ($memberType instanceof GenericClassType && $memberType->instanceof(TypeFactory::reflectedClass($reflector, 'Prophecy\Prophecy\ObjectProphecy'))->isTrue()) {
+            if ($member->name() === 'getObjectProphecy') {
```
dantleech, line 32: this should probably be genalised, but i'm too tired right now.

PR #2136, opened 2023-03-11
Adding types to `WorseExtractMethod` reflection

mamazu:
This should have a smaller scope to review. A random file, but what I noticed is that the functions `parameterVariables` and `returnVariables` return a hashmap variable name => variable where the key is used nowhere. In the code if you want the variable name you just say `$variable->name()`. Is there a reason you did it as a hash map and not as a list?


dantleech (approved):

dantleech:
normally it's to de-duplicate, but if there's not test for it :shrug:

PR #2138, opened 2023-03-12
Experimental: Inlay Hint Support

dantleech:
Add experimental support for inlay hints:

- [x] Show variable type
- [x] Show method parameter names
- [ ] Show function call and class instantiation parameter names, created follow up #2139 
- [x] Make it non-blocking AND/OR outsource it to an external process (it's v. slow)

In addition:

- Bumps the `phpactor/container` version and introduces `$container->parameter('foo')->toInt()` to provide type safety for container parameters cc @mamazu 
- Refactors the frame resolver to use a generator (i.e. to be non-blocking) this made a huge difference in performance and could further be applied to the remaining in-process WR diagnostics to speed things up.
- Incorporates and fixes https://github.com/phpactor/phpactor/pull/2135 because I'm too lazy to rebase.

image

*In Neovim using [lspinlayhints](https://github.com/lvimuser/lsp-inlayhints.nvim) plugin*

Note that I cannot get this to do anything in VS code


lib/Extension/LanguageServerWorseReflection/InlayHint/InlayHintWalker.php
```diff
@@ -0,0 +1,113 @@
...
+            return $frame;
+        }
+        if ($this->options->types && $node instanceof Variable) {
+            $this->fromVariable($resolver, $frame, $node);
+            return $frame;
+        }
+        if ($this->options->params && $node instanceof CallExpression) {
+            $this->fromCall($resolver, $frame, $node);
+            return $frame;
+        }
```
dantleech, line 55: can add more pretty easily as with #2139

iamreinder:
@dantleech Can you tell me how you configured `lsp-inlayhints.nvim`? I'm trying out this feature, but can't seem to get it working. I've add `"language_server_worse_reflection.inlay_hints.enable": true` to my phpactor.json, when I check the config from within Neovim it shows true as well. The inlayhints plugin is loaded. So I think I'm missing something. Did you add any config to the inlayhints plugin setup function? Thanks in advance!

PR #2143, opened 2023-03-16
Using a version of phpcsfixer that supports php8.2

mamazu:
Since we are using php8.2 in the CI we should also use a version of phpcsfixer that supports that this way we don't need to enable it the hacky way with env vars.


dantleech:
:vulcan_salute:

PR #2148, opened 2023-03-19
Adding types to all parameters that could be typed

mamazu:


dantleech (approved):

mamazu:
Okay, I have no idea why this pipeline failed. I think I caught all errors with floats in the config.

dantleech:
Nice work! The failing VIM test could be annoying.. Away from keyboard atm but if there is nothing obvious I'll check later

dantleech:
I think this was due to calling `phpactor set:config ..cache_lifetime 5.0` before the VIM test, which somehow ended up writing an `int` to the config file, I update Container to cast int to float: https://github.com/phpactor/phpactor/pull/2149

mamazu:
Yeah I screwed that up but I couldn't find the place where it was buggy. But thanks for fixing it.

PR #2157, opened 2023-03-21
Towards an intermediate representation. of a sort.

dantleech:
This PR aims to result in static analysis being done in a single pass (in phpactor) and being represented by a single document.

The outcome should be:

- Faster analysis
- Better ways to intercept and modify the the context

Notes:

- The PR aims to change as little as possible: there are lots of dragons but I don't want to fix the world.
- `NodeContext` and subclasses of `NodeContext` will have children (all the better to invalidate with)
- The `FrameWalker` concept will be destroyed leaving only resolvers to walk themselves.

Nice to haves (but maybe better left for later):

- Use of Promises (i.e. non-blocking analysis).
- Cache reimplemented to be per URI (i.e. invalidate only when the document is updated)
- "NodeContext" to provide an invalidation _scope_ (e.g. local, class, global) to prevent analysing the entire document each time

## Benchmarks

Before (with cache):

```
\Phpactor\WorseReflection\Tests\Benchmarks\AnalyserBench

    benchAnalyse............................I0 - Mo1.785s (±0.00%)
```

After (**without**( cache):

```
\Phpactor\WorseReflection\Tests\Benchmarks\AnalyserBench

    benchAnalyse............................I0 - Mo1.556s (±0.00%)
```

Adding the same cache (per-node resolution cache) as previously results in saving a further 200ms **BUT** it breaks some tests, but this is not important as the caching strategy will change.

The new caching strategy will be based on invalidating local, class or global scopes, so we do not need to re-process every single node for each small change.

### Resolvers

As of previousoly, resolvers are responsible for converting an AST node to a corresponding `NodeContext` (which can either be the default one or a more specialised one). The difference is that they will form a tree structure. representing an enhanced AST.

- They have a new signature: `public function resolve(NodeContextResolver $resolver, NodeContext $context, Node $node): NodeContext
`
- The `NodeContext` is created automatically, a minimum implementation can simply return the `$context`
- The `$context` has a `->parent` which is set to the context resolved by the parent of the corresponding node.

### Visitors

`Visitors` are able to "visit" and modify the context after it has been resolved. In this PR it replaces the original "walker" concept


dantleech:
Giving up, it's getting out of control with no strategy or end in sight, some learnings:

- An IR would solve issues:
  - Traversing the nodes in the correct order (e.g. assignments will evaluate the right hand side before the left hand side)
  - Being able to invalidate a scope (e.g. no need to rebuild the entire document)
  - Remove the "frame walker" concept, simplifying things.

Some ideas that could be carried forward, possibly on the existing code:

- Can invalidate the cache based on URL when text edits are made
- We can permit the use of promises and make WR non-blocking.

PR #2158, opened 2023-03-22
Some minor clean ups

mamazu:
* Adding more types in some places
* Removing the unused NameSearcherCompletion.php which is the same as NameSearcherCompletor.php
* Using a generator to generate Suggestions
* Fixing typo in a function in the NameUtil
* Regenerating the documentation


dantleech (approved):

dantleech:
thanks!

PR #2167, opened 2023-03-26
Adding guide for adding code actions

mamazu:
This pull request adds a check list for what is needed when wanting to write a new code action. (because I also forgot that myself after writing the generate decorator).

I am not good at writing documentation, if you need any stylistic changes let me know.


doc/other/developing.rst
```diff
@@ -16,6 +16,25 @@ Phpactor which are in `lib/Extension`.
...
+Let's have a look at the concept by the example of the generate decorator functionality. First off we have the
+`GenerateDecoratorProvider` this class provides a list of code actions for the file. Here we can do some pre checks e.g.
+if the file is usable to generate a decorator in. In the end the `GenerateDecoratorProvider` will return a list of
+commands that are available.
+
+The command is defined in our example in the `GenerateDecoratorCommand` which is just some glue between the actual
+code action logic and the provider.
+
+And then there is the implementation of the `WorseGenerateDecorator` which contains the logic for generating the
+decorator or more generally to apply the code action.
```
dantleech, line 36: I think this needs to be a new document rather than a section here, so `doc/development/language-server/code-action.rst`.

But contentfully this looks OK. I realise that the process for creating a code-action is quite intensive currently and could be simplified.

mamazu: I have broken up the documentation and added some more to it, about what a code action is and what kinds there are.

mamazu:
Looks like I broke the docs, but I have no idea of how RST works, could you give me a hint?

doc/development/language-server/debugging.rst
```diff
@@ -1,62 +1,5 @@
...
-    [OK] Server listening on tcp://127.0.0.1:9912
-
-    // Quit the server with CONTROL-C
-
-You can use the `dump` function in the code and the variable will be shown in
-the console output of the server you started: `dump($var)`.
-
 Debugging RPC
--------------
+=============
```
dantleech, line 2: this is h1? this should stay as h2 `----` if there is more than one section in the document, only the first header (h1) should use `===`

dantleech:
if you haven't already you can check the `.github` code action for how to install Sphinx then do `make docs` to build locally

mamazu:
I ran it locally and it said that it was fine. Very strange.

mamazu:
Should be fixed now.

doc/development/language-server/debugging.rst
```diff
@@ -76,7 +19,7 @@ Phpactor from the shell and replay your last command:
 
 .. code:: bash
 
-   $ phpactor rpc --replay 
+   $ phpactor rpc --replay
```
dantleech, line 22: nothing in this document is specific to the language server I think, it can be moved to the parent level

dantleech:
opened a PR against this PR: https://github.com/mamazu/phpactor/pull/2/files :)

mamazu:
I have merged it.

PR #2170, opened 2023-03-29
Do not overwrite constants when extracting

### LINKED ISSUE #2166: Refactor: Extract constant overwrites already existing constant
Hey,
i just noticed that the refactor option `extract constant` overwrites already existing constants with the same name. Here is an example:

```php
class Example
{
	public function __construct()
	{
		$this->foobar("This is an example", 'This is also a test');
	}

	public function foobar(string $param1, string $param2): void
	{
	}
}
```

When i extract the first string i get the following:
```php
class Example
{
    const NEW_CONSTANT = 'This is an example';

	public function __construct()
	{
		$this->foobar(self::NEW_CONSTANT, 'This is also a test');
	}

	public function foobar(string $param1, string $param2): void
	{
	}
}
```

But when i want to extract the second string i get the following:
```php
class Example
{
    const NEW_CONSTANT = 'This is an example';

	public function __construct()
	{
		$this->foobar(self::NEW_CONSTANT, self::NEW_CONSTANT);
	}

	public function foobar(string $param1, string $param2): void
	{
	}
}
```

This is wrong, because the first and the second string are not the same. A quick workaround for this is to rename the first constant before refactoring the second one. But maybe phpactor should choose a new name (`NEW_CONSTANT_2` for example) to avoid data loss or confusion while refactoring.

Thanks so much and greetings
Leo

mamazu:
Closes #2166

### DIFF
--- a/doc/reference/configuration.rst
+++ b/doc/reference/configuration.rst
@@ -1237,7 +1237,7 @@ Specify which diagnostic providers should be active (default to all)
 If applicable diagnostics should be "outsourced" to a different process
 
 
-**Default**: ``false``
+**Default**: ``true``
 
 
 .. _param_language_server,file_events:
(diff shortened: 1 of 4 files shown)


dantleech (approved):

dantleech:
nice work!

PR #2171, opened 2023-04-01
Fix documentation typo

allendoerfer:


dantleech:
Hi, this document is auto generated, so not sure what's happened here..

mamazu:
The error is in the file `lib/Extension/LanguageServer/LanguageServerExtension.php` at line 100. However, if we change the code this is technically going to be a bc breaks which should also be mentioned in the `UPGRADE.md` file.

dantleech:
It's correct documentation, there is a typo in the code:

https://github.com/phpactor/phpactor/blob/f9c74b7121919c9d088baea712fa8956dfad5122/lib/Extension/LanguageServer/LanguageServerExtension.php#L100

allendoerfer:
I changed the mentioned file and added a line to the `CHANGELOG.md`, because I could not find an `UPGRADE.md` file.

dantleech:
Thanks!

PR #2177, opened 2023-04-07
Fix indexing bug for attributes

dantleech:


mamazu:
Do you know how this could happen?

dantleech:
:shrug:

PR #2179, opened 2023-04-07
Benchmark which exposes performance issue

dantleech:
This PR adds a PHPStan source file which takes an extreme amount of time to analyse for some reason.. the goal of this PR is to fix the performance issue.

-----
Removing "sorting" from the frame improves performance 100x in this example. I'm pretty sure the sorting was there to solve another issue though, although the tests seem to pass now :thinking:

### DIFF
--- a/lib/WorseReflection/Core/Inference/Assignments.php
+++ b/lib/WorseReflection/Core/Inference/Assignments.php
@@ -27,7 +27,6 @@ abstract class Assignments implements Countable, IteratorAggregate
         foreach ($variables as $variable) {
             $this->variables[$variable->key()] = $variable;
         }
-        $this->sort();
     }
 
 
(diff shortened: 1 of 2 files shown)


dantleech:
merging, will see if it causes any regressions, if it does it's missing a test and sorting 1ks of times with 5ks of local variables is not a good solution.

mamazu:
Nice catch. This will change the result of the functions `last()` `lastOrNull()` and `first()` in this class. Previously they returned the variable with the highest or lowest offset. Now they return the first and most recent variable insertions into the `$this->variables` array. Which I guess if you parse and process the tree top to bottom should be the same.

dantleech:
Yes... I'm assuming we do now, but waiting for the bug reports.

PR #2180, opened 2023-04-07
Adding generator for boilerplate methods in PhpUnit

mamazu:


lib/Extension/PHPUnit/CodeTransform/Domain/Refactor/GenerateTestMethods.php
```diff
@@ -0,0 +1,80 @@
...
+            return;
+        }
+
+        $class = $classes->classes()->first();
+        if (!$class instanceof ReflectionClass) {
+            return;
+        }
+
+        if (!$class->isInstanceOf(ClassName::fromString(TestCase::class))) {
+            dump('Not a unit test');
```
mamazu, line 44: @dantleech I don't know why classes extending the `TestCase` class still fall into this case here. Maybe you have an idea.

lib/Extension/PHPUnit/CodeTransform/Domain/Refactor/GenerateTestMethods.php
```diff
@@ -0,0 +1,80 @@
...
+        if (count($classes->classes()) !== 1) {
+            return;
+        }
+
+        $class = $classes->classes()->first();
+        if (!$class instanceof ReflectionClass) {
+            return;
+        }
+
+        if (!$class->isInstanceOf(ClassName::fromString(TestCase::class))) {
```
dantleech, line 43: note `TestCase` will not be loadable without the `--dev` dependencies, which _may_ be fine

lib/Extension/PHPUnit/CodeTransform/Domain/Refactor/GenerateTestMethods.php
```diff
@@ -0,0 +1,80 @@
...
+use Phpactor\TextDocument\TextDocumentUri;
+use Phpactor\TextDocument\WorkspaceEdits;
+use Phpactor\WorseReflection\Core\ClassName;
+use Phpactor\WorseReflection\Core\Exception\NotFound;
+use Phpactor\WorseReflection\Core\Reflection\ReflectionClass;
+use Phpactor\CodeTransform\Domain\SourceCode;
+use Phpactor\WorseReflection\Reflector;
+
+
+class GenerateTestMethods /* implements GenerateMethod */
```
dantleech, line 22: this isn't "domain" a it knows about worse reflection. and this type of abstraction isn't really worth it for extensions. Putting itin `lib/Extension/PHPUnit/CodeTransform/GenerateMethods.php` would be fine.

lib/Extension/PHPUnit/CodeTransform/Domain/Refactor/GenerateTestMethods.php
```diff
@@ -0,0 +1,80 @@
...
+
+    public function generateMethod(TextDocument $document, ByteOffset $offset): WorkspaceEdits
+    {
+        $class = $this->reflector->reflectClassesIn($document)->classes()->first();
+
+        $builder = SourceCodeBuilder::create();
+        $builder->namespace($class->name()->namespace());
+        $classBuilder = $builder->class($class->name()->short());
+
+        foreach(['setUp', 'tearDown'] as $methodName) {
```
dantleech, line 59: We should offer to generate only one. It is extremely rare that I need a `tearDown` method f.e.

lib/Extension/PHPUnit/CodeTransform/Domain/Refactor/GenerateTestMethods.php
dantleech: probably rather that the PHPUnit test case does not exist in your test, so the instanceof fails. You could add a `namespace PHPUnit\Framework; class TestCase {}` to the workspace

lib/Extension/PHPUnit/CodeTransform/Domain/Refactor/GenerateTestMethods.php
mamazu: Done.

lib/Extension/PHPUnit/CodeTransform/Domain/Refactor/GenerateTestMethods.php
mamazu: Yeah, the end goal is to generate a code action for every method and then let the user choose.

lib/Extension/PHPUnit/Tests/Unit/CodeTransform/Domain/Refactor/GenerateTestMethodsTest.php
```diff
@@ -0,0 +1,62 @@
...
+        $sourceCode = SourceCode::fromStringAndPath('<?php '.$source, 'file:///source');
+
+        // Adding a phpunit stub
+        $this->workspace()->put('TestCase.php', <<<PHP
+            <?php
+
+            namespace PHPUnit\Framework;
+
+            class TestCase {}
+        PHP);
```
mamazu, line 27: Is that the correct way to add the file to the workspace? Because it still doesn't find it.

mamazu: Never mind I forgot the use statement in the other code. :D

lib/Extension/PHPUnit/Tests/Unit/CodeTransform/GenerateTestMethodsTest.php
```diff
@@ -0,0 +1,120 @@
...
+     *
+     * @dataProvider dataCanMethodBeGenerated
+     */
+    public function testCanMethodBeGenerated(string $source, array $expected): void
+    {
+        $sourceCode = SourceCode::fromStringAndPath('<?php '.$source, 'file:///source');
+
+        $methodNames = $this->createTestMethodGenerator($source)->getGeneratableTestMethods($sourceCode);
+
+        $this->assertEquals($expected, iterator_to_array($methodNames));
```
dantleech, line 36: `self::asset` to be consistent (also it _is_ a static method)

dantleech:
this isn't working for me, investigating

mamazu:
I have implemented your suggestions. But I don't know how to test it. I think I have two language servers installed currently: one from the phpactor vim plugin and one from the repository I am developing in. I'll have to look into how I make my vim use the development server for testing.

mamazu:
related to 2045

dantleech:
I have this:

```
lspconfig.phpactor.setup{
    capabilities = capabilities,
    cmd = { '/home/daniel/www/phpactor/phpactor/bin/phpactor', 'language-server' },
    filetypes = { 'php', 'cucumber' },
    on_attach = on_attach,
    init_options = {
    }
}
```

and often use `:LspRestart` to reload the LS after making chages.

mamazu:
Yeah that works. The thing is it correctly executes the command and it shows the `Workspace Edit: Generated "setup" method.` But the edit doesn't get updated in the document. 

Do you have any idea why?

dantleech:
haven't got a development env at the moment - are the text edits getting sent with the correct URI?

mamazu:
It looks like it. They are send with the URI of the document on which the code action was invoked.

mamazu:
Okay I have rebased this onto the master and now it works. Sorry for the difficult review again. But merging the master into this branch somehow caused some problems.

lib/Extension/PHPUnit/CodeTransform/GenerateTestMethods.php
```diff
@@ -0,0 +1,85 @@
...
+            if ($class->methods()->has($methodName)) {
+                return TextEdits::none();
+            }
+        } catch (NotFound) {
+            return  TextEdits::none();
+        }
+
+        $setUpMethod = $classBuilder->method($methodName);
+        $setUpMethod->visibility('public');
+        $setUpMethod->returnType('void');
```
dantleech, line 81: _could_ use the fluent interface here.

lib/Extension/PHPUnit/Tests/Unit/CodeTransform/GenerateTestMethodsTest.php
```diff
@@ -0,0 +1,120 @@
...
+{
+    public function setUp(): void
+    {
+        // Adding a phpunit stub
+        $this->workspace()->put('TestCase.php', <<<PHP
+                <?php
+
+                namespace PHPUnit\Framework;
+
+                class TestCase {}
```
dantleech, line 21: let's add the setUp and tearDown methods to this class

lib/Extension/PHPUnit/CodeTransform/GenerateTestMethods.php
```diff
@@ -0,0 +1,85 @@
...
+        $builder->namespace($class->name()->namespace());
+        $classBuilder = $builder->class($class->name()->short());
+
+        try {
+            if ($class->methods()->has($methodName)) {
+                return TextEdits::none();
+            }
+        } catch (NotFound) {
+            return  TextEdits::none();
+        }
```
dantleech, line 77: this try/catch block is not required I think

lib/Extension/PHPUnit/CodeTransform/GenerateTestMethods.php
```diff
@@ -0,0 +1,85 @@
...
+        );
+
+        $class = $this->reflector->reflectClassesIn($document)->classes()->first();
+
+        $builder = SourceCodeBuilder::create();
+        $builder->namespace($class->name()->namespace());
+        $classBuilder = $builder->class($class->name()->short());
+
+        try {
+            if ($class->methods()->has($methodName)) {
```
dantleech, line 72: this is why the method is not being added, it needs to be:

```
        if ($class->ownMembers()->methods()->has($methodName)) {
            return TextEdits::none();
        }
```

perhaps we could make the tests fail before applying the fix to make them green :+1:

mamazu: Yeah removing your fix, will break the test. (I checked.) But I have no idea how I should commit that.

dantleech: It's only a development approach (TDD) to help ensure the test is a good one. (I will often commit when the test is failing and then again when it's green)

dantleech:
danke!

PR #2191, opened 2023-04-09
Invalidate diagnositc cache on update

dantleech:
Adds a caching mechanism for diagnostics based on the text document URI - has _dramatic_ impact on large files where it can take > than the default cache TTL for code actions to be resolved (e.g. on large file it took > 1 minute to resolve code actions, now it takes 7 seconds on first run) 

A further optimisation would be to cache reflection classes - although this might look different as we don't know the text document URI, so would be more like tag invalidation.


lib/Extension/WorseReflection/WorseReflectionExtension.php
```diff
@@ -137,6 +140,11 @@ private function registerReflection(ContainerBuilder $container): void
         $container->register(Cache::class, function (Container $container) {
             return new TtlCache($container->parameter(self::PARAM_CACHE_LIFETIME)->float());
         });
+        $container->register(CacheForDocument::class, function (Container $container) {
+            return new CacheForDocument(
+                fn () => new StaticCache(),
```
dantleech, line 145: this _should_ be fine -- the cache is removed when documents are closed.

PR #2201, opened 2023-04-12
Fix uncaught exception

vlada-dudr:
Solves bug caused by not checking status of subprocess after sleeping. If child finished between sleeping and attempting to kill it, phpactor crashed hard.


dantleech (approved):

dantleech:
good find, thanks :tada:

PR #2208, opened 2023-04-15
Keeping class references fully qualified after moving

### LINKED ISSUE #2115: Error when using `class:move` with FQCN
There is an error in the refactoring when using `class:move` with code that uses FQCN (Fully Quallified Class Name).

```
$ phpactor class:move src/Foo/Bar/Baz/Qux.php src/Foo/Qux.php
[REPL] /home/pierrick/test/src/Foo/Bar.php
       51:67 \Foo\Bar\Baz\Qux => Foo\Qux
[REPL] /home/pierrick/test/src/Foo/Bar/Baz/Qux.php
       63:66 Qux => Qux
[MOVE] /home/pierrick/test/src/Foo/Bar/Baz/Qux.php => /home/pierrick/test/src/Foo/Qux.php
```

The result of this refactoring for the Bar.php file is this

```diff
diff --git a/src/Foo/Bar.php b/src/Foo/Bar.php
index 9105219..bfd0da7 100644
--- a/src/Foo/Bar.php
+++ b/src/Foo/Bar.php
@@ -1,8 +1,10 @@
 <?php
 
+use Foo\Qux;
+
 class Bar
 {
-    public function __construct(\Foo\Bar\Baz\Qux $qux)
+    public function __construct(Foo\Qux $qux)
     {
     }
 }
```

The expected result should be 

```diff
-    public function __construct(\Foo\Bar\Baz\Qux $qux)
+    public function __construct(\Foo\Qux $qux)
```

mamazu:
If a class name was fully qualified before moving, then it should be fully qualified after moving as well.

Fixes #2115

### DIFF
--- a/lib/ClassMover/Adapter/TolerantParser/TolerantClassReplacer.php
+++ b/lib/ClassMover/Adapter/TolerantParser/TolerantClassReplacer.php
@@ -34,6 +34,18 @@ class TolerantClassReplacer implements ClassReplacer
 
         foreach ($classRefList as $classRef) {
             assert($classRef instanceof ClassReference);
+
+            // Check if the class name is fully qualified and replace it with a fully qualified version of the new name
+            $firstClassNameCharacter = $source->__toString()[$classRef->position()->start()];
+            if ($firstClassNameCharacter === '\\') {
+                $edits[] = TextEdit::create(
+                    $classRef->position()->start(),
+                    $classRef->position()->length(),
+                    '\\'.$newName->__toString()
+                );
+                continue;
+            }
+
             if (false === $importClass) {
                 $importClass = $this->shouldImportClass($classRef, $originalName);
             }
@@ -59,6 +71,7 @@ class TolerantClassReplacer implements ClassReplacer
         });
 
         $edits = TextEdits::fromTextEdits($edits);
+        $source = Code::fromString($source);
         if (true === $importClass) {
             $edits = $edits->merge($this->addUseStatement($source, $newName));
         }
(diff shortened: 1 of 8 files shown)


lib/ClassMover/Adapter/TolerantParser/TolerantClassReplacer.php
```diff
@@ -34,6 +34,18 @@ public function replaceReferences(
 
         foreach ($classRefList as $classRef) {
             assert($classRef instanceof ClassReference);
+
+            // Check if the class name is fully qualified and replace it with a fully qualified version of the new name
+            $firstClassNameCharacter = $source->__toString()[$classRef->position()->start()];
```
mamazu, line 39: I think that this is probably not the right way to check if a class name is fully qualified or not. Any suggestions?

dantleech: maybe add `NodeUtil::isFullyQualified($name)` as with:

https://github.com/phpactor/phpactor/blob/1b10fbb0042d5f25ff59f39b94981c8ba76e3739/lib/Name/NameUtil.php#L34

dantleech: `$classRef instanceof FullyQualifiedName` ?

dantleech: looks like you could add a method to `QualifiedName`: `wasFullyQualified`. ATM it seems it then has the contents `["", "Foo", "Bar"]` otherwise you could check if it was fully qualified in the static constructor.

mamazu: This should work but the ClassMover namespace has it's own version of a FullyQualified name. Should I try to also add it to that? Because this changes a lot of tests if I do.

dantleech: Yes , class mover has it's own stuff...

mamazu: I've added it in both versions.

dantleech: AFAIK ClassMover doesn't use the `Name` package. Can we revert those changes? A name doesn't normally know if it _was_ fully qualfiied ,,,,

mamazu: Sure thing.

lib/ClassMover/Adapter/TolerantParser/TolerantClassReplacer.php
```diff
@@ -103,13 +112,16 @@ private function classIsTheOriginalInstance(ClassReference $classRef, FullyQuali
...
+    private function addUseStatement(Code $source, FullyQualifiedName $newName): TextEdits
     {
-        return $this->updater->textEditsFor(SourceCodeBuilder::create()->use($newName->__toString())->build(), Code::fromString($source->__toString()));
+        return $this->updater->textEditsFor(SourceCodeBuilder::create()->use($newName->__toString())->build(), $source);
     }
 
-    private function addNamespace(TextDocument $source, QualifiedName $qualifiedName): TextEdits
+    private function addNamespace(Code $source, QualifiedName $qualifiedName): TextEdits
     {
-        return $this->updater->textEditsFor(SourceCodeBuilder::create()->namespace($qualifiedName->__toString())->build(), Code::fromString($source->__toString()));
```
dantleech, line 113: note `Code::` need probably be entirely replcaed by TextDocument, so this type of refatoring isnt really necessaary

dantleech (approved):

dantleech:
merci!

PR #2215, opened 2023-04-17
Self-documenting/testing providers

### LINKED ISSUE #2212: Self-testing/documenting units
Currently many features of Phpactor have an inconsistent approach to testing, and amost all of them require manual documentation which is often omitted completely.

It would be good to investigate self-testing/documenting for features such as diagnostics and code actions and provide a consistent way to configure them.

Such an approach could look like this:

```
/**
 * Prose documentation for the unit goes in the docblock here.
 * 
 * ```markdown
 * Markdown can be used
 * ```
 */
class SomeDiagnosticProvider
public function examples(): Examples {
    return Examples::from(
         new DiagnosticExample(
             title: "Undefined variable",
             source: '<?php echo $notExisting;',
             valid: false,
             hidden: false,
             assertion: function(Diagnostics $diagnostics) {
                 Assert::assertEquals("some message", $diagnostics->at(0)->message());
             }
         )
    );
}
```

this will allow us to:

- ensure that we the documentation is comprehensive and accurate
- reduce the barrier to entry and delivery time for features
- encourage the addition of tests
- reduce the LOC necessary to add these features

dantleech:
Resolves #2212 

Inspired by PHP-CS-Fixer this PR experiments with adding self-contained diagnostics that:

- Document themselves
- Test themselves

This means that to implement a diagnostic provider it would be necessary only to:

- Implement the diagnostics provider
- Wire up the provider to the DI container

After this it should:

- Automatically be tested according to the examples it provides
- Documented according to the example and the prose in the docblock


dantleech:
Have implmented this for diagnostic providers, will do a follow up PR for adding documentation and then consider an approach for code actions and replacing the Refactoring.rst

PR #2221, opened 2023-04-20
Small bugfix for null exception on empty var

haringsrob:
In the logs the the following would constantly throw errors:

```php
class demo {
  public string $<CUR>
}
```

This is because the variable has no name yet.

### DIFF
--- a/lib/WorseReflection/Bridge/TolerantParser/Diagnostics/UndefinedVariableProvider.php
+++ b/lib/WorseReflection/Bridge/TolerantParser/Diagnostics/UndefinedVariableProvider.php
@@ -25,13 +25,19 @@ class UndefinedVariableProvider implements DiagnosticProvider
         if ($node->parent?->parent instanceof PropertyDeclaration) {
             return [];
         }
-        foreach ($frame->locals()->byName($node->getName()) as $variable) {
-            if ($variable->wasDefinition()) {
-                return [];
-            }
-        }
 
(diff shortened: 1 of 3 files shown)


lib/WorseReflection/Bridge/TolerantParser/Diagnostics/UndefinedVariableProvider.php
```diff
@@ -25,13 +25,19 @@ public function enter(NodeContextResolver $resolver, Frame $frame, Node $node):
...
+
+        if ($name) {
+            foreach ($frame->locals()->byName($name) as $variable) {
+                if ($variable->wasDefinition()) {
+                    return [];
+                }
             }
+        } else {
+            $name = 'UNDEFINED';
         }
```
dantleech, line 39: in general avoid `else`  - but can we not simply `if (!$name) { return; }` ? why set it to `UNDEFINED`?

haringsrob: Because I did not look at the code properly and did not realise it :D

dantleech (approved):

dantleech:
sorry, conflicted while you were working on it

haringsrob:
Should be good now?

dantleech (approved):

dantleech:
yep, thanks

PR #2226, opened 2023-04-21
Gh 2223: Fix list assignment 

### LINKED ISSUE #2223: Array Destructuring doesn't register as a variable definition
## Steps to reproduce

Create a test case containing:
```php
<?php

[$a, $b] = [1, 2];

echo $a;
echo $b;
```

This should not throw any diagnostics. Here are two problems:

1. The diagnostic provider should throw an error on the array destructuring expression (easy, early return in the diagnostic)
2. The variables should be defined after the array destructuring and be part of the frame (I don't know where this logic should be)

If you can point me in the right direction for number 2. I can implement a fix for that.

dantleech:
Fixes  #2223


mamazu:
Or you do it. :D Looks good.

dantleech:
Wanted to fix the diagnostic but it fixed both so 🤷

PR #2227, opened 2023-04-21
Adding Support for `willImplement` and `willExtend`

mamazu:
Adding support for those methods in prophecies. However, this only works with the last instance eg.
```
$object = $this->prophecise(A::class);

$object->willExtend(B::class); // type: A|B
$object->willImplement(C::class); // type A|C

As far as I know there is no way you can specify a return type of a class with a new generic because will extens should return ObjectProphecy<T&U> with T being the original type and U being the type of the parameter.

### CODE REVIEW

**Comment by mamazu** [comment]
Refs #2123

File: lib/Extension/Prophecy/stubs/Prophecy.stub [hunk 1]
```diff
@@ -24,6 +24,42 @@ class ObjectProphecy
...
     {
     }
+
+    /**
+     * Forces double to extend specific class.
+     *
+     * @template U
+     * @param class-string<U> $class
+     *
+     * @return self|T|U
```
dantleech, line 34: should be `ObjectProphecy<T>` ?

lib/Extension/Prophecy/stubs/Prophecy.stub
```diff
@@ -24,6 +24,42 @@ class ObjectProphecy
...
+    }
+
+    /**
+     * Forces double to implement specific interface.
+     *
+     * @template U
+     *
+     * @param class-string<U> $interface
+     *
+     * @return self|T|U
```
dantleech, line 47: should be `ObjectProphecy<T&U>` and might require more than stubs

dantleech:
you'd also need to add tests here: https://github.com/phpactor/phpactor/blob/d59f228113ad30ff0239e35ed8cdcc7e70eb353c/lib/Extension/Prophecy/Tests/Integration/WorseReflection/ProphecyMemberContextResolverTest.php#L12

lib/Extension/Prophecy/stubs/Prophecy.stub
mamazu: Well it should be `ObjectProphecy<T&U>` but as far as I know, we don't support that.

dantleech: Arguably that might be what the issue is about :)

dantleech: And it is "T" as it extends, it is not an intersection

mamazu: Yeah I thought so, then I might need to dig into the code a little more.

mamazu:
@dantleech Could you have a look why the test is failing on an unrelated thing?

dantleech:
restarted, flakey test

mamazu:
Then it's ready to merge.

dantleech:
ok - but this isn't correct is it? it's _not_ a union? or?

mamazu:
Well it should be. If I write this code:
```
$mock = $this->prophizize(A::class);
$mock->willExtend(B::class);
```

Then this means the mock Object has this signiture:

```php
class A extends B {}
```

Which then should mean that we should be able to call all functions from `A` and all functions from `B` (technically only the public ones, but this also goes for A). This is the union of A ∪ B. At least that's what I am reading.

dantleech:
It's not A _or_ B. It's just A

mamazu:
Then I don't know what the correct behaviour would be.

dantleech:
if you have `interface Foo` and (for some reason) want to prophesize a class that extends `Bar` _and_ implements `Foo` then you would `->prophesize(Foo::class)->willExtend(Bar::class)`, It's not a union it's `XXX extends Bar implements Foo`, you would the correct type is `XXX`, not a union.

This missing feature is creating that type dynamically, which isn't that trivial.

mamazu:
Any idea what we should do with this PR?

PR #2242, opened 2023-04-25
Removing deprecated DefinitionLocation class

mamazu:
And replacing it with the more generally named `Location` class.


lib/Extension/LanguageServerReferenceFinder/Tests/Unit/Handler/GotoDefinitionHandlerTest.php
```diff
@@ -31,10 +29,11 @@ class GotoDefinitionHandlerTest extends TestCase
     public function testGoesToDefinition(): void
     {
         $document = TextDocumentBuilder::create(self::EXAMPLE_TEXT)->uri(self::EXAMPLE_URI)->build();
+
         $locations = [
             new TypeLocation(
                 TypeFactory::class('Foo'),
-                new DefinitionLocation($document->uri(), ByteOffset::fromInt(2))
+                new PhpactorLocation($document->uriOrThrow(), ByteOffset::fromInt(2))
```
dantleech, line 36: can we remove the alias and just have `Location` ?

lib/ReferenceFinder/Tests/Unit/LocationTest.php
```diff
@@ -3,18 +3,18 @@
 namespace Phpactor\ReferenceFinder\Tests\Unit;
 
 use PHPUnit\Framework\TestCase;
-use Phpactor\ReferenceFinder\DefinitionLocation;
 use Phpactor\TextDocument\ByteOffset;
+use Phpactor\TextDocument\Location;
 use Phpactor\TextDocument\TextDocumentUri;
 
-class DefinitionLocationTest extends TestCase
+class LocationTest extends TestCase
```
dantleech, line 10: this test no longer maps to a class, we can either simply remove it or move the test cases to the test for `TextDocument\Location`

dantleech (approved):

dantleech:
:+1:  looks good!

lib/Extension/LanguageServerReferenceFinder/Tests/Unit/Handler/GotoDefinitionHandlerTest.php
mamazu: In this file there is this `Phpactor\TextDocument\Location` and `Phpactor\LanguageServerProtocol\Location` the later one is not aliased so the first one needs an alias.

lib/ReferenceFinder/Tests/Unit/LocationTest.php
mamazu: Removing since there already is a test for the location in the TextDocument namespace.

dantleech:
Danke!

Issue #2233, opened 2023-04-22
If > 1 definition exists offer choice of definitions to jump to


mamazu:
I don't know if that is a local problem but in every project I opened phpactor never found the "real" Phpunit and always referred to stubs.

## Steps to reproduce
* Checkout the phpactor repository
* Open a test (it should already show that the assertion methods like `assertCount` `assertSame` etc could not be found)
* Go to definition of the Phpunit `Testcase` class the test is extending 

Then you should be redirected to `lib/CodeTransform/Tests/Workspace/TestCase.php` which is just a stub in this project and not the real test case.

This isn't a phpactor problem, but I think this is a problem in general. If your project has polyfills like older version of phpunit had the Testcase in a different namespace so some packages alias the new version to a stub.

## Other considerations
What I could do on a per project basis is exclude the files that I know are stubs in the index pattern. But I don't know if that is a good idea.

dantleech:
I think this should only happen in the Phpactor project since we introduced the possiblity that the "TestCase" artifact is left in the file tree?

More generally we should probably allow navigating to any of the matching classes.

dantleech:
Will probably need to support indexing a list of paths for any given class FQN in the index, which probably isn't a big deal.

mamazu:
As far as I can see the definition finder already returns a list of definitions it's just that we only ever take the first one. But improvement is already on the way.

dantleech:
yes, but the indexer only supports a single path for any given FQN

mamazu:
Refs #2243

dantleech:
As mentioned in the other PR, PHPStorm does not support this feature and it makes no sense to analyse > 1 file for the same class definition, in this specific case the problem can be resolved by ignoring the Workspace folder in `.phpactor.json`:

```
    "indexer.exclude_patterns": [
        "/vendor/**/Tests/**/*",
        "/vendor/**/tests/**/*",
        "/**/Workspace/**/*",
        "/var/cache/**/*",
        "/vendor/composer/**/*"
    ],
```

PR #2243, opened 2023-04-26
Select which declaration to implementation to jump to

mamazu:
Related to #2233

### DIFF
--- a/lib/Extension/ReferenceFinderRpc/Handler/GotoDefinitionHandler.php
+++ b/lib/Extension/ReferenceFinderRpc/Handler/GotoDefinitionHandler.php
@@ -2,11 +2,14 @@
 
 namespace Phpactor\Extension\ReferenceFinderRpc\Handler;
 
+use Phpactor\Extension\ReferenceFinderRpc\LocationSelector;
+use Phpactor\Extension\Rpc\Response;
 use Phpactor\MapResolver\Resolver;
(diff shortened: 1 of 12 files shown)


lib/Extension/ReferenceFinderRpc/Handler/GotoDefinitionHandler.php
```diff
@@ -55,18 +60,20 @@ public function configure(Resolver $resolver): void
...
-            $location->uri()->path(),
-            $location->offset()->toInt()
-        )->withTarget($arguments[self::PARAM_TARGET]);
+        $locations = [];
+        foreach($this->locator->locateDefinition($document, $offset) as $typeLocation) {
+            $locations[] = $typeLocation->location();
+        }
+        $locationsObject = new Locations($locations);
+
+        return $this->locationSelector->selectFileOpenResponse($locationsObject, $arguments[self::PARAM_TARGET]);
```
dantleech, line 77: we don't need to implement this for RPC ...

dantleech:
I'm a bit confused... :) we already support multiple goto-definition targets in the language server. the problem is that we only index one path per FQN .... we don't need to support this in the legacy RPC stuff.

mamazu:
Ah I got confused by the lagacy code. Well then I will have to look into the implementation of the multiple definitions for one FQN in the indexer.

Just out of curiosity, is there a reason we still support the lagacy stuff?

PR #2245, opened 2023-04-28
Failing tests for multiple references for one FQN

mamazu:


dantleech:
after implementing this locally I'm not sure we want this behavior, PHPStorm does not support this f.e. and analysing > 1 file for the same class also makes no sense really. the solution is to ignore the `Workspace` class when working in Phpactor to avoid the duplciate PHPUnit TestCase definition.

dantleech:
closing for now...

haringsrob:
I think this was also briefly mentioned in one of the Laravel story. IDE-helpers may document classes that have magic methods. 
Because those classes due to the "stubs" exist in multiple locations phpactor only picks up one of them.

dantleech:
Yes, need to understand better how that works f.e. in phpstorm

haringsrob:
I will check that, I still have my licence

PR #2250, opened 2023-05-01
Fixing the ObjectProphecy type

mamazu:
## What's in there

* ObjectProphecy|A => ObjectProphecy<A> which is the correct type
* Now php-cs-fixer also added type hints to the properties

## Why?

For some reason my version of the phpstan analyser returns 220 errors when running it with the same configuration as in the pipeline. Even a fresh install doesn't help. :shrug:  Well some of the issues are valid and have been fixed in this PR.

### DIFF
--- a/lib/Completion/Tests/Unit/CompletorTest.php
+++ b/lib/Completion/Tests/Unit/CompletorTest.php
@@ -17,9 +17,9 @@ class CompletorTest extends TestCase
     const EXAMPLE_OFFSET = 1234;
 
     /**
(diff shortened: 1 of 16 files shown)


dantleech (approved):

dantleech:
thanks, you could try `phpstan --debug` to run without the cache, and if that works, find the cache and delete it.

mamazu:
Nope, same with --debug. Could it be that it is related to me running it with 8.2.4?

PR #2253, opened 2023-05-03
Removing some of the deprecations

mamazu:
* Removing the `nameImport` function
* Removing deprecated `LineColFromOffset` class
* Removing parameter `method()`


dantleech (approved):

dantleech:
:bow:

PR #2257, opened 2023-05-06
Removing unused and untested phpunit type assertions

mamazu:


mamazu:
Okay this class is neither tested nor registered in the container anywhere. Would it make more sense to delete this? I think the problem in here would be that we don't want to resolve the return type of the function but rather narrow or change the type of the argument.

dantleech:
> Would it make more sense to delete this?

this extension was imported from the original [phpunit extension](https://github.com/phpactor/phpunit-extension) which used a different API IIRC, there would've been a test for this but it seems to have been lost, feel free to delete it.

dantleech:
:tada:

Issue #2195, opened 2023-04-09
References to `__construct`: exclude superclasses


przepompownia:
Thanks for #2194, this change increases the convenience of searching. It has a minor bug: see

```diff
diff --git a/lib/Indexer/Tests/Adapter/ReferenceFinder/IndexedReferenceFinderTest.php b/lib/Indexer/Tests/Adapter/ReferenceFinder/IndexedReferenceFinderTest.php
index 7cde458d..e0940f74 100644
--- a/lib/Indexer/Tests/Adapter/ReferenceFinder/IndexedReferenceFinderTest.php
+++ b/lib/Indexer/Tests/Adapter/ReferenceFinder/IndexedReferenceFinderTest.php
@@ -263,6 +263,20 @@ class IndexedReferenceFinderTest extends IntegrationTestCase
         ,
             1
         ];
+
+        yield 'show new object expressions when finding references on __construct except for superclass' => [
+            <<<'EOT'
+                // File: project/Bar.php
+                <?php class Bar { public function __construct() {} }
+                // File: project/subject.php
+                <?php class SubBar extends Bar { public function __c<>onstruct() {} }
+
+                // File: project/class1.php
+                <?php new Bar(); new SubBar();
+                EOT
+        ,
+            1
+        ];
     }
```
```
 $ ./vendor/bin/phpunit lib/Indexer/Tests/Adapter/ReferenceFinder/IndexedReferenceFinderTest.php --filter 'with data set "show new object expressions when finding references on __construct except for superclass"'
PHPUnit 9.6.6 by Sebastian Bergmann and contributors.

F                                                                   1 / 1 (100%)

Time: 00:00.136, Memory: 10.00 MB

There was 1 failure:

1) Phpactor\Indexer\Tests\Adapter\ReferenceFinder\IndexedReferenceFinderTest::testFinder with data set "show new object expressions when finding references on __construct except for superclass" ('// File: project/Bar.php\n<?p...Bar();', 1)
Total confirmed
Failed asserting that actual size 2 matches expected size 1.
```

przepompownia:
> we would want the child classes

Yes, only `Subbar` in the example - but not `Bar` as its superclass.

mamazu:
@przepompownia It's merged, could you see if that resolves your issue.

przepompownia:
Yes, nice to have it: for example in Phpactor let's compare references to the constructor in

[...]

PR #2261, opened 2023-05-07
Improve finding references to only take subclasses into account

mamazu:
Ref #2195 

The idea of this PR is. Get the implementations for the class your are calling the function on and from there on only subclasses eg.

```php
class Foo extends Bar {}
class Other extends Bar {}
```

## Expected behaviour

It should make the following queries:
* Get all implementations of `Foo`
* Get all subclasses and interfaces of `Bar`

## Actual behaviour

Currently it does the following queries
* Get all implementations of `Foo`
* Get all implementations of `Bar`

Which means that it also returns Other.

## Current problem
The `IndexReferenceFinder` already resolves the parent class before doing calling the `implementationOf` function. Which means it still does the wrong thing. But I also don't want to change the implementation of the `$containerType` variable here:
https://github.com/phpactor/phpactor/blob/9d3a87f148773949ab8ad5dbbbc1342fb50595bc/lib/Indexer/Adapter/ReferenceFinder/IndexedReferenceFinder.php#L91


lib/Indexer/Tests/Adapter/ReferenceFinder/IndexedReferenceFinderTest.php
```diff
@@ -159,6 +159,23 @@ public function provideFunctions(): Generator
...
     public function provideMembers(): Generator
     {
+        yield 'show new object expressions when finding references on __construct except for superclass' => [
+            <<<'EOT'
+                // File: project/Bar.php
+                <?php class Bar { public function __construct() {} }
+                // File: project/subject.php
+                <?php class SubBar extends Bar { public function __c<>onstruct() {} }
+                // File: project/subject.php
+                <?php class Other extends Bar { public function __c<>onstruct() {} }
```
przepompownia, line 169: Is the second cursor marker intentional?

przepompownia:
@mamazu thanks for initial draft.

At the first glance it looks like if `resolveDeclaringContainerType` were search through parents of `SubBar` ending with `Bar`, for whose implementation we then look for references.

https://github.com/phpactor/phpactor/blob/50f791e835d05655c9dc6449e3e05fc96a67cca3/lib/Indexer/Adapter/ReferenceFinder/IndexedReferenceFinder.php#L91

Notice `->original()` here:
https://github.com/phpactor/phpactor/blob/50f791e835d05655c9dc6449e3e05fc96a67cca3/lib/Indexer/Adapter/ReferenceFinder/Util/ContainerTypeResolver.php#L28

@mamazu @dantleech any ideas on that?

mamazu:
To be honest I haven't looked into the resolution process much myself. But that could be the issue here yeah.

lib/Indexer/Tests/Adapter/ReferenceFinder/IndexedReferenceFinderTest.php
```diff
@@ -159,6 +159,23 @@ public function provideFunctions(): Generator
...
+                // File: project/subject.php
+                <?php class SubBar extends Bar { public function __c<>onstruct() {} }
+                // File: project/subject.php
+                <?php class Other extends Bar { public function __construct() {} }
+
+                // File: project/class1.php
+                <?php new Bar(); new SubBar(); new Other();
+                EOT
+                ,
+                2
```
przepompownia, line 175: Only one `new Subbar()` is called.

mamazu: Yeah you are right. I think 2 is the number it currently returns.

mamazu:
@przepompownia Well, I dug into it and I think that adding the `->original()` call is not correct in any case. I have replaced this method and now the tests fail in multiple places. All of them should be updated, because

```php
class A { public function foo(); }
class B extends A { public function foo(); }

$b = new B();
$b->fo<>o();
```

Then it should not find the reference to `A::foo()` since it's been overwritten. This should only happen when the function is not present on the current class.

przepompownia:
@mamazu I still can't find good time to looking through your proposition.

mamazu:
The only thing I don't get is now the tests are failing on something unrelated that I haven't changed (or at least not touched knowingly)

lib/Indexer/Tests/Adapter/ReferenceFinder/IndexedReferenceFinderTest.php
```diff
@@ -206,17 +250,35 @@ public function provideMembers(): Generator
...
+                // File: project/class1.php
+                <?php class ParentalClass { public function method() {}}
+
+                // File: project/class2.php
+                <?php class Foobar extends ParentalClass { }
                 EOT
         ,
-            2, 4 // total number is multipled due to implementation recursion
+            2, 4 // total number is multiple due to implementation recursion
         ];
```
przepompownia, line 283:
```php
        yield 'deep members no override' => [
            <<<'EOT'
                // File: project/subject.php
                <?php $foo = new Foobar(); $foo->method();
                // File: project/subject1.php
                <?php $bar = new ParentalClass(); $bar->met<>hod();
                // File: project/class1.php
                <?php class ParentalClass { public function method() {}}
                // File: project/class2.php
                <?php class Foobar extends ParentalClass { }
                EOT
        ,
            ?, ? // 
        ];
```
I have no opinion whether we should include overridden classes or not but probably we should test such case too.

lib/Indexer/Adapter/ReferenceFinder/IndexedReferenceFinder.php
```diff
@@ -129,8 +129,21 @@ private function implementationsOf(string $fqn): Generator
...
+
+        // From now on we can only get it's subclasses
+        foreach($implemenations as $implementation) {
+            yield $implementation;
+
+            foreach($this->query->class()->subClasses($implementation) as $subClass) {
+                if ((string) $subClass !== $fqn) {
+                    yield $subClass;
+                }
+            }
```
przepompownia, line 146: Do we need to keep it in two separate loops (and use `$implementations`)?

lib/Indexer/Tests/Adapter/ReferenceFinder/IndexedReferenceFinderTest.php
mamazu: There is already a test for that. (The one you posted). Or what do you mean exactly?

przepompownia: We have four test cases here. In all of them the cursor position is on method defined in subclass. That's why I thought to check references of some parent method too unless it is out of the scope of this change.

lib/Indexer/Tests/Adapter/ReferenceFinder/IndexedReferenceFinderTest.php
```diff
@@ -206,17 +250,35 @@ public function provideMembers(): Generator
                 <?php namespace Bar; $foo = new Foobar(); $foo->b<>ar();
 
                 // File: project/subject1.php
-                <?php namespace Bar; $foo = new Barfoo(); $foo->bar();
+                <?php namespace Bar; $bar = new Barfoo(); $bar->bar();
 
                 // File: project/class1.php
-                <?php namespace Bar; class Foobar extends Barfoo { public function bar() {}}
+                <?php namespace Bar; class Barfoo { public function bar() {}}
```
przepompownia, line 256: `public static` according to the description?

mamazu: What description?

lib/Indexer/Tests/Adapter/ReferenceFinder/IndexedReferenceFinderTest.php
mamazu: Which four test cases are you thinking of? Like having the cursor on the method definition?

przepompownia: https://github.com/phpactor/phpactor/pull/2261/files#diff-fe706911329feb79e7b2e1f0a11109e1518c601584a9319a377ab8458f3857abR173
https://github.com/phpactor/phpactor/pull/2261/files#diff-fe706911329feb79e7b2e1f0a11109e1518c601584a9319a377ab8458f3857abR192
https://github.com/phpactor/phpactor/pull/2261/files#diff-fe706911329feb79e7b2e1f0a11109e1518c601584a9319a377ab8458f3857abR250
https://github.com/phpactor/phpactor/pull/2261/files#diff-fe706911329feb79e7b2e1f0a11109e1518c601584a9319a377ab8458f3857abR269

mamazu: You mean that we don't test that if a method is overriden it still is able to find references? That's a good point. I mean it shouldn't be a problem. But I can add a test case for that as well.

lib/Indexer/Tests/Adapter/ReferenceFinder/IndexedReferenceFinderTest.php
mamazu: Yeah the test should actually have a static property. I'm not sure why it doesn't have it. I'll add it. :)

lib/Indexer/Tests/Adapter/ReferenceFinder/IndexedReferenceFinderTest.php
przepompownia: Yes, but at the moment I am still not sure what to expect from such test.

przepompownia:
Independently of comments it looks good for me.

lib/Indexer/Tests/Adapter/ReferenceFinder/IndexedReferenceFinderTest.php
mamazu: Oh there is a break in the git diff just noticed, the test case isn't called static. :laughing:

lib/Indexer/Tests/Adapter/ReferenceFinder/IndexedReferenceFinderTest.php
mamazu: In other IDEs :tm: you can select if you want to see the usage of the base method or the overriden methods. But this is definitely outside the scope of the PR.

mamazu:
@dantleech If you could have a look at it and review it, that would be great. Otherwise it's ready for merging.

dantleech:
what could go wrong

przepompownia:
I went through positive path but in practice I found a side effect not present before: for example in `lib/Indexer/Extension/IndexerExtension.php` look at references to `Watcher`:
image

przepompownia:
https://github.com/phpactor/phpactor/pull/2546 - it seems that `subclasses()` isn't needed.

Issue #2256, opened 2023-05-05
Trait method not recognized when applied to an Enum


CalebDW:
Hello!

It seems that when a trait with a method is applied to an Enum, the method is not recognized as existing:

![image](https://user-images.githubusercontent.com/4176520/236446802-fec20aa8-0eeb-4d7c-a24c-c821188c6fa7.png)

Any thoughts?

Thanks!

CalebDW:
In addition, it seems that the code actions want to remove the traits altogether:

Prompt:

![image](https://user-images.githubusercontent.com/4176520/236447328-c910be99-4078-42b0-8a97-ead64c62c937.png)


After applying action:

![image](https://user-images.githubusercontent.com/4176520/236447401-8eb9899a-612b-4813-a355-a8a325579daa.png)

dantleech:
I guess we don't consider traits on Enums at all, so this is a bug/missing feature :+1:

mamazu:
Where would the best way to start with this be?

CalebDW:
Also looks like the native enum methods aren't recognized either (`::cases`, `::tryFrom`, `::from`).

mamazu:
@CalebDW Can you check if it is fixed with the most recent version of phpactor?

CalebDW:
@mamazu, the native PHP methods work (`::cases()`, `::from()`) but static and instance methods are still not being recognized.

mamazu:
Are you talking about static and instance methods via traits?

mamazu:
Thanks for the feedback. I will have to dig deeper then.

PR #2267, opened 2023-05-20
Adding indexing of use trait in enums

mamazu:
Refs #2256 

This might be all but I haven't tested if this also updates the completion and stuff.


dantleech:
which issue does this fix?

mamazu:
The linked issue: Traits not working in enums. I don't know if you need to change more stuff or if this change is sufficient.

PR #2279, opened 2023-05-29
Use the right word

greg0ire:


greg0ire:
Yes, it's indeed a well-known confusion but luckily, I am not… (•_•) ( •_•)>⌐■-■ (⌐■_■) … affected by it.

PR #2292, opened 2023-06-04
Adding support for the see tag in diagnostic providers

### LINKED ISSUE #2268: Reference `@see` (if exists) in deprecation diagnostic
Hello!

I love how Neovim has started showing deprecation warnings! However, the [standard](https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md#54-deprecated) recommends using `@see` when an element is superseded by another element and it would be great if the diagnostic included this information.

Thanks!

mamazu:
Fixes #2268


lib/DocblockParser/Ast/Tag/SeeTag.php
```diff
@@ -0,0 +1,28 @@
...
+
+use Phpactor\DocblockParser\Ast\TagNode;
+use Phpactor\DocblockParser\Ast\TextNode;
+use Phpactor\DocblockParser\Ast\Token;
+
+class SeeTag extends TagNode
+{
+    public const CHILD_NAMES = [
+        'token',
+        'text',
```
CalebDW, line 13: Note that [`@see`](https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md#513-see) has an FQSEN/URL and an optional description

lib/WorseReflection/Bridge/TolerantParser/Diagnostics/DeprecatedUsageDiagnosticProvider.php
```diff
@@ -220,6 +220,36 @@ public function foo(Closure $foobar) {
...
+        yield new DiagnosticExample(
+            title: 'Deprecate with replacement',
+            source: <<<'PHP'
+                <?php
+
+                class SomeClass {
+                    /**
+                    * @deprecated Will be removed
+                    *
+                    * @see SomeClass::new
```
dantleech, line 233: what implies that SomeClass::new is a replacement?

CalebDW: From the [standard](https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md#54-deprecated):

> If the associated element is superseded by another it is RECOMMENDED to add a @see tag in the same 'PHPDoc' pointing to the new element.

mamazu: That's by convention.

> If [the deprecated function] is superseded by another it is RECOMMENDED to add a @see tag in the same 'PHPDoc' pointing to the new element.

https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md#54-deprecated

lib/DocblockParser/Ast/Tag/SeeTag.php
mamazu: Yes, that is correct. But as far as I know we don't have a way to reference members here. So it might as well will be text.

lib/WorseReflection/Bridge/TolerantParser/Diagnostics/DeprecatedUsageDiagnosticProvider.php
dantleech: It's not really a standard.... but the "see" applies to the entire docblock, it's only coincidentally related to the deprecation message

CalebDW: So it is mentioned on the official website:

https://docs.phpdoc.org/3.0/guide/references/phpdoc/tags/deprecated.html#deprecated

In the case where there's an `@deprecated` and `@see` tag on the same element, then the `@see` is understood to point to the new element to replace the deprecated one

dantleech: Seems more coincidental to me, you can add "see" to any docblock. It doesn't mean that it references the replacement. It just means "see" this other thing, which may be the replacement or something else entirely.  It's certainty added context, would be good to see what PHPStorm does.

CalebDW: I had a coworker send me some screenshots of what PHPStorm does with `@deprecated` and `@see`:

Without `@see`:
image


With `@see`:
image

CalebDW, line 233: This should have parenthesis:
```suggestion
                    * @see SomeClass::new()
```

`SomeClass::new` points to a property `new` instead of the method `new()`

dantleech: Yes, so deprecation or not it will show "see also"?

CalebDW: Looks like it

dantleech: LSP has `relatedInformation` in the diagnostic:

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnostic

which could be used here - we can also specify `DiagnosticTag::DEPRECATED`

dantleech: in VS code it looks as you might want:

image

in Neovim the `DiagnostcTag::DEPRECATED` causes a strike-through, but the related information is not rendered:

image


but maybe it could be with some tweaks, in anycase, I think `$relatedInformation` is how we should render this info, and we should also parse the class name so that we can get the location to it.

mamazu: That sounds like the much better option. Then jumping to it also would be fairly easy.

Is that only a diagnostic feature. Because it can be written in any tag. It already is part of the method documentation as part of the doc block. If we want to parse the see tag everywhere, it would be nice to have a general "is related" list of references that the see tag could provide.

mamazu: True, I mean *technically* this is the value of the new call so if you wanted a function you technically have to do this: `SomeClass::new(...)`. But you are right, adding braces to signify it's the function makes sense

mamazu:
As a follow up to the fact that the see tag is a more general thing, should I repurpose the MR or create a new one?

lib/WorseReflection/Bridge/TolerantParser/Diagnostics/DeprecatedUsageDiagnosticProvider.php
mamazu: I haven't yet found the place where the WR diagnostics are converted to LSP diagnostics. Otherwise the relatedInfo thing should be easy to implement.

Issue #2296, opened 2023-06-17
Discover functions in files autoloaded with the autoload.files method in composer.json


rodrigoaguilera:
I use phpactor mainly for Drupal development with neovim and I updated phpactor from `master` of August 2022 to the 2023.4.10 version.
It worked alright even leaving the phpactor default configuration (meaning no special autoloading for Drupal) but after the update I got many errors about `Function "t" not found`.

This `t()` function lives in the `drupal/core` package included via 
```json
{
  "autoload": {
    "files": [
      "includes/bootstrap.inc"
    ]
  }
}
```
https://github.com/drupal/core/blob/10.0.0/composer.json#L106

I played a bit with the configuration of phpactor. First I tried `indexer.include_patterns` with
```
["\/**\/*.php","\/**\/*.inc"]
```
with no success.

Then I tried `"composer.class_maps_only": false` that makes the `t()` function to be included with autocompletion but I don't understand why.

I'm opening this issue as a mix of a bug report because I feel that there might be some regression there as the `t()` function was included back in August with no configuration and also as a support request since I don't understand `composer.class_maps_only` and what problems I might face when I set it to `false`. I'm I doing the right thing dealing with the oddities of Drupal?

dantleech:
I'm not sure what changed, but the indexer has never supported `.inc` files:

https://github.com/phpactor/phpactor/blob/2c9c6dc908f15b3b88fb1db104bce584cedcb711/lib/Indexer/Adapter/Filesystem/FilesystemFileListProvider.php#L31

Changing that to support `.inc` files solves the issue:

![image](https://github.com/phpactor/phpactor/assets/530801/abe999bb-323a-4461-a33b-52eb376cf12a)

We can make the supported PHP file extensions a configuration option I think.

rodrigoaguilera:
I went back to the commit of phpactor that I had from August and you are right, the function was not detected. I guess I'm noticing now because now I get I line prompt in neovim about the function not being found and before it was more silent about it. 

I also did a general upgrade of neovim from 0.7 to 0.9 and all of its plugins.

So the indexer and the composer autoloading work at the same time?
By setting `composer.class_maps_only` to `false` am I including more files to be considered for the language server?

dantleech:
https://github.com/phpactor/phpactor/compare/gh-2296 should do the trick, will merge later.

rodrigoaguilera:
Wow. Amazing. I will go back to using `master` once is merged and try.

[...]

PR #2305, opened 2023-06-18
Update Drupal 8+ docs.

rodrigoaguilera:
As I promised in https://github.com/phpactor/phpactor/issues/2296 I tried the new release but it didn't work 100% because the include patterns also need to be tweaked. I don't know why it worked for @dantleech 

It seems like something is not quite right when two config items need to be configured in a very similar fashion to include a new file extension but I can't think of a way right now of only using `indexer.include_patterns`.

There is a plan in Drupal to have all php files with the php extension but being such a big project I think it is a long way there.
https://www.drupal.org/project/drupal/issues/7269
There is also `.install`, `.theme`, `.profile` and `.engine` but I think those can be ignored.

### DIFF
--- a/doc/integrations/drupal8.rst
+++ b/doc/integrations/drupal8.rst
@@ -1,15 +1,19 @@
-Drupal 8
-========
+Drupal 8+
+=========
 
-Inc Files
-^^^^^^^^^
+Inc and module files
+^^^^^^^^^^^^^^^^^^^^
 
-By default Phpactor will not index `.inc` files.
+By default Phpactor will not index `.inc` nor `.module` files.
 
-Run the following on your project to enable the indexing of `.inc` files.
+Run the following on your project to enable the indexing of `.inc` and `.module` files.
 
 ```
-phpactor config:set indexer.supported_extensions '["php", "inc"]'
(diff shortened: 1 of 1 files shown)


doc/integrations/drupal8.rst
```diff
@@ -1,15 +1,19 @@
...
-Run the following on your project to enable the indexing of `.inc` files.
+Run the following on your project to enable the indexing of `.inc` and `.module` files.
 
 ```
-phpactor config:set indexer.supported_extensions '["php", "inc"]'
+phpactor config:set indexer.supported_extensions '["php", "inc", "module"]'
+```
+
+```
+phpactor config:set indexer.include_patterns '["/**/*.php", "/**/*.inc", "/**/*.module"]'
```
dantleech, line 16: hmm, I think I fixed this in the wrong way :flushed: with the current logic we first reduce the set of files to those with the "supported_extensions" and then again with the "include_patterns", so the "supported_extensions" wasn't strictly necessary.

dantleech:
thanks @rodrigoaguilera . I think the indexing stuff needs another iteration at some point (use multiple processes, support multiple workspaces) so maybe this can be improved then.

PR #2321, opened 2023-07-12
Fix typo: chown becomes chmod

Einenlum:
I think you meant chmod here. a+x is not a compatible parameter when changing owner.


dantleech:
You, merci!

PR #2324, opened 2023-07-16
Gh 2320: Add passing test

dantleech:


mamazu:
The issue for this is closed? This should be fixed, right?

PR #2325, opened 2023-07-17
Add parameter to disable LanguageServerWorseReflection diagnostics.

shello:
Disabling `WorseReflection` diagnostics may be useful when using a different diagnostics extension (or when diagnostics are just not wanted).

This short patch adds that ability with a new parameter `language_server_worse_reflection.diagnostics.enable`.


dantleech (approved):

dantleech:
thanks!

PR #2330, opened 2023-08-04
Do not report undeclared variables that are passed by reference as undefined

### LINKED ISSUE #2329: Undefined variable when dynamically declared and passed by ref
Phpactor marks variables as undefined if they are dynamically declared when passed by reference to a function:

```php
function foo(?int &$value): void
{
    $value = 123;
}

foo($num); // phpactor: Undefined variable "$num"
```

Curiously, no error is raised for built-in PHP functions that do the same, like `preg_match()` and its 3rd argument:

```php
preg_match('/^php(.*)$/', 'phpactor', $matches);
```

I couldn't find any documentation on config options to tweak this behavior. Is this intended behavior?

---

Sales pitch:
I understand that this form of function output is archaic 😅 but I have legitimate reasons and I really need to suppress these errors. I have hundreds of them in my code-base and they are over-shadowing other more important diagnostics.

mecha:
Closes #2329

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Bug fixes:
   - Fix generic extends with templated argument #2295
   - Do not report statically declared variables as undefined #2311
   - Fix PHP linter #2318
+  - Do not report undeclared variables that are passed by reference as undefined #2329 @mecha
 
 ## 2023.06.17-1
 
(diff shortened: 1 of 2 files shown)


mecha:
It should be mentioned that this PR only fixes the false-negative diagnostic messages for function calls and method calls. I couldn't figure out how to fix it for closures. Example:

```php
$foo = function(?int &$v1, ?string &$v2) {
	$v1 = 123;
	$v2 = '456';
}

$foo($var1, $var2);
```

The above still produces 2 diagnostic messages for `$var1` and `$var2`.

dantleech (approved):

PR #2333, opened 2023-08-06
Do not promote parameters from the parent constructor

mamazu:
This stops the promotion of parameters in the constructor that have the same name as the properties in the parent constructor. This does not work when the names differ.

Refs #2119

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,16 @@
 Changelog
 =========
 
+## master
+
+  - Do not promote parameters that are used in parent constructor #2119
+
 ## 2023.08.06
 
 Improvements:
 
   - Improve Diagnostics: Run linters in parallel #2327
   - Index documents on save #2326
-  - Do not promote parameters that are used in parent constructor #2119
 
 Bug fixes:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Improvements:
 
   - Improve Diagnostics: Run linters in parallel #2327
   - Index documents on save #2326
+  - Do not promote parameters that are used in parent constructor #2119
 
 Bug fixes:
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ Improvements:
 
   - Improve Diagnostics: Run linters in parallel #2327
   - Index documents on save #2326
+  - Do not promote parameters that are used in parent constructor #2119
```
dantleech, line 10: this is the first feature in a new release now, you can add a new `## master` section :)

dantleech (approved):

mamazu:
I moved the changelog entry to the master, is the implementation of "parameters with the same name" good enough? Or do you have a better idea?

lib/CodeTransform/Adapter/WorseReflection/Transformer/CompleteConstructor.php
```diff
@@ -130,14 +130,41 @@ private function transformPromote(SourceCode $source): TextEdits
...
         return TextEdits::fromTextEdits($edits);
     }
 
+    /**
+    * Get the names of the parent constructor to know which parameters should not be promoted.
+    * @return array<string>
+    */
+    private function getParentClassParamaterNames(ReflectionClass $class): array
+    {
+        $ancestor = $class->ancestors()->firstOrNull();
```
dantleech, line 151: `parent()`? ancestors includes interfaces etc.

mamazu: Well we probably need a loop in here to see if the classes even higher up have constructors as well like so:
```php
class B {
    public function __construct(private string $a) {}
}
class A extends B {}

class Foo extends A {
    public function __construct(string $a) {parent::__construct($a);}
}
```

Then it wouldn't pick up on the constructor right?

dantleech: If you call `parent()->methods()->get('__construct')` it should return the first constructor in the heirarchy starting from the parent

mamazu: Added a test for it and it seems to pass.

dantleech:
Ta!

mamazu:
You also want to close the issue for that? #2119

dantleech:
Yep thanks

PR #2343, opened 2023-08-19
Generate @param for `iterable` type

### LINKED ISSUE #2084: Generate @param refactoring not triggered for `iterable`
image

mamazu:
Closes #2084

### DIFF
--- a/lib/CodeTransform/Tests/Adapter/WorseReflection/Transformer/UpdateDocblockReturnTransformerTest.php
+++ b/lib/CodeTransform/Tests/Adapter/WorseReflection/Transformer/UpdateDocblockReturnTransformerTest.php
@@ -383,6 +383,49 @@ class UpdateDocblockReturnTransformerTest extends WorseTestCase
                 EOT
         ];
 
+        yield 'add docblock for iterables' => [
+            <<<'EOT'
+                <?php
+
+                class IFoo {
+                    public function bar(): iterable
+                    {
+                        return $this->baz();
+                    }
+
+                    /**
+                    * @return iterable<int>
+                    */
+                    public function baz(): iterable
+                    {
+                        yield 22;
+                    }
+                }
+                EOT
+            ,
+            <<<'EOT'
+                <?php
+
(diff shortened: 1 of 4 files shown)


lib/CodeTransform/Tests/Adapter/WorseReflection/Transformer/UpdateDocblockReturnTransformerTest.php
```diff
@@ -387,34 +387,44 @@ public function baz(): array
...
                     /**
-                     * @return mixed
-                     */
-                    private function foo() {}
+                    * @return iterable<int>
+                    */
+                    public function baz(): iterable
+                    {
+                        yield 22;
+                    }
```
mamazu, line 424: This is a new test.

lib/WorseReflection/Bridge/TolerantParser/Diagnostics/DocblockMissingReturnTypeProvider.php
```diff
@@ -74,12 +73,16 @@ public function exit(NodeContextResolver $resolver, Frame $frame, Node $node): i
             return;
         }
 
-        if (
-            $claimedReturnType->isDefined() && !$claimedReturnType->isClass() && !$claimedReturnType->isArray() && !$claimedReturnType->isClosure()
+        if ($claimedReturnType->isDefined()
+            && !$claimedReturnType->isClass()
+            && !$claimedReturnType->isArray()
+            && !$claimedReturnType->isClosure()
+            && !$claimedReturnType->isIterable()
```
mamazu, line 80: It would probably be easier to list the cases where it's true. Like scalar types?

dantleech (approved):

dantleech:
thanks! although I'm not sure what the original issue was as it was only a truncated screenshot. Probably this is the fix :sweat_smile:

mamazu:
The problem was that if you're function return type was iterable it didn't suggest generating a docblock for the specific type. Like it does with arrays.

PR #2347, opened 2023-08-22
Improve detection of Xdebug to avoid false positivity

bart-jaskulski:
Besides loaded PHP extensions, Xdebug may be disabled either by `xdebug.mode=off` or environment variable `XDEBUG_MODE`. Relying on `composer/xdebug-handler` helps to respect those conditions.


dantleech (approved):

dantleech:
great, thanks :tada:

Issue #2064, opened 2023-01-30
`textDocument/references` returns same position for start and end


smjonas:
Originally reported as an issue on this Neovim plugin repo: https://github.com/smjonas/inc-rename.nvim/issues/30

When sending a `textDocument/references` request for a given identifier, the server always responds with a result like this (formatted as a Lua table):
```lua
{
    range = {
      ["end"] = {
        character = 8,
        line = 15
      },
      start = {
        character = 8,
        line = 15
      }
    },
    uri = "file:///some_file.php"
 }
``` 
Notice how the `start` and `end` characters are the same.

## Expected result
```lua
{
    range = {
      ["end"] = {
        character = 11, -- (assuming the variable is 4 characters long)
        line = 15
      },
      start = {
        character = 8,
        line = 15
      }
    },
    uri = "file:///some_file.php"
 }
```

dantleech:
confirmed, thanks for reporting :+1:

dantleech:
For reference it's because the underlying implementation does not record the "end" positions of the references, could be fixed by working back from here: https://github.com/phpactor/phpactor/blob/ab023ca8143a5253673d394cb255ec1420094b45/lib/Extension/LanguageServerBridge/Converter/LocationConverter.php#L43

mamazu:
Can be closed as the feature made it into the master. :partying_face:

dantleech:
🎉

PR #2352, opened 2023-08-23
Using ranges for locations

mamazu:
Refs #2064


mamazu:
So it looks like go to definition expects a location and not a range although we have the range of the definition of a function. Should I just use the start position or could it be that the language server also allows to go to a range?

dantleech:
thanks for this. just to note that this is a large piece of work and I don't necessarily have the time to give it the attention it needs.

> could it be that the language server also allows to go to a range?

yes it does the LS `Location` object has a range in it: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_declaration

lib/TextDocument/LocationRange.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+namespace Phpactor\TextDocument;
+
+class LocationRange
```
dantleech, line 5: do we need to introduce this? `Location` can have the `$range` as with the LSP Location

lib/TextDocument/LocationRanges.php
```diff
@@ -0,0 +1,89 @@
...
+
+use ArrayIterator;
+use Countable;
+use IteratorAggregate;
+use RuntimeException;
+
+/**
+ * @implements IteratorAggregate<LocationRange>
+ */
+final class LocationRanges implements IteratorAggregate, Countable
```
dantleech, line 13: as with the above, I doubt we need this. the `Locations` collection should be sufficient

lib/TextDocument/LocationRange.php
mamazu: Should I extend the Phpactor `Location` class for this? Because it only as a URI and an offset. And I don't think that we want to replace the Phpactor `Location` with the LSP `Location` class right?

dantleech: So a Location should probably have a range and not refer to a single offset, so this is why this may be a can of worms.

mamazu: Sounds easy enough. Then I will update the `Location` class of phpactor to always use a range. This is the cleaner solution anyways if you ask me. That way you are forced to provide the size of the text object you are referencing.

mamazu:
Should be done now. I have replaced the `Location`'s byteoffset with a ByteoffsetRage. Fixed the phpstan errors and the tests. And the editor doesn't crash when trying to go to reference or 

Things that could still be done:
* The test (`lib/WorseReferenceFinder/Tests/Unit/TolerantVariableReferenceFinderTest.php`) only checks the start positions of the location (to keep backwards compatibility). It's much uglier than before, since we have to check everything manually.

lib/Extension/Behat/Behat/Step.php
```diff
@@ -59,8 +60,13 @@ public function path(): string
...
-    public function byteOffset(): int
+    public function start(): int
     {
         return $this->startByteOffset;
     }
+
+    public function end(): int
+    {
+        return $this->endByteOffset;
+    }
```
dantleech, line 71: this would be better as a `ByteOffsetRange` and a single method `byteOffsetRange()` or similar

lib/Extension/LanguageServerReferenceFinder/Tests/Unit/Handler/TypeDefinitionHandlerTest.php
```diff
@@ -50,24 +46,19 @@ public function testGoesToSingleType(): void
...
-                )
+                PhpactorLocation::fromPathAndOffsets(self::EXAMPLE_URI, 2, 2),
             ),
             new TypeLocation(
                 TypeFactory::class('Barfoo'),
-                new PhpactorLocation(
-                    TextDocumentUri::fromString(self::EXAMPLE_URI),
-                    ByteOffset::fromInt(2)
-                )
+                PhpactorLocation::fromPathAndOffsets(self::EXAMPLE_URI, 2, 2),
```
dantleech, line 61: maybe add `fromPathAndOffset` for locations at a single offset

lib/Indexer/Adapter/ReferenceFinder/IndexedReferenceFinder.php
```diff
@@ -227,8 +227,9 @@ private function newObjectReferences(string $containerType): Generator
                     continue;
                 }
                 yield LocationConfidence::surely(
-                    Location::fromPathAndOffset(
+                    Location::fromPathAndOffsets(
                         $file->filePath() ?? '',
+                        $fileReference->offset(),
```
dantleech, line 232: `fromPathAndOffset`. does the indexer not support ranges?

lib/Indexer/Model/Query/ClassQuery.php
```diff
@@ -50,7 +50,11 @@ public function referencesTo(string $identifier): Generator
 
             foreach ($fileRecord->references()->to($record) as $classReference) {
                 yield LocationConfidence::surely(
-                    Location::fromPathAndOffset($fileRecord->filePath(), $classReference->offset())
+                    Location::fromPathAndOffsets(
+                        $fileRecord->filePath() ?? '',
+                        $classReference->offset(),
+                        $classReference->offset(),
```
dantleech, line 56: ditto why no range?

lib/Indexer/Model/Query/FunctionQuery.php
```diff
@@ -34,7 +35,11 @@ public function referencesTo(string $identifier): Generator
 
             foreach ($fileRecord->references()->to($record) as $functionReference) {
                 yield LocationConfidence::surely(
-                    Location::fromPathAndOffset($fileRecord->filePath(), $functionReference->offset())
+                    Location::fromPathAndOffsets(
+                        TextDocumentUri::fromString($fileRecord->filePath()),
+                        $functionReference->offset(),
+                        $functionReference->offset()
```
dantleech, line 41: ...

lib/Indexer/Model/Query/MemberQuery.php
```diff
@@ -55,7 +55,11 @@ public function referencesTo(string $type, string $memberName, ?string $containe
                     $memberReference = $this->enhancer->enhance($fileRecord, $memberReference);
                 }
 
-                $location = Location::fromPathAndOffset($fileRecord->filePath(), $memberReference->offset());
+                $location = Location::fromPathAndOffsets(
+                    $fileRecord->filePath() ?? '',
+                    $memberReference->offset(),
+                    $memberReference->offset()
```
dantleech, line 61: ...

lib/Rename/Adapter/ReferenceFinder/AbstractReferenceRenamer.php
```diff
@@ -78,14 +78,15 @@ protected function renameEdit(Location $location, ?ByteOffsetRange $range, strin
     {
         $referenceDocument = $this->locator->get($location->uri());
 
-        $range = $this->getRenameRange($referenceDocument, $location->offset());
+        // We could probably use the complete range here instead of only using the location
+        $range = $this->getRenameRange($referenceDocument, $location->range()->start());
```
dantleech, line 82: not sure what this comment means, let's remove it.

lib/Rename/Tests/Adapter/ReferenceFinder/MemberRenamerTest.php
```diff
@@ -114,12 +114,18 @@ public function testRename(string $source): void
 
         $renamer = new MemberRenamer(
             new PredefinedReferenceFinder(...array_map(function (ByteOffset $reference) use ($textDocument) {
-                return PotentialLocation::surely(new Location($textDocument->uri(), $reference));
+                return PotentialLocation::surely(new Location(
+                    $textDocument->uri(),
+                    ByteOffsetRange::fromByteOffsets($reference, $reference)
```
dantleech, line 119: `fromByteOffset` and below

lib/Rename/Tests/Adapter/ReferenceFinder/MemberRenamerTest.php
```diff
@@ -114,12 +114,18 @@ public function testRename(string $source): void
...
+                ));
             }, $references)),
             InMemoryDocumentLocator::fromTextDocuments([$textDocument]),
             new Parser(),
             new PredefiniedImplementationFinder(new Locations(array_map(function (ByteOffset $reference) use ($textDocument) {
-                return new Location($textDocument->uri(), $reference);
+                return new Location(
+                    $textDocument->uri(),
+                    ByteOffsetRange::fromByteOffsets($reference, $reference)
```
dantleech, line 127: fromByteOffset

lib/Rename/Tests/Adapter/ReferenceFinder/ReferenceRenamerIntegrationTestCase.php
```diff
@@ -18,7 +19,10 @@ abstract class ReferenceRenamerIntegrationTestCase extends TestCase
     public function offsetsToReferenceFinder(TextDocument $textDocument, array $references): ReferenceFinder
     {
         return new PredefinedReferenceFinder(...array_map(function (ByteOffset $reference) use ($textDocument) {
-            return PotentialLocation::surely(new Location($textDocument->uri(), $reference));
+            return PotentialLocation::surely(new Location(
+                $textDocument->uriOrThrow(),
+                ByteOffsetRange::fromByteOffsets($reference, $reference)
```
dantleech, line 24: fromByteOffset

lib/Rename/Tests/Adapter/ReferenceFinder/VariableRenamerTest.php
```diff
@@ -125,11 +125,16 @@ public function testRename(string $source): void
         $renamer = $this->createRenamer(
             array_map(
                 function (ByteOffset $reference) use ($textDocument) {
-                    return PotentialLocation::surely(new Location($textDocument->uriOrThrow(), $reference));
+                    return PotentialLocation::surely(
+                        new Location(
+                            $textDocument->uriOrThrow(),
+                            ByteOffsetRange::fromByteOffsets($reference, $reference)
```
dantleech, line 131: fromByteOffset

lib/TextDocument/Tests/Unit/LocationAssertions.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Phpactor\TextDocument\Tests\Unit;
+
+use PHPUnit\Framework\TestCase;
+use Phpactor\TextDocument\Location;
+
+trait LocationAssertions
```
dantleech, line 10: anything in the `Tests` folder is private to this package and should not be used by other packages.

for it to be used by other packges it should be in `lib/Test` but:
```
self::assertEquals($location, Location::fromPathAndOffsets('foobar', 1, 2)
```
might be better anyway?

lib/WorseReferenceFinder/Tests/Unit/TolerantVariableReferenceFinderTest.php
```diff
@@ -31,7 +32,15 @@ public function testReferences(string $source, bool $includeDefinition = false):
         $finder = new TolerantVariableReferenceFinder(new Parser(), $includeDefinition);
         $actualReferences = iterator_to_array($finder->findReferences($document, ByteOffset::fromInt($selectionOffset)), false);
 
-        $this->assertEquals($expectedReferences, $actualReferences);
+        $this->assertEquals(count($expectedReferences), count($actualReferences));
+        foreach($expectedReferences as $index => $reference) {
+            // We need to manually compare objects here because the end location should not be checked. There should be a second marker
```
dantleech, line 37: remove comment

dantleech (commented):
thanks, left some comments. it's a very big PR and I will need to test this locally for some time. it seems we don't use ranges on the reference finder for some reason?

lib/Extension/Behat/Behat/Step.php
mamazu: I used the location object. That made things simpler anyways.

lib/Indexer/Adapter/ReferenceFinder/IndexedReferenceFinder.php
mamazu: Range has been added

lib/Indexer/Model/Query/ClassQuery.php
mamazu: Range has been added

lib/Indexer/Model/Query/FunctionQuery.php
mamazu: Range has been added

lib/Indexer/Model/Query/MemberQuery.php
mamazu: Range has been added

lib/TextDocument/Tests/Unit/LocationAssertions.php
mamazu: Yeah that would have been too obvious for me to spot, of course I'll refactor it.

lib/Rename/Adapter/ReferenceFinder/AbstractReferenceRenamer.php
mamazu: The question here was, we already get a range passed into the function. Why is that not the same as the correct rename range?

lib/WorseReferenceFinder/Tests/Unit/TolerantVariableReferenceFinderTest.php
mamazu: Done

lib/Rename/Tests/Adapter/ReferenceFinder/VariableRenamerTest.php
mamazu: Done

lib/Rename/Tests/Adapter/ReferenceFinder/ReferenceRenamerIntegrationTestCase.php
mamazu: Done

lib/Rename/Tests/Adapter/ReferenceFinder/MemberRenamerTest.php
mamazu: Done

lib/Rename/Tests/Adapter/ReferenceFinder/MemberRenamerTest.php
mamazu: Done

lib/Extension/LanguageServerReferenceFinder/Tests/Unit/Handler/TypeDefinitionHandlerTest.php
mamazu: Done.

mamazu:
Well, I opened up a follow up issue #2363 and commented on one of your comments. Otherwise the changes have been quite stable. I am using the version for development right now and it's not crashing after rebuilding all indices.

lib/Rename/Adapter/ReferenceFinder/AbstractReferenceRenamer.php
dantleech: because the range passed in is the user input not the range of the text that should be replaced

dantleech:
I think that issue needs to be solved now by making $End nullable and defaulting to $start otherwise we'll have lots of crashes after merging this?

lib/Rename/Adapter/ReferenceFinder/AbstractReferenceRenamer.php
mamazu: Ah yeah, will be removed then.

mamazu:
Yes, they all go away when reindexing the project obviously. But your backwards compatible solution doesn't sound too hard.

lib/Indexer/Model/Record/FileRecord.php
```diff
@@ -58,7 +58,8 @@ public function addReference(RecordReference $reference): self
         $this->references[] = [
             $reference->type(),
             $reference->identifier(),
-            $reference->offset(),
+            $reference->start(),
+            $reference->end(),
```
mamazu, line 62: This change might be a little tricky to work around though when deserializing.

lib/Indexer/Model/Record/FullyQualifiedReferenceTrait.php
```diff
@@ -45,6 +53,11 @@ public function start(): ByteOffset
         return ByteOffset::fromInt($this->start);
     }
 
+    public function end(): ByteOffset
```
bart-jaskulski, line 56: I think, both `start` and `end` function should belong to `HasLocation` interface, as keeping it as a part of `FullyQualifiedReferenceTrait` is a bit misleading. Without this, methods aren't backed by any interface and an only reliable way to ensure, the methods are there is using assertions (which are heavily used around indexer).

Although, it may seem irrelevant, I found this itching, while working on WordPress hooks indexer, which (the hooks) are basically the first argument of a function call. It makes it a good candidate for a location data, but it doesn't have any *fqn*, what complicates the use of `FullyQualifiedReferenceTrait`

mamazu: Yeah that would definitely make sense.

dantleech:
I'm assuming this still needs to be made backwards compatible, I haven't got a great deal of time atm but I can try and look at it later today or next week

mamazu:
If you have a little time, wouldn't it be better to have a general solution for this issue? Currently the language server crashes when the index could not be read and deserialized properly. My proposal would be if the document can not be deserialized just re-index the document and try again? That way we don't need to have a BC layer that is possibly going to stay there forever.

bart-jaskulski:
Enforcing re-index seems reasonable to me (for instance, PhpStorm is re-indexing/validating files *each time* you open the project). Yet, there's one condition – the index should be built from scratch, including stub files. By default, Phpactor refreshes index only for project files, but in such case it would be the best to include dependencies as well. I think, without this you will be left with crashing LS, no matter what because files outside the project would be invalid. This would also be more user-friendly, skipping the alert for manual re-indexing.

If the above would be covered, I think re-indexing on failure is more future-proof resolution than burden of keeping BC layer.

dantleech:
I see that adding a B/C layer here isn't as easy as I thought. I think a better solution is to version the index "scheme" and force a reindex when it changes. I will implement this now.

dantleech:
ok, i've had a chance to test this.

what does it support?

- :heavy_check_mark: methods
- :heavy_check_mark:  constants
- :heavy_check_mark: properties
- :no_entry: class references
- :no_entry: implemenations
- anything else?

see notes below:


## Find implementatoins:

Returns range for entire class:

```
      {
        "uri": "file:///home/daniel/www/phpactor/phpactor/lib/Indexer/Model/Record/ClassRecord.php",
        "range": {
          "start": {
            "line": 7,
            "character": 0
          },
          "end": {
            "line": 114,
            "character": 1
          }
        }
      },
 ```

## References

Returns single offset locations still:

```
      {
        "uri": "file:///home/daniel/www/phpactor/phpactor/lib/Extension/LanguageServerRename/Tests/Extension/TestExtension.php",
        "range": {
          "start": {
            "line": 18,
            "character": 29
          },
          "end": {
            "line": 18,
            "character": 29
          }
        }
      },
 ```

Additionally I've added B/C below and made an [MR](https://github.com/phpactor/phpactor/pull/2370) (which is not dependent on this) to introduce index versioning so the index will be rebuilt for changes like this. 

```diff
diff --git a/lib/Indexer/Adapter/ReferenceFinder/IndexedReferenceFinder.php b/lib/Indexer/Adapter/ReferenceFinder/IndexedReferenceFinder.php
index 3dec734a..f969e13a 100644
--- a/lib/Indexer/Adapter/ReferenceFinder/IndexedReferenceFinder.php
+++ b/lib/Indexer/Adapter/ReferenceFinder/IndexedReferenceFinder.php
@@ -105,8 +105,6 @@ class IndexedReferenceFinder implements ReferenceFinder
                 return;
             }
 
-
-
             // note that we check the all implementations: this will multiply
             // the number of NOT and MAYBE matches
             foreach ($this->implementationsOf($containerType) as $containerType) {
diff --git a/lib/Indexer/Model/Record/FileRecord.php b/lib/Indexer/Model/Record/FileRecord.php
index 4cb8ffce..5ebd5377 100644
--- a/lib/Indexer/Model/Record/FileRecord.php
+++ b/lib/Indexer/Model/Record/FileRecord.php
@@ -59,9 +59,9 @@ class FileRecord implements HasPath, Record
             $reference->type(),
             $reference->identifier(),
             $reference->start(),
-            $reference->end(),
             $reference->contaninerType(),
-            $reference->flags()
+            $reference->flags(),
+            $reference->end(),
         ];
 
         return $this;
diff --git a/lib/Indexer/Model/Record/FullyQualifiedReferenceTrait.php b/lib/Indexer/Model/Record/FullyQualifiedReferenceTrait.php
index e2b92315..a91aff74 100644
--- a/lib/Indexer/Model/Record/FullyQualifiedReferenceTrait.php
+++ b/lib/Indexer/Model/Record/FullyQualifiedReferenceTrait.php
@@ -12,7 +12,7 @@ trait FullyQualifiedReferenceTrait
 
     private int $start;
 
-    private int $end;
+    private ?int $end = null;
 
     public function __construct(string $fqn)
     {
@@ -55,7 +55,7 @@ trait FullyQualifiedReferenceTrait
 
     public function end(): ByteOffset
     {
-        return ByteOffset::fromInt($this->end);
+        return $this->end ? ByteOffset::fromInt($this->end) : $this->start();
     }
 
     public function identifier(): string
diff --git a/lib/Indexer/Model/RecordReference.php b/lib/Indexer/Model/RecordReference.php
index 451daf19..6258cdd4 100644
--- a/lib/Indexer/Model/RecordReference.php
+++ b/lib/Indexer/Model/RecordReference.php
@@ -8,15 +8,18 @@ use Phpactor\Indexer\Model\Record\HasFlagsTrait;
 class RecordReference implements HasFlags
 {
     use HasFlagsTrait;
-    const FLAG_NEW_OBJECT = 1;
+    public const FLAG_NEW_OBJECT = 1;
 
+    /**
+     * Order matters here for B/C, new parameters must be added after old ones.
+     */
     public function __construct(
         private string $type,
         private string $identifier,
         private int $start,
-        private int $end,
         private ?string $contaninerType = null,
-        int $flags = 0
+        int $flags = 0,
+        private ?int $end = null,
     ) {
         $this->flags = $flags;
     }
@@ -28,7 +31,7 @@ class RecordReference implements HasFlags
 
     public function end(): int
     {
-        return $this->end;
+        return $this->end ?? $this->start;
     }
 
     public function identifier(): string
@@ -47,12 +50,26 @@ class RecordReference implements HasFlags
         int $end,
         ?string $containerType
     ): self {
-        return new self($record->recordType(), $record->identifier(), $start, $end, $containerType);
+        return new self(
+            $record->recordType(),
+            $record->identifier(),
+            $start,
+            $containerType,
+            0,
+            $end,
+        );
     }
 
     public function withContainerType(string $type): self
     {
-        return new self($this->type, $this->identifier, $this->start, $this->end, $type);
+        return new self(
+            $this->type,
+            $this->identifier,
+            $this->start,
+            $type,
+            $this->flags,
+            $this->end,
+        );
     }
 
     public function contaninerType(): ?string
```

mamazu:
The issue with the class references is now fixed. This was due to the ordering of the parameters change.

The problem with the "find implementation" I didn't really get. What is it supposed to return there if not a range to the entire class?

dantleech:
It should return the range of the name of the class.

mamazu:
Okay I think I found the place. in the `lib/Indexer/Adapter/ReferenceFinder/IndexedImplementationFinder.php` Lines 67 to 79. 

There it basically goes and iterates over all classes that implement a thing and takes the record start and the record end as it's respective location. The question is should we add add a property to the `ClassRecord` to also index the class name position or just get the class name position in the `IndexedImplementationFinder`?

dantleech:
The indexer should index the start and end of the class name e.g. `class <>Foobar<> extends Barfoo`

mamazu:
Done it.

dantleech:
Thanks @mamazu let's see how it goes :tada:

PR #2353, opened 2023-08-25
Add PHP_CodeSniffer extension

bart-jaskulski:
Based on php-cs-fixer extension, add another diagnostics provider with formatting support: [PHP_CodeSniffer](https://github.com/squizlabs/PHP_CodeSniffer).

This PR includes some minor factorization of current php-cs-fixer extension, with the main focus on moving utils to own namespace (`Phpactor/Diff` at the moment, but this may be a subject to change if some namespace would be more suitable). On top of that, there's a *copy* of the extension adapted to `phpcs` support.


lib/Extension/LanguageServerPhpCodeSniffer/Provider/PhpCodeSnifferDiagnosticsProvider.php
```diff
@@ -0,0 +1,209 @@
...
+                    $diagnostics[] = yield $this->createRuleDiagnostics($rule, $range);
+                }
+            }
+
+            return $diagnostics;
+        });
+    }
+
+    /**
+     * @return Promise<Diagnostic>
```
bart-jaskulski, line 179: Phpstan complains about this return, claiming that `\Amp\call` may return `Promise<mixed>` instead of declared type, but the same code for creating rule diagnostic is in php-cs-fixer extension and is not either silenced from phpstan or reported by it.

Any suggestions to help with my confusion would be appeciated.

lib/Extension/LanguageServerPhpCodeSniffer/Exception/PhpCodeSnifferError.php
```diff
@@ -0,0 +1,21 @@
...
+
+use RuntimeException;
+
+class PhpCodeSnifferError extends RuntimeException
+{
+    public function __construct(int $exitCode, string $command, string $stderr, string $stdout)
+    {
+        parent::__construct(
+            sprintf(
+                "phpcs exited with code '%s'; cmd: %s; stderr: '%s'; stdout: '%s'",
```
dantleech, line 13: in general I'd prefer this to just be `new RuntimeException(sprintf('...', ...))` unless there is a reason to have a dedicated runtime exception for this.

lib/Extension/LanguageServerPhpCodeSniffer/LanguageServerPhpCodeSnifferExtension.php
```diff
@@ -0,0 +1,97 @@
+<?php
+
+namespace Phpactor\Extension\LanguageServerPhpCodeSniffer;
```
dantleech, line 3: Let's just call this `Extension\PhpCodeSniffer` and remove the `LanguageServer` prefix. In hindsight the other extensions shouldn't have the `LanguageServer` prefix either (these extensions should encapsulate all functionality for the integration, not just for "language server" even if in practice that's what they do)

lib/Extension/LanguageServerPhpCodeSniffer/LanguageServerPhpCodeSnifferExtension.php
```diff
@@ -0,0 +1,97 @@
...
+use Phpactor\Extension\Logger\LoggingExtension;
+use Phpactor\LanguageServer\Core\Server\ClientApi;
+use Phpactor\MapResolver\Resolver;
+
+class LanguageServerPhpCodeSnifferExtension implements OptionalExtension
+{
+    public const PARAM_PHP_CODE_SNIFFER_BIN = 'language_server_php_code_sniffer.bin';
+    public const PARAM_ENV = 'language_server_php_code_sniffer.env';
+    public const PARAM_SHOW_DIAGNOSTICS = 'language_server_php_code_sniffer.show_diagnostics';
+    public const PARAM_ENABLED = 'language_server_php_code_sniffer.enabled';
```
dantleech, line 25: same here, just `php_code_sniffer.enabled` etc

lib/Extension/LanguageServerPhpCodeSniffer/LanguageServerPhpCodeSnifferSuggestExtension.php
```diff
@@ -0,0 +1,50 @@
...
+use Phpactor\Configurator\Model\Changes;
+use Phpactor\Configurator\Model\JsonConfig;
+use Phpactor\Container\Container;
+use Phpactor\Container\ContainerBuilder;
+use Phpactor\Container\Extension;
+use Phpactor\Extension\Configuration\ChangeSuggestor\PhpactorComposerSuggestor;
+use Phpactor\Extension\Configuration\ConfigurationExtension;
+use Phpactor\MapResolver\Resolver;
+
+class LanguageServerPhpCodeSnifferSuggestExtension implements Extension
```
dantleech, line 16: ditto r.e. the prefix

lib/Extension/LanguageServerPhpCodeSniffer/Provider/PhpCodeSnifferDiagnosticsProvider.php
dantleech: probably ignored in the baseline. Does adding a return type to the anonymous function help? `use (...): Diagnostic`

dantleech:
:+1: looks good, just some renaming etc, I'll have to test it before merging.

lib/Extension/LanguageServerPhpCodeSniffer/Provider/PhpCodeSnifferDiagnosticsProvider.php
bart-jaskulski: I blindfolded copied php-cs-fixer, which was `yield`ing rules description. This `yield` made a difference. As we don't need anything async in phpcs, I just return object directly and phpstan is happy again (keeping code works as intended).

composer.json
```diff
@@ -47,7 +47,8 @@
         "jangregor/phpstan-prophecy": "^1.0",
         "phpstan/extension-installer": "^1.1",
         "vimeo/psalm": "^4.22",
-        "blackfire/php-sdk": "^1.31"
+        "blackfire/php-sdk": "^1.31",
+        "squizlabs/php_codesniffer": "^3.7"
```
dantleech, line 51: we should really stop adding these dev dependencies as they provide limited protection from breakage (only an arbitrary major version in our lock file) and would be better to have a separate pipeline for them probably. anyway, at least `phpcs` doesn't have any dependencies itself.

dantleech:
sorry, was going to merge rebase and merge this but on testing I had some issues:

- it crashes if it can't find the bin or if the bin exits (which happened on my first go on a very old project)
- testing a project with `phpcs 3.7.1` the output is colored by default and I needed to add the `--no-colors` flag for fomratting to work.
- there are no diagnostics (might need to dig into that one)

the diff below fixes the first two issues, I'm not sure about the diagnostics.

```diff
diff --git a/lib/Extension/PhpCodeSniffer/Model/PhpCodeSnifferProcess.php b/lib/Extension/PhpCodeSniffer/Model/PhpCodeSnifferProcess.php
index 43d05d0fb..6a4b9f555 100644
--- a/lib/Extension/PhpCodeSniffer/Model/PhpCodeSnifferProcess.php
+++ b/lib/Extension/PhpCodeSniffer/Model/PhpCodeSnifferProcess.php
@@ -64,6 +64,7 @@ class PhpCodeSnifferProcess
             $textDocument,
             [
               '--report=diff',
+              '--no-colors',
               empty($sniffs) ? '' : sprintf('--sniffs=%s', implode(',', $sniffs))
             ]
         );
@@ -101,7 +102,7 @@ class PhpCodeSnifferProcess
                 && $exitCode !== self::EXIT_FOUND_NON_FIXABLE_ERRORS
                 && $exitCode !== self::EXIT_FILES_NEEDS_FIXING
             ) {
-                throw new RuntimeException(
+                $this->logger->error(
                     sprintf(
                         "phpcs exited with code '%s'; cmd: %s; stderr: '%s'; stdout: '%s'",
                         $exitCode,
@@ -110,6 +111,7 @@ class PhpCodeSnifferProcess
                         $stdout
                     )
                 );
+                return '[]';
             }
 
             return $stdout;
```

dantleech:
It seems that applying the fix on the document multiple times without saving the document can cause some unexpected edits, maybe we need to add `--no-cache` too (?)

composer.json
bart-jaskulski: Actually, we could inspire by `composer/composer` test package and create some fixtures, providing *closed* execution environment, based on their [`initTempComposer`](https://github.com/composer/composer/blob/main/tests/Composer/Test/TestCase.php#L127).

Creating a new temporary *project* for test case would spare us from keeping unnecessary dev dependencies fixed to a specific version (I guess declaring `"squizlabs/php_codesniffer": "<4"` wouldn't help either). But on the other side, such tests would be slow as it would require to install dependency during runtime.

dantleech: yeah, like I say it's fine for now, but at some point I might clean up and stub out the processes and/or execute them in separate pipelines using a similar approach to what you suggest.

bart-jaskulski:
This gets a bit more complex, taking under consideration your findings. Applying `--no-cache` option on fixer does the job, and edits apply cleanly, but there are some issues about how `phpcs` is generating diff. It simply uses a systemwide `diff` executable which compares the currently saved file. Diff just reads the file from path, and you cannot pass an edited, unsaved document.

All revolves around this line in `phpcs`: https://github.com/squizlabs/PHP_CodeSniffer/blob/master/src/Fixer.php#L252

I see the following options for us at the moment:

1. Allow to use diagnostics only with `language_server.diagnostics_on_update` option disabled, so we ensure document is always readable for diff. This sucks, but is the least complex solution.
2. Implement own reporter based on PHP_Codesniffer, which would *improve* the report in our use case. This seems worse as we need to keep up with `phpcs`
3. Improve diff parsing, to make it aware of changes proposed by tool against changes introduced by user. This may be impossible, especially when the user edits the same chunk of code.

dantleech:
I think we could also write to a temporary file and use phpcs on that?

bart-jaskulski:
Well, I consider writing to a temporary file an overengineered solution, as we always need to write a preliminary file to provide diagnostics. This might be an overhead, especially for large files and editors opening multiple buffers simultaneously.

After a brief search, I found one tool, which adheres to LSP design and chooses to display phpcs diagnostic as one line range (not specifically one character, but still simplified). https://github.com/ikappas/vscode-phpcs/blob/develop/phpcs-server/src/linter.ts#L241

Again, if that solution doesn't bring too many flaws into the overall design, I would strongly recommend it as it seems the most straightforward to implement.

dantleech:
Well, the overhead is negligible/imperceptible, esp. considering the overhead of running phpcs. It would also be reliable if I understand correctly. Alternatively we can oursevles diff before/after. There is something for that somewhere...

bart-jaskulski:
You convinced me :smile: I've taken a peek at phpstan extension implementation and accordingly adapted solution for phpcs.

lib/Extension/PhpCodeSniffer/Model/PhpCodeSnifferProcess.php
```diff
@@ -124,4 +145,14 @@ private function runDiagnosticts(TextDocumentItem $textDocument, array $options
             return $stdout;
         });
     }
+
+    private function createTempFile(string $text): string
+    {
+        $tmpName = tempnam(sys_get_temp_dir(), 'phpcsls');
+        $name = sprintf('%s.php', $tmpName);
+        rename($tmpName, $name);
+        file_put_contents($name, $text);
```
dantleech, line 158: probably need some error handling here, I think any one of these operations can fail

lib/Extension/PhpCodeSniffer/Model/PhpCodeSnifferProcess.php
```diff
@@ -146,6 +146,10 @@ private function runDiagnosticts(string $url, string $text, array $options = [])
         });
     }
 
+    /**
+     * @return string Filename MUST include PHP extension, otherwise
+     * phpcs will not process it.
```
dantleech, line 151: Move the "comment" to the line above so it's not part of `@return` (alternatively could be `@param` as intended? but I'd just make it a general comment on the mehtod)

dantleech:
seems to work great, thanks @bart-jaskulski !

bart-jaskulski:
Thank you for the opportunity to contribute! And I apologize for my wording, if any of my previous comments caused unhealthy discord :innocent:

PR #2369, opened 2023-09-11
Change clone URL to HTTPS instead of SSH

warigan:
Attempting to clone with SSH will fail due to permission deined, for non-collaborators.

### DIFF
--- a/doc/usage/standalone.rst
+++ b/doc/usage/standalone.rst
@@ -36,7 +36,7 @@ You can checkout the project and then create a symlink.
 .. code:: bash
 
    $ cd ~/home/you/somewhere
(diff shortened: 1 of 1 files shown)


bart-jaskulski:
Please, make sure you have your SSH keys set up correctly in GitHub profile[^1][^2]. I don't think it is possible to block SSH access for a specific repository (I haven't had any troubles with cloning it, even before I've submitted any PRs).

[^1]: https://docs.github.com/en/authentication/connecting-to-github-with-ssh
[^2]: https://docs.github.com/en/authentication

dantleech:
It may be that people don't have a github account at all, so this makes sense I think 👍

PR #2371, opened 2023-09-16
Gh 2366: Fix undefined var behavior for `@var` declarations

### LINKED ISSUE #2366: @var annotation behaving erratically
I have a PHP template file, which is included from my template engine class. Hence the variables are not defined within the template file, and I am trying to define the variables using @var annotation. But phpactor is behaving erratically when @var annotation is used to specify the type of a variable.

```
<?php

/** @var string $a */
/** @var string $b */
/** @var string $c */
?>
<?php

echo $a + $b;

echo $a;
echo $b;

```

image

But if I change the `+` operator to `.`, it no longer works.

image

dantleech:
Fixes #2366

### DIFF
--- a/lib/WorseReflection/Core/Inference/Resolver/BinaryExpressionResolver.php
+++ b/lib/WorseReflection/Core/Inference/Resolver/BinaryExpressionResolver.php
@@ -297,7 +297,6 @@ class BinaryExpressionResolver implements Resolver
 
         if (!in_array($operator, [
             TokenKind::DotEqualsToken,
-            TokenKind::DotToken,
         ])) {
             return;
         }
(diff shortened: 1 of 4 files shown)


lib/WorseReflection/Core/Inference/Resolver/BinaryExpressionResolver.php
```diff
@@ -297,7 +297,6 @@ private function addVariable(
 
         if (!in_array($operator, [
             TokenKind::DotEqualsToken,
-            TokenKind::DotToken,
```
dantleech, line 300: there may well have been a reason for this, but it's lost in time and the tests don't care so :shrug:

PR #2384, opened 2023-10-02
phpcs-fixer custom config

zordsdavini:
Added possibility to use custom config for php-cs-fixer

### DIFF
--- a/lib/Extension/LanguageServerPhpCsFixer/LanguageServerPhpCsFixerExtension.php
+++ b/lib/Extension/LanguageServerPhpCsFixer/LanguageServerPhpCsFixerExtension.php
@@ -22,6 +22,7 @@ class LanguageServerPhpCsFixerExtension implements OptionalExtension
     public const PARAM_PHP_CS_FIXER_BIN = 'language_server_php_cs_fixer.bin';
     public const PARAM_ENV = 'language_server_php_cs_fixer.env';
     public const PARAM_SHOW_DIAGNOSTICS = 'language_server_php_cs_fixer.show_diagnostics';
+    public const PARAM_CONFIG = 'language_server_php_cs_fixer.config';
     public const PARAM_ENABLED = 'language_server_php_cs_fixer.enabled';
 
     public function load(ContainerBuilder $container): void
@@ -31,10 +32,16 @@ class LanguageServerPhpCsFixerExtension implements OptionalExtension
             function (Container $container) {
                 $path = $container->get(FilePathResolverExtension::SERVICE_FILE_PATH_RESOLVER)->resolve($container->parameter(self::PARAM_PHP_CS_FIXER_BIN)->string());
 
+                $configPath = null;
+                if ($container->parameter(self::PARAM_CONFIG)->value()) {
+                    $configPath = $container->get(FilePathResolverExtension::SERVICE_FILE_PATH_RESOLVER)->resolve($container->parameter(self::PARAM_CONFIG)->string());
+                }
+
                 return new PhpCsFixerProcess(
                     $path,
                     LoggingExtension::channelLogger($container, 'php-cs-fixer'),
                     $container->parameter(self::PARAM_ENV)->value(),
+                    $configPath
                 );
             }
         );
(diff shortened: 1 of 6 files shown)


lib/Extension/LanguageServerPhpCsFixer/LanguageServerPhpCsFixerExtension.php
```diff
@@ -31,10 +32,16 @@ public function load(ContainerBuilder $container): void
             function (Container $container) {
                 $path = $container->get(FilePathResolverExtension::SERVICE_FILE_PATH_RESOLVER)->resolve($container->parameter(self::PARAM_PHP_CS_FIXER_BIN)->string());
```
dantleech, line 33: you can use `$container->expect(FilePathResolverExtension::SERVICE_FILE_PATH_RESOLVER, PathResolver::class)` to avoid the phpstan error
`

lib/Extension/LanguageServerPhpCsFixer/LanguageServerPhpCsFixerExtension.php
```diff
@@ -31,10 +32,16 @@ public function load(ContainerBuilder $container): void
...
 
+                $configPath = null;
+                if ($container->parameter(self::PARAM_CONFIG)->value()) {
+                    $configPath = $container->get(FilePathResolverExtension::SERVICE_FILE_PATH_RESOLVER)->resolve($container->parameter(self::PARAM_CONFIG)->string());
```
dantleech, line 37: same - infact you coud extract the path resolver to a variable

lib/Extension/LanguageServerPhpCsFixer/LspCommand/FormatCommand.php
```diff
@@ -35,7 +35,14 @@ public function __invoke(string $uri, ?array $rules = null): Promise
             $textDocument = $this->workspace->get($uri);
 
             $rulesOpt = $rules ? ['--rules', ...$rules] : [];
-            $diff = yield $this->phpCsFixer->fix($textDocument->text, ['--diff', '--dry-run', ...$rulesOpt]);
+
+            $configOpt = [];
+            if ($this->phpCsFixer->hasConfigPath()) {
+                $rulesOpt = [];
+                $configOpt = ['--config', $this->phpCsFixer->getConfigPath()];
```
dantleech, line 42: this should be in the `PhpCsFixerProcess` class. The "config path" should not escape from the `PhpCsFixerProcess`

dantleech:
nice :tada: you need to move merge the config path in the `PhpCsFixerProcess` class though, not pass it from the LspCommand.

lib/Extension/LanguageServerPhpCsFixer/LspCommand/FormatCommand.php
zordsdavini: we can't or we need to add option ignore config as in diagnostics there are processed per rule. Config will overwrite. What do you think?

dantleech: confused, shouldn't the config be the same regardless of where it's invoked?

zordsdavini: the flow of diagnostics is like that in short:
- get all errors/warnings in json
- get rules from json
- go through all rules and format them (so, no config with all rules)

Maybe I am wrong or the flow is wrong but for me it works. Maybe we can get descriptions in other way, then we could ignore this run per rule and give error/warning in different way

zordsdavini:
p.s. CI/PHPFixer updates code? Because I don't add `use \RuntimeException;` or formatting `__construct`

lib/Extension/LanguageServerPhpCsFixer/LspCommand/FormatCommand.php
dantleech: I still don't see why we'd want to have "no config" in one situation and not in another. It's using the `.php_cs_.dist.php` file (or whatever) by default anyway - so this option is changing the location it's looking for that file basically?

zordsdavini: default file is `.php-cs-fixer.dist.php` in root directory, but good practice is to use `.php-cs-fixer.php` (added to `.gitignore`). So, yes, this `--config` is to locate configuration. 

Before I said that there is overwrittings. I was wrong:
 
```
[PhpCsFixer\ConfigurationException\InvalidConfigurationException (16)]
  Passing both `--config` and `--rules` options is not allowed.
```

zordsdavini: There is `FormatCommand` what takes `$rules` as argument. I don't get when it is used. If it doesn't take argument then everything can be simplified: on `fix()` check does options contains rules and if not - try to add config

zordsdavini: OK. So, tried to move into `PhpCsFixerProcess`. For my workflow: show diagnostics on the fly and fix on demand - it works. Config path is used and it listens to its rules.

dantleech:
Thanks!

PR #2406, opened 2023-10-15
Removing webmozart-path plugin

### LINKED ISSUE #1616: Package webmozart/path-util is abandoned, you should avoid using it. Use symfony/filesystem instead.
I got a warning when I run `composer install` on phpactor directory.
```
Package webmozart/path-util is abandoned, you should avoid using it. Use symfony/filesystem instead.
```
I think we should change `webmozart/path-util` package with `symfony/filesystem`.
How about that?

mamazu:
We still have a dependency in the lockfile because of psalm but we don't used it anymore. 

Updated psalm from 4.x to 5.x. This shouldn't be a bc break as this is a dev package.

Closes #1616

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -45,7 +45,7 @@
         "phpstan/phpstan-phpunit": "^1.0",
         "jangregor/phpstan-prophecy": "^1.0",
         "phpstan/extension-installer": "^1.1",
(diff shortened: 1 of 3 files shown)


mamazu:
Looks like the the diagnostics have changed. Updating the tests

PR #2411, opened 2023-10-21
fix: LSP Position is 0 based where LineCol/LineColRange are 1 based.

zobo:
In PhpLint the reported Diagnostic Range is incorrect as it is 1 based instead of 0.

A good alternative would be to provide mapping classes. `PositionConverter` and `RangeConverter` seem like good candidates to add a `fromLineCol` and `fromLineColRange` methods.


dantleech:
didn't notice that, thanks!

Issue #2409, opened 2023-10-20
bug: Cannot generate method on enum


dantleech:
Although it would be possible to generate a method on an enum, there is a guard condition which prevents that happening if the class is not an interface or "class".

mamazu:
```text
1) Phpactor\CodeTransform\Tests\Adapter\WorseReflection\Refactor\WorseGenerateMethodTest::testGenerateMethod with data set "public method on enum" ('generateMethod_enum.test', 'play')
Phpactor\WorseReflection\Core\Exception\MethodCallNotFound: Did not know how to reflect node of type "Microsoft\PhpParser\Node\EnumMembers"

/home/mamazu/packages/phpactor/phpactor/lib/WorseReflection/Bridge/TolerantParser/Reflector/TolerantSourceCodeReflector.php:98
/home/mamazu/packages/phpactor/phpactor/lib/WorseReflection/Core/Reflector/CompositeReflector.php:74
/home/mamazu/packages/phpactor/phpactor/lib/CodeTransform/Adapter/WorseReflection/Refactor/WorseGenerateMethod.php:40
/home/mamazu/packages/phpactor/phpactor/lib/CodeTransform/Tests/Adapter/WorseReflection/Refactor/WorseGenerateMethodTest.php:94
/home/mamazu/packages/phpactor/phpactor/lib/CodeTransform/Tests/Adapter/WorseReflection/Refactor/WorseGenerateMethodTest.php:22

Caused by
Phpactor\WorseReflection\Core\Exception\CouldNotResolveNode: Did not know how to reflect node of type "Microsoft\PhpParser\Node\EnumMembers"

/home/mamazu/packages/phpactor/phpactor/lib/WorseReflection/Core/Inference/NodeReflector.php:44
/home/mamazu/packages/phpactor/phpactor/lib/WorseReflection/Bridge/TolerantParser/Reflector/TolerantSourceCodeReflector.php:140
/home/mamazu/packages/phpactor/phpactor/lib/WorseReflection/Bridge/TolerantParser/Reflector/TolerantSourceCodeReflector.php:96
/home/mamazu/packages/phpactor/phpactor/lib/WorseReflection/Core/Reflector/CompositeReflector.php:74
/home/mamazu/packages/phpactor/phpactor/lib/CodeTransform/Adapter/WorseReflection/Refactor/WorseGenerateMethod.php:40
[...]

PR #2416, opened 2023-10-31
Generate methods on enum

mamazu:
Fixes: #2409

Adds the ability to generate methods on enum classes if they don't exist. In the process of implementing the feature I found out that phpactor doesn't have a code builder for enum classes which I added. However, the code builder only supports non-backed enums and might break enums that are backed. I'm not sure this is a requirement though or if I should just throw an error if the enum is backed and don't do anything.


dantleech:
nice :slightly_smiling_face:  can you add a test to see if it breaks backed enums?

mamazu:
Added a test of breaking enums. It breaks different than I expected it to break. And as for the 8.0 phpunit. I don't know why they work differently than the other ones. Does it have anything to do with the enums not being fully supported yet in that version?

dantleech:
probably, you need to skip enum support for 8.0

dantleech:
nice!

PR #2422, opened 2023-11-07
Gh 2399 enum cases and constant generation

dantleech:


flake.nix
```diff
@@ -30,7 +30,7 @@
             xdebug
           ]));
           extraConfig = ''
-            xdebug.mode=debug
+            xdebug.mode=none
```
dantleech, line 33: rever

PR #2428, opened 2023-11-12
[Docs][vim-lsp] code command snippets were broken

Numkil:
Hiya,

This is a fix for one of the documentation pages on the website. On the vim-lsp page it shows a few lua scripts to help in using phpactor with neovim. 

However showWindow would throw an error whenever res['result'] is not a proper string causing the function to stop and no information to be displayed. I fixed this by adding some protected call functions so that we make sure each result is added to the popup window.


dantleech:
Thanks!

Issue #2407, opened 2023-10-16
Request for additional  phpstan integration


Goku-San:
Hello,

This is a request for additional phpstan integration if its possible...
I use Vim and have coc-nvim installed and I have set up language server in `coc-settings.json` 
```json
  "languageserver": {
    "phpactor": {
      "command": "~/.vim/autoload/plugged/phpactor/bin/phpactor",
      "args": ["language-server"],
      "trace.server": "verbose",
      "filetypes": ["php"]
    }
  }
```
In `~/.config/phpactor/phpactor.yml` I have set these options according to [language-server-phpactor-bin](https://phpactor.readthedocs.io/en/master/reference/configuration.html#language-server-phpactor-bin)
```yaml
language_server_phpstan.enabled: true
language_server_phpstan.bin: ~/.config/composer/vendor/bin/phpstan
language_server.diagnostics_on_update: false
language_server_completion.trim_leading_dollar: true
```
Could you please add config options if possible, for 
```yaml
language_server_phpstan.memoryLimit: 'to set memory limit ex. 1024M or 1G.. or what ever ram memory desired'
language_server_phpstan.configFile: 'for config file ex. ~/.config/composer/phpstan.neon'
```
With these two options I think the phpstan integration will be complete, especially with the second option because I can use global ``phpstan.neon`` file that will work say for all Laravel projects in my case.

Thanks...

ungrim97:
I have given it a go (mostly because having the --configuration option working would save me running phpstan via null-ls in nvim).

Will confess that PHP isn't a primary lang for me, so no doubt it could be improved. I also had some issues running the `compose run intgrate` script, so the docs haven't been regenerated (I can run make docs if required). 

Let me know if you want me to make any changes.

dantleech:
looks like we fixed it?

PR #2430, opened 2023-11-13
Support setting --configuration setting of Phpstan

ungrim97:
Attempt to add --configuration and --memory-limit PHPstan cli option support as requested in ticket #2407

I haven't updated more than the code yet, and would request input on if the docs are auto generated in some way. I attemped to run `composer run integrate` but it failed with some memory issues that appeared unrelated

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -4112,11 +4112,16 @@ parameters:
 
 		-
 			message: "#^Cannot call method resolve\\(\\) on mixed\\.$#"
-			count: 2
+			count: 3
 			path: lib/Extension/LanguageServerPhpstan/LanguageServerPhpstanExtension.php
 
 		-
 			message: "#^Parameter \\#2 \\$level of class Phpactor\\\\Extension\\\\LanguageServerPhpstan\\\\Model\\\\PhpstanConfig constructor expects string\\|null, mixed given\\.$#"
+			count: 1
+			path: lib/Extension/LanguageServerPhpstan/LanguageServerPhpstanExtension.php
+
+		-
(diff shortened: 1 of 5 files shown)


phpstan-baseline.neon
```diff
@@ -4112,14 +4112,19 @@ parameters:
...
 		-
 			message: "#^Parameter \\#2 \\$level of class Phpactor\\\\Extension\\\\LanguageServerPhpstan\\\\Model\\\\PhpstanConfig constructor expects string\\|null, mixed given\\.$#"
 			count: 1
 			path: lib/Extension/LanguageServerPhpstan/LanguageServerPhpstanExtension.php
 
+		-
+			message: "#^Parameter \\#4 \\$memLimit of class Phpactor\\\\Extension\\\\LanguageServerPhpstan\\\\Model\\\\PhpstanConfig constructor expects string\\|null, mixed given\\.$#"
+			count: 1
+			path: lib/Extension/LanguageServerPhpstan/LanguageServerPhpstanExtension.php
+
```
dantleech, line 4127: Can we fix these issues?

phpstan-baseline.neon
```diff
@@ -4112,14 +4112,19 @@ parameters:
 
 		-
 			message: "#^Cannot call method resolve\\(\\) on mixed\\.$#"
-			count: 2
+			count: 3
```
dantleech, line 4115: Let's fix this rather than add another caee

dantleech:
Looks good, let's fix the phpstan issues rather than adding to the baseline though. As for the docs, I can generate then after merging.

ungrim97:
@dantleech Have resolved the PHPstan issue rather than adding them to the ignore list. Php-cs-fixer seems to have prefered a change in the code layout as well, hopefully thats correct.

* Removed Deprecated `$container->getParameter` call for `$container->parameter()->string()` call
* Checked for presence of parameter and use ->string() rather than ->value() to avoid mixed type usage

dantleech (approved):

dantleech:
thanks! will update the changelog and docs now

PR #2435, opened 2023-11-18
fix: does not comply with psr-4 autoloading standard

gerardroche:
Fixes PSR-4 autoloading issues as seen when installing via:

```
composer install --no-dev -o
```

These are the warnings that this PR fixes:

```
Class Phpactor\ClassMover\Tests\Adapter\TolerantParser\ExampleTrait located in ./lib/ClassMover/Tests/Adapter/TolerantParser/examples/Example6.php does not comply with psr-4 autoloading standard. Skipping.
Class Phpactor\ClassMover\Tests\Adapter\TolerantParser\Example7 located in ./lib/ClassMover/Tests/Adapter/TolerantParser/examples/Example7.php does not comply with psr-4 autoloading standard. Skipping.
Class Phpactor\ClassMover\Tests\Adapter\TolerantParser\Example5Interface located in ./lib/ClassMover/Tests/Adapter/TolerantParser/examples/Example5.php does not comply with psr-4 autoloading standard. Skipping.
Class Phpactor\Extension\LanguageServerPsalm\Model\Excepteion\PsalmProcessError located in ./lib/Extension/LanguageServerPsalm/Model/Excepteion/PhpstanProcessError.php does not comply with psr-4 autoloading standard. Skipping.
Class Phpactor\CodeTransform\Tests\Adapter\WorseReflection\Refactor\WorseGenerateMethodTest located in ./lib/CodeTransform/Tests/Adapter/WorseReflection/Refactor/WorseGenerateMemberTest.php does not comply with psr-4 autoloading standard. Skipping.
```

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,12 @@
+# Handle line endings automatically for files detected as text and leave all
+# files detected as binary untouched.
+* text=auto
+
+# Files and directories with the attribute export-ignore won’t be added to
+# archive files. See http://git-scm.com/docs/gitattributes for details.
+.gitattributes          export-ignore
+.gitignore              export-ignore
+/.github                export-ignore
+/.php-cs-fixer.dist.php export-ignore
+/phpunit.xml.dist       export-ignore
+/tests/                 export-ignore
--- a/lib/ClassMover/Tests/Adapter/TolerantParser/TolerantClassReplacerTest.php
+++ b/lib/ClassMover/Tests/Adapter/TolerantParser/TolerantClassReplacerTest.php
@@ -120,19 +120,19 @@ class TolerantClassReplacerTest extends TestCase
             ],
             'Change namespace of interface' => [
                 'Example5.php',
-                'Phpactor\ClassMover\Tests\Adapter\TolerantParser\Example5Interface',
-                'Phpactor\ClassMover\Tests\Adapter\TolerantParser\BarBar\FoobarInterface',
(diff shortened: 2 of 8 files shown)


dantleech:
Thanks - I think for the test files it would be better to add a `.gitattributes` file:

```
*.php diff=php
/path1/to/tests/to/ignore export-ignore
```

that will prevent them from being exported. but otherwise if we can't get composer to ignore them easily happy to make the change you suggest here. 

the "PslamProcessError" file is not used and can be deleted AFAICS

gerardroche:
I removed two unused exceptions `PslamProcessError` and `PhanProcessError`. There is no reference to them anywhere, and there was a typo in the namespace "Excepteion".

I added gitattributes export rules and baseline line endings rule. I wasn't sure what files can be ignored,  so I just added some obvious ones. You will know better than me which files should be ignored.

The export-ignore rules are good, but they don't help for my use case because I install phpactor directly from a clone. 

I noticed that there are tests nested inside `lib/*/Tests/`. All of those files get populated into the autoload classmap  generated by composer. The `tests/` files don't when you do `composer --no-dev` because those are declared in the composer `autoload-dev`. 

I think composer has ways to handle those `lib/*/Tests/` files, but I don't know off the top of my head. I don't think it's a major deal. It might save some memory, but performance-wise it should be negligible.

dantleech (approved):

dantleech:
Great, thanks!

PR #2447, opened 2023-12-01
Initialize instances with parentheses

lizhineng:
Added style fixer [new_with_parentheses](https://mlocati.github.io/php-cs-fixer-configurator/#version:3.40|fixer:new_with_parentheses) and bumped the minimum version of PHP CS Fixer to **v3.32** which deprecated the `new_with_braces` fixer and introduced its successor `new_with_parentheses`.


dantleech:
Thanks!

PR #2456, opened 2023-12-03
Gh 2419 completion param context

### LINKED ISSUE #2419: completion: parameter suggestions should take into account type of parameter
e.g. 

```
function foo(LoggerInterface $logger) {}
```

only instances of `LoggerInterface` should be suggested for `foo(<>` or `foo(new <>` (in addition to variables etc)

dantleech:
This PR will narrow completion suggestions to the parameter being completed.

Fixes #2419 

Note this is a relatively risky change in a butterfly-flaps-its-wings-and-everything-is-broken kind of way - so I'll merge it after releasing.

Before:

image

After:

image


lib/Completion/Tests/Integration/Bridge/TolerantParser/ReferenceFinder/ExpressionNameCompletorTest.php
```diff
@@ -235,6 +244,9 @@ protected function createTolerantCompletor(TextDocument $source): TolerantComple
         $searcher->search('b', null)->willYield([
             NameSearchResult::create('class', 'Foo\\Bar'),
         ]);
+        $searcher->search('', null)->willYield([
+            NameSearchResult::create('class', 'EmptyString'),
+        ]);
         $searcher->search('\NS1\Foo\Fo', null)->willYield([
```
dantleech, line 250: this test is terrible.

PR #2468, opened 2023-12-21
Adding global cli variables

mamazu:
If you are using php in console mode you also have access to the $argv and $argc variables and they are not undefined.

### DIFF
--- a/lib/WorseReflection/Bridge/TolerantParser/Diagnostics/UndefinedVariableProvider.php
+++ b/lib/WorseReflection/Bridge/TolerantParser/Diagnostics/UndefinedVariableProvider.php
@@ -308,6 +308,22 @@ class UndefinedVariableProvider implements DiagnosticProvider
             }
         );
         yield new DiagnosticExample(
+            title: 'SAPI global variables',
+            source: <<<'PHP'
+                    <?php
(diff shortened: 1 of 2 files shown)


lib/WorseReflection/Bridge/TolerantParser/Diagnostics/UndefinedVariableProvider.php
```diff
@@ -307,6 +307,22 @@ function preg_match(string $pattern, string $string, array &$matches): bool {}
...
+                if ($argc === 2) {
+                    echo "Hello ".$argv[1].PHP_EOL;
+                } else {
+                    echo "Usage ".__FILE__. " <name>";
+                }
+            PHP,
+            valid: true,
+            assertion: function (Diagnostics $diagnostics): void {
+                Assert::assertCount(0, $diagnostics);
+            }
```
mamazu, line 323: This assertion is kinda superfluous as it is also checked when `valid: true` is set. Maybe we could make the assertion optional when `valid` is true?

dantleech: yeah possibly

dantleech:
thanks!

PR #2469, opened 2023-12-21
Fixing this in anonoymous class

### LINKED ISSUE #2467: `$this` not recognized inside anonymous class
Hello!

Please see the following:

```php
<?php

return new class
{
    public function foo(): void
    {
        $this->bar();
    }

    public function bar(): bool
    {
        return false;
    }
};
```

image

Thanks!

mamazu:
Fixes #2467

### DIFF
--- a/lib/WorseReflection/Core/Inference/Walker/FunctionLikeWalker.php
+++ b/lib/WorseReflection/Core/Inference/Walker/FunctionLikeWalker.php
@@ -5,6 +5,7 @@ namespace Phpactor\WorseReflection\Core\Inference\Walker;
 use Microsoft\PhpParser\MissingToken;
 use Microsoft\PhpParser\Node;
 use Microsoft\PhpParser\Node\Expression\ArrowFunctionCreationExpression;
+use Microsoft\PhpParser\Node\Expression\ObjectCreationExpression;
 use Microsoft\PhpParser\Node\Statement\EnumDeclaration;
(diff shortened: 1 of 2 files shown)


lib/WorseReflection/Core/Inference/Walker/FunctionLikeWalker.php
```diff
@@ -71,7 +72,8 @@ private function walkFunctionLike(FrameResolver $resolver, Frame $frame, Functio
             ClassDeclaration::class,
             InterfaceDeclaration::class,
             TraitDeclaration::class,
-            EnumDeclaration::class
+            EnumDeclaration::class,
+            ObjectCreationExpression::class, // For Inline classes
```
dantleech, line 76: not sure this will work as object creation expressions can be inside class declarations etc.

mamazu: I have looked at the node tree. And where a normal definition has a ClassDefinitionNode there is an ObjectCreactionExpression

dantleech: ah i see, i'm confusing ancestor and descendant :facepalm:

dantleech:
the next bit would probably be adding a new `AnonymousClassType` implementing `ClassLikeType`  and adding the class members to it, probably quite some work to do there.

PR #2470, opened 2023-12-21
Removing removing broken debug function

mamazu:
Reasons:
* It's nowhere used (neither the file itself nor the function)
* It references a class that doesn't exist
* It produces phpstan errors because of the non-existing class

But I totally agree that it would be nice to have a Node debugging Tool.


dantleech:
thanks, it's an artifact from when it lived in a separate package

PR #2476, opened 2023-12-28
Indexer: fix `#[\Attribute]` resolution

przepompownia:
It's not perfect because 

```php
        $this->workspace()->put('project/Bax.php', '<?php use Not\Attribute; #[Attribute] class Bax {}');
```
still can be considered attribute.

### DIFF
--- a/lib/Indexer/Adapter/Tolerant/Indexer/ClassDeclarationIndexer.php
+++ b/lib/Indexer/Adapter/Tolerant/Indexer/ClassDeclarationIndexer.php
@@ -54,7 +54,7 @@ class ClassDeclarationIndexer extends AbstractClassLikeIndexer
                     continue;
                 }
                 /** @phpstan-ignore-next-line */
-                if (ltrim($attribute->name?->getText(), '\\') === 'Attribute') {
+                if ((string) $attribute->name?->getResolvedName() === \Attribute::class) {
                     $record->addFlag(ClassRecord::FLAG_ATTRIBUTE);
                     return;
                 }
--- a/lib/Indexer/Adapter/Tolerant/Indexer/ClassDeclarationIndexer.php
+++ b/lib/Indexer/Adapter/Tolerant/Indexer/ClassDeclarationIndexer.php
@@ -54,7 +54,7 @@ class ClassDeclarationIndexer extends AbstractClassLikeIndexer
                     continue;
                 }
                 /** @phpstan-ignore-next-line */
-                if ($attribute->name?->getText() === 'Attribute') {
+                if (ltrim($attribute->name?->getText(), '\\') === 'Attribute') {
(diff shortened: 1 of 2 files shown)


mamazu:
In consideration that we might want to do more with attributes in the future we might as well fix it properly. I have seen that for implementations of the class there is a `getResolvedName` function.
https://github.com/phpactor/phpactor/blob/efc74ee9043e59dd8d1ded5cb6186308fb0e6efc/lib/Indexer/Adapter/Tolerant/Indexer/ClassDeclarationIndexer.php#L96C29-L96C44

Maybe we could use this for attributes and also fix the issue described above with the namespace not being taken into consideration.

przepompownia:
@mamazu thanks for suggestion.

lib/Indexer/Adapter/Tolerant/Indexer/ClassDeclarationIndexer.php
```diff
@@ -54,7 +54,7 @@ public function indexAttributes(ClassRecord $record, ClassDeclaration $node): vo
                     continue;
                 }
                 /** @phpstan-ignore-next-line */
-                if ($attribute->name?->getText() === 'Attribute') {
+                if ((string) $attribute->name?->getResolvedName() === \Attribute::class) {
```
dantleech, line 57: [getResolvedName](https://github.com/microsoft/tolerant-php-parser/blob/main/src/Node/QualifiedName.php#L82) is however potentially enormously inefficient. For this case it's probably enough to just check both variations?

przepompownia: Btw. I did not checked how much the indexing time increases after this change.

dantleech: would be relative to the numnber of attributes in the source tree i guess

mamazu: True, it could be expensive. I haven't thought about it too much and just having a "good enough" solution would probably be sufficient.

przepompownia: Independently of expensiveness I added such example to the test.

dantleech:
Performance is probably "fine" we already do this for all class references, so a fee more shouldn't hurt.

przepompownia:
Now Symfony `Route` attribute became available in completion results. Thanks for merging.

dantleech:
Oh we might want to bump the indexer version to force reindexing when people upgrade, but no biggie

mamazu:
> Now Symfony `Route` attribute became available in completion results. Thanks for merging.

Oh that's why it wasn't showing on my project. Was wondering why. :D Thanks for fixing that (It probably also affects more symfony attributes) like `AsContainer`.

PR #2478, opened 2023-12-29
Replace `strpos()` and `mb_strpos()` with `str_starts_with()`

zonuexe:


lib/Completion/Bridge/TolerantParser/WorseReflection/WorseClassMemberCompletor.php
```diff
@@ -81,7 +85,7 @@ public function complete(Node $node, TextDocument $source, ByteOffset $offset):
 
         foreach ($type->expandTypes()->classLike() as $type) {
             foreach ($this->populateSuggestions($nodeContext, $type, $static, $shouldCompleteOnlyName, $isInstance) as $suggestion) {
-                if ($partialMatch && 0 !== mb_strpos($suggestion->name(), $partialMatch)) {
+                if ($partialMatch && !str_starts_with($suggestion->name(), $partialMatch)) {
```
zonuexe, line 88: Generally, don't need the `mb_` version of the function for this purpose. Look at <https://wiki.php.net/rfc/str_contains#case-insensitivity_and_multibyte_strings>.

lib/Indexer/Model/Query/Criteria/FileAbsolutePathBeginsWith.php
```diff
@@ -18,7 +19,6 @@ public function isSatisfiedBy(Record $record): bool
             return false;
         }
 
-        $begins = strpos($record->filePath(), $this->prefix);
-        return $begins !== false && 0 === $begins;
+        return str_starts_with($record->filePath() ?? '', $this->prefix);
```
zonuexe, line 22: Removing this `?? ''` will cause a runtime error.

dantleech:
looks good! is it just a tidy up or is it related to a bug?

zonuexe:
@dantleech Although I have not yet investigated the detailed conditions of occurrence, I found from the runtime error that there is a possibility of optimization here.

dantleech:
thanks :slightly_smiling_face:

mamazu:
This might have been an oversight on the Rector part and me, when migrating to php8. There might be more of those hidden in the code.

PR #2479, opened 2023-12-29
Replacing empty with the corresponding type checks

mamazu:
## Why

According to the documentation [empty](https://www.php.net/manual/en/function.empty.php) checks that the variable is present and not false. Which 
* Checks variables if they exist (we know they all exist)
* We possibly need to type cast values to bool
* This helps to make the code more readable (imo) because you can infer the type from the check.

So this might even be a little performance improvement (but very very negligable)


dantleech:
nice :ok_hand:

PR #2481, opened 2023-12-30
Rectoring the codebase to php 8.1

mamazu:
Applied rules:
* Closures to arrow functions
* strpos to str_starts_with
* isset to collocate
* Removing unused parameters from method calls
* Using 8.1 Initializers

This merge request is best reviewed commit by commit (every commit is a different rule)


dantleech:
good idea, but I'd not apply the short function refactoring, as it:

- makes some of the code harder to read (lines very long)
- chances are at some point you need to change them to be not short closures for debugging or whatever

mamazu:
Yeah for some functions it makes sense, for some functions it doesn't but I'll just pop the commits out.

dantleech:
great, thanks!

PR #2488, opened 2024-01-04
Use non-deprecated checkhealth syntax

dundargoc:


dantleech:
I think this is Neovim specific?

dantleech:
thanks!

PR #2494, opened 2024-01-07
Fix willRename behavior

afonsocarlos:
Rename only the original reference as described in [WillRenameFiles LSP specifications](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_willRenameFiles)


dantleech:
can you describe what the problem this fixes?

afonsocarlos:
Of course, my bad :sweat_smile:

If you rename a file from within a neovim plugin that supports willRename LSP integration this neovim plugin will request the LSP renaming BEFORE it actually renames the file. Therefore, the TextDocumentLocator can't find any reference to the new file (as it still doesn't exist) and throws an error "Text Document not found". 
According to the LSP specifications, we should send a response with the WorkspaceEdits which will be applied to workspace before the files are renamed, so it would make sense to find references for the class references before the rename.

By the way, I tested the rename feature with both [nvim-lsp-file-operations](https://github.com/antosha417/nvim-lsp-file-operations) and [oil.nvim](https://github.com/stevearc/oil.nvim) plugins and both broke with the same error from phpactor lsp server

afonsocarlos:
Have the issue and the solution become clear?

przepompownia:
I didn't see this PR till now and fixed it independently in #2531 along with adding missing response part.

Issue #2344, opened 2023-08-21
Is there a way to tell phpactor what "$this" is, in some files?


Einenlum:
Hi Dan,

Hope you're doing well.

I'm using [Pest](https://pestphp.com/) more and more and Pest tests look like this:

```php
<?php
// tests/Feature/BrandsTest.php

test('brands', function () {
    $response = $this->get('/brands');

    $response->assertStatus(200);
});
```

`$this` in this context is not known to phpactor (neither to intelephense), which makes sense. But under the hood it's `Tests\TestCase`, thanks to this line:

```php
// tests/Pest.php
uses(TestCase::class)->in('Feature');
```

For now I'm stuck with adding this docblock to each test:

```php
<?php

test('brands', function () {
    /** @var Tests\TestCase $this */
    $response = $this->get('/brands');

    $response->assertStatus(200);
});
```

I imagine it would be out of the scope of phpactor to parse the Pest file, but is there a way to tell phpactor (in a `.phpactor.yml` file for example) what `$this` refers to in a specific set of files (like in a whole folder)?

Thanks!

dantleech:
I would imagine it would be possible to add a `PestExtension` which could parse `Pest.php` if that's the convention.

CalebDW:
PHPStan also has an issue with `$this` inside a closure. I just use `test()` instead of `$this` (works for properties too):


```php
<?php
// tests/Feature/BrandsTest.php

test('brands', function () {
   test()->get('/brands')->assertStatus(200);

    test()->foo = 42;
});
```

Whax:
@Einenlum 
I saw you use neovim. What you can do for now is to disable phpactor for Test files only :
```lua
local str = {
  endsWith = function(s, ending)
    return ending == "" or s:sub(- #ending) == ending
  end
}
```
```lua
on_attach = function(client, bufnr)
  client.server_capabilities.completion = false
  client.server_capabilities.publishDiagnostics = true
  client.server_capabilities.diagnosticProvider = true

  local path = vim.fn.expand("%:p")
  if str.endsWith(path, "Test.php") then
    client.handlers["textDocument/publishDiagnostics"] = function(...)
      local result = select(2, ...)
      result.diagnostics = {}
    end
  end
end
```
It's a quick & dirty solution :3 For my case, it's fine (I use intelephense and phpactor).

@CalebDW 
This is normal, Pest use a lot of https://www.php.net/manual/en/closure.bind.php to redefine context of the closure.
[...]

PR #2502, opened 2024-01-12
Draft: Pest Extension

Einenlum:
Only a very early draft for now.

Not sure where I'm going with this but if you have some input/suggestions, it's very welcome :blue_heart: 

Tries solving #2344.

Goals:
- [x] Suggest Pest if it's installed
- [ ] Suggest completion if in a test file (for that we need to detect we're in a closure and suggest methods to `$this` as if it were bound to a TestCase class)
- [ ] For a first step, give completions from `PHPUnit\Framework\TestCase` if in `Unit` folder and from `Tests\TestCase` if in `Feature` folder. Not sure yet, but I think `Illuminate\\Foundation\\Testing\\RefreshDatabase` is only used by default if [pest laravel plugin](https://github.com/pestphp/pest-plugin-laravel) is installed (I need to check that).
- [ ] Then it would be great to analyze the [Pest.php](https://pestphp.com/docs/configuring-tests) to check which TestCase is used in which folder.

### DIFF
--- a/lib/Extension/Pest/Completion/PestCompletion.php
+++ b/lib/Extension/Pest/Completion/PestCompletion.php
@@ -1,0 +1,20 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Phpactor\Extension\Pest\Completion;
+
+use Generator;
+use Microsoft\PhpParser\Node;
+use Phpactor\Completion\Bridge\TolerantParser\TolerantCompletor;
+use Phpactor\Completion\Core\Suggestion;
+use Phpactor\TextDocument\ByteOffset;
+use Phpactor\TextDocument\TextDocument;
+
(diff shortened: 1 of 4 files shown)


dantleech:
> Suggest completion if in a test file

I think it's less about "completion" and more about telling the static analyser what `$this` is in a Closure.

This is currently done via. [frame walkers](https://github.com/phpactor/phpactor/blob/6f23af24bfd45a07861c705b8e551cae8b6fc328/lib/WorseReflection/Core/Inference/Walker/FunctionLikeWalker.php#L30) which are registered with a tag [here](https://github.com/phpactor/phpactor/blob/6f23af24bfd45a07861c705b8e551cae8b6fc328/lib/Extension/WorseReflection/WorseReflectionExtension.php#L118):

PR #2507, opened 2024-01-14
Suggesting named constructor params on attributes

### LINKED ISSUE #2495: Complete constructor params inside attribute usage
We have such completion available when calling the constructor directly
image
but not if we call it from an attribute node. 
image

mamazu:
Fixes #2495


mamazu:
Any ideas why the indexing would failed. I also have the same problem when running the tests on the master locally. Is there some reset logic not properly cleaning up?

dantleech:
thanks! it fails maybe 1 out of 10 times.

Issue #2199, opened 2023-04-10
Attribute completion: restrict result list by target, if present


przepompownia:
Based on  https://www.php.net/manual/en/language.attributes.classes.php we have 
```
    Attribute::TARGET_CLASS
    Attribute::TARGET_FUNCTION
    Attribute::TARGET_METHOD
    Attribute::TARGET_PROPERTY
    Attribute::TARGET_CLASS_CONSTANT
    Attribute::TARGET_PARAMETER
    Attribute::TARGET_ALL
```

przepompownia:
Maybe we can consider indexing `Attribute::IS_REPEATABLE` information too for diagnostics.

PR #2509, opened 2024-01-17
DocblockCompletor: add completion for `@throws`

przepompownia:


przepompownia:
At the moment I have no idea how to restrict results to exceptions.

przepompownia:
Another thing I noticed: after removing import of such class cannot restore it by any code action.

dantleech:
- we would need to filter the classes after retrieving the suggestions from the index somehow.
- probably we need to explicirtly add support to the missing classes diagnostic to support `@throws`
- otherwise I keep meaning to add a code action to generate `@throws` based on the method content.

przepompownia:
Thanks for adding it as is. This simple change very useful for me even in the current form, although I found the place for those changes accidentally, trying to find how to add snippet to `new Class<>` (found `ScfClassCompletor` only at the moment) :wink: 

I also don't have checked if we need to add some additional flags to index (I didn't forget about #2199 too) to search for exceptions only. 

image
It looks like working completion for union `@return` type so maybe it will be enough to find its implementation. 

A code action to generate `@throws` would be probably more convenient than this feature. 

Let's not forget about references and renaming classes used by `@throws`.

lib/Completion/Bridge/TolerantParser/WorseReflection/DocblockCompletor.php
```diff
@@ -41,6 +42,7 @@ class DocblockCompletor implements TolerantCompletor
         '@property',
         '@implements',
         '@extends',
+        '@throws',
```
przepompownia, line 45: This constant does not seem to have any usages.

przepompownia:
> 
>     * I guess union/intersection doesn't work on any docblock element.

Strange thing similar to #2513: now I get suggestion on the next part of union type:
image

PR #2513, opened 2024-01-20
Complete inside attribute parameter value

przepompownia:


przepompownia:
I suspect it's not the correct way to such completion - please close it if there is any better way. 

I have a look into `WorseParameterCompletor` but at the moment have no idea how to deal with attribute there.

przepompownia:
It does not restrict any result to constants or values.

dantleech:
:wave: can you add a description about what this does and why?

przepompownia:
I could not get such completion, but now I can on master:

image

Let me verify it later.

dantleech:
Maybe due to (mostly fixed) regressions with class filtering

PR #2514, opened 2024-01-21
KeywordCompletor: complete magic methods

przepompownia:


lib/Completion/Bridge/TolerantParser/WorseReflection/KeywordCompletor.php
```diff
@@ -14,6 +14,26 @@
...
+        '__isset' => "(string \\\$\${1:name}): bool\n{\$0\n}",
+        '__serialize' => "(): array\n{\$0\n}",
+        '__set' => "(string \\\$\${1:name}, mixed \\\$\${2:value}): void\n{\$0\n}",
+        '__set_state' => "(array \\\$\${1:properties}): object\n{\$0\n}",
+        '__sleep' => "(): array\n{\$0\n}",
+        '__toString' => "(): string\n{\$0\n}",
+        '__unserialize' => "(array \\\$\${1:data}): void\n{\$0\n}",
+        '__unset' => "(string \\\$\${1:name}): void\n{\$0\n}",
+        '__wakeup' => "(): void\n{\$0\n}",
+    ];
```
dantleech, line 35: hmm, I basically did _not_ include these because `__construct` is probably the 99% use case, would it make sense to prioritize that?

image

for example brings it to the top of the list?

lib/Completion/Bridge/TolerantParser/WorseReflection/KeywordCompletor.php
```diff
@@ -58,6 +76,19 @@ public function complete(Node $node, TextDocument $source, ByteOffset $offset):
...
+    /**
+     * @return Generator<Suggestion>
+     */
+    private function methods(): Generator
+    {
+        foreach (self::MAGIC_METHODS as $name => $snippet) {
+            yield Suggestion::createWithOptions($name . '(', [
+                'type' => Suggestion::TYPE_METHOD,
+                'priority' => 1,
+            ] + ['snippet' => $name . $snippet]);
```
dantleech, line 88: why don't we make `snippet` part of the array? (i.e. why use `+`?)

przepompownia: A habit from other places where it was conditional...

lib/Completion/Bridge/TolerantParser/WorseReflection/KeywordCompletor.php
przepompownia: I didn't look into the image, let me set special priority.

przepompownia:
Please look at the recent two commits. The change can be controversial. I test expectation I added sorting by priority first. Then it turned out that only two data sets require require a look - both had exceptions different than the real instance does.

dantleech:
LGTM, thanks!

PR #2515, opened 2024-01-21
Add snippet to type suggestions for object creation

przepompownia:


dantleech:
:bow:

przepompownia:
Thanks. I thought about adding exclusion if the next node is an existing parameter list (to avoid adding duplicated parens) but it can be done separately.

dantleech:
Just a heads up - I may revert this. It makes a mess when replacing an existing isntantiation

```
new Foobar();
```

I want to change it to:

```
new Barfo<>()
```

but it's then completed as:

```
new Barfoo()()
```

dantleech:
Edit I opened the PR above to revert the revert.

PR #2521, opened 2024-01-25
fix tests

gerardroche:
- fix(tests): ansi output causes test failures

  On some terminals, e.g. gnome-terminal, kitty (ubuntu), some tests fail
  because the output of process contain color codes. For example in
  the following assertion:

      $this->assertStringContainsString('class ⟶Badger⟵', $process->getOutput());

  The characters surrounding "Badger" will be in color and so the
  assertion fails because there are color codes in the string.

  This patch adds `--no-ansi` to the `-php-cs-fixer` test command so that
  the output never has color codes.

- fix(test): passing null to type string is deprecated

  When error reporting includes deprecations, the follow notice is emitted
  when running the tests:

  ```
  Deprecated: SplFileInfo::__construct(): Passing null to parameter #1 ($filename) of type string is deprecated in vendor/phpspec/prophecy/src/Prophecy/Doubler/Generator/ClassCreator.php(44) : eval()'d code on line 6
  ```

  The test has been updated to use Symfony's SplFileInfo class which
  Prophecy has an internal patch to help avoid the deprecations.

### DIFF
--- a/lib/CodeBuilder/Tests/Unit/Domain/TemplatePathResolver/FilterPhpVersionDirectoryIteratorTest.php
+++ b/lib/CodeBuilder/Tests/Unit/Domain/TemplatePathResolver/FilterPhpVersionDirectoryIteratorTest.php
@@ -6,11 +6,12 @@ use ArrayIterator;
 use Iterator;
 use PHPUnit\Framework\TestCase;
 use Phpactor\CodeBuilder\Domain\TemplatePathResolver\FilterPhpVersionDirectoryIterator;
+use Prophecy\PhpUnit\ProphecyTrait;
 use SplFileInfo;
 
 class FilterPhpVersionDirectoryIteratorTest extends TestCase
 {
-    use \Prophecy\PhpUnit\ProphecyTrait;
+    use ProphecyTrait;
 
     /**
      * @dataProvider provideDirectoriesToFilter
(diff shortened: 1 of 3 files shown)


lib/CodeBuilder/Tests/Unit/Domain/TemplatePathResolver/FilterPhpVersionDirectoryIteratorTest.php
```diff
@@ -56,7 +57,7 @@ private function createFakeDirectory(string $filename): SplFileInfo
 
     private function createFakeSplFileInfo(string $filename, bool $isDir = false): SplFileInfo
     {
-        $file = $this->prophesize(SplFileInfo::class);
+        $file = $this->prophesize(\Symfony\Component\Finder\SplFileInfo::class);
```
dantleech, line 60: not sure why this change? it should not be the FQN - should it be the Symfony SPL info?

dantleech:
all clear, thanks :) we should probably import the class rather than using the FQN but CS fixer should take care of that so not important.

lib/CodeBuilder/Tests/Unit/Domain/TemplatePathResolver/FilterPhpVersionDirectoryIteratorTest.php
gerardroche: This is the class that the Prophecy spltfileinfo patch checks against here: https://github.com/phpspec/prophecy/blob/master/src/Prophecy/Doubler/ClassPatch/SplFileInfoPatch.php#L119

gerardroche: Notice that the return type is still `\SplFileInfo` because we're only using the symfony one to avoid the deprecation. Apparently there is an internal implementation detail about `\SplFileInfo` that makes it difficult for Prophecy to prophesize it without arguments.

PR #2524, opened 2024-01-28
Fixing the documentation

mamazu:
I've built a little documentation parser and validator to validate that the code in the documentation is actually syntactically correct and it found somethings that I fixed.


dantleech:
thanks

PR #2527, opened 2024-01-31
Revert "Revert "Add snippet to type suggestions for object creation (…

dantleech:
…#2515)""

This reverts commit 0d1053036af67320a9c13524bd670b5eeb7e7688.

Have reverted the original PR as it causes a behavioral regression:

If I want to change the class:

```
new Barfoo()
```

to 

```
new Foo<>()
```

then it's completed as:

```
new Foo()()
```

### DIFF
--- a/lib/Completion/Bridge/TolerantParser/SourceCodeFilesystem/ScfClassCompletor.php
+++ b/lib/Completion/Bridge/TolerantParser/SourceCodeFilesystem/ScfClassCompletor.php
@@ -4,6 +4,7 @@ namespace Phpactor\Completion\Bridge\TolerantParser\SourceCodeFilesystem;
 
 use Generator;
 use Microsoft\PhpParser\Node;
+use Microsoft\PhpParser\Node\Expression\ObjectCreationExpression;
 use Microsoft\PhpParser\Node\QualifiedName;
 use Microsoft\PhpParser\ResolvedName;
 use Phpactor\ClassFileConverter\Domain\ClassName;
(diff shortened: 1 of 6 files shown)


dantleech:
cc @przepompownia

przepompownia:
In short, probably we need to detect if the next sibling of the node under cursor is a parenthesized expression. I looked for such method before you merged the reverted PR and the comment https://github.com/phpactor/phpactor/pull/2515#issuecomment-1911191638 is about it. 

If I remember correctly, first we need to implement a method to determine the next sibling.

przepompownia:
https://github.com/phpactor/phpactor/blob/1419d7137d5ead44b2f4d3ad3d15152792bb46d0/lib/Completion/Bridge/TolerantParser/ChainTolerantCompletor.php#L31-L33

With truncated source we cannot determine what is right after `$node`.

przepompownia:
@dantleech Is this an unsolvable problem here?

Maybe (at least for Neovim users) treesitter+completion plugin will be better in this case.

dantleech:
> unsolvable problem

without refactoring lots, yes I think so. this is also a code style choice, `new Foobar` is also valid PHP.

przepompownia:
It's a code style choice while we don't need to pass any parameter there.

przepompownia:
> > unsolvable problem
> 
> without refactoring lots, yes I think so. 

I guess the only thing left to do is close this issue.

dantleech:
ok, thanks for the effort. maybe one day I'll refactor the entire completion bit as it's a bit of a convoluted mess currently and this sort of thing will be easier.

przepompownia:
Warming up to change anything there is still also of some value for me. 

Let's look here: 
https://github.com/phpactor/phpactor/blob/1419d7137d5ead44b2f4d3ad3d15152792bb46d0/lib/Completion/Bridge/TolerantParser/WorseReflection/WorseClassMemberCompletor.php#L76

There is no truncation like in `ChainTolerantCompletor` so looking ahead is possible.

przepompownia:
> Maybe (at least for Neovim users) treesitter+completion plugin will be better in this case.

This early workaround for nvim-cmp seems to work: 
```lua

local function confirmDone(evt)
  local context = evt.entry.context
  if context.filetype ~= 'php' then
    return
  end

  if vim.startswith(context.cursor_after_line, '(') then
    return
  end

  local endRange = evt.entry.source_insert_range['end']
  vim.treesitter.get_parser(context.bufnr):parse({endRange.line, endRange.line})
  local node = vim.treesitter.get_node({pos = {endRange.line, endRange.character - 1}})

  if node:parent():type() ~= 'object_creation_expression' then
    return
  end

  vim.api.nvim_feedkeys('(', 'i', false)
end

cmp.event:on('confirm_done', confirmDone)
```

Issue #2522, opened 2024-01-26
`workspace/willRenameFiles` - destination file must exist before


przepompownia:
I don't know how many LSP clients can send `workspace/willRenameFiles` request and handle the response. Neovim comes with no handler and I didn't find any plugin for it. 

On the other side I can send such request and would see how the response if it worked. In https://github.com/phpactor/phpactor/blob/6cbae8d357ad10fe4a1932db4ca995069101e130/lib/Rename/Adapter/ClassMover/FileRenamer.php#L80 
I get `TextDocumentNotFound` because the new file doesn't exist yet. Is it intentional?

I check it with
```lua
local function willRenameCurrentBuffer(newPath)
  if not vim.startswith(newPath, '/') then
    newPath = vim.fs.joinpath(vim.uv.cwd(), newPath) -- replace vim.uv with vim.loop on <=0.9
  end
  local response = vim.lsp.buf_request_sync(0, 'workspace/willRenameFiles', {files = {{
    oldUri = vim.uri_from_bufnr(0), -- the current file
    newUri = vim.uri_from_fname(newPath)
  }}})

  vim.print(response)
end

vim.api.nvim_create_user_command('WillRenameCurrentBuffer', function (opts)
  willRenameCurrentBuffer(opts.args)
end, {nargs = 1, complete = 'file'})
```

PR #2531, opened 2024-02-04
Make `workspace/willRenameFiles` working 

przepompownia:


przepompownia:
@dantleech at the moment it seems to work 

- we need to verify if we need the part described in #2522 (currently commented out)
- I'm aware that it's badly designed at the moment - I mean using `WorkspaceEdit` directly in `FileRenamer`. Probably we need to define some equivalent of `Phpactor\Rename\Model\LocatedTextEditsMap` instead.

przepompownia:
To test it with Neovim (master) I use 
```lua
local function willRenameFilesHandler(response)
  for clientId, clientResponse in pairs(response) do
    local client = vim.lsp.get_client_by_id(clientId)
    if nil == client then
      vim.notify(('Client %s does not exist'):format(clientId), vim.log.levels.ERROR, {title = 'LSP: workspace/willRenameFiles'})

      return
    end

    vim.lsp.util.apply_workspace_edit(clientResponse.result, client.offset_encoding)
    -- vim.print(clientResponse.result)
  end
end

local function willRenameCurrentBuffer(newPath)
  if not vim.startswith(newPath, '/') then
    newPath = vim.fs.joinpath(vim.uv.cwd(), newPath)
  end
  vim.lsp.buf_request_all(
    0,
    'workspace/willRenameFiles', 
    {
      files = { { oldUri = vim.uri_from_bufnr(0), newUri = vim.uri_from_fname(newPath) } },
    },
    willRenameFilesHandler
  )
end
vim.api.nvim_create_user_command('WillRenameCurrentBuffer', function (opts)
  willRenameCurrentBuffer(opts.args)
end, {nargs = 1, complete = 'file'})
```

przepompownia:
> @dantleech at the moment it seems to work on simple cases but
> 
>     * we need to verify if we need the part described in [`workspace/willRenameFiles` - destination file must exist before #2522](https://github.com/phpactor/phpactor/issues/2522) (currently commented out)
> 
>     * I'm aware that it's badly designed at the moment - I mean using `WorkspaceEdit` directly in `FileRenamer`. Probably we need to define some equivalent of `Phpactor\Rename\Model\LocatedTextEditsMap` instead and convert it to `WorkspaceEdit` in the handler.

At the moment both above things are done: 
- `replaceDefinition()` is needed to edit the rename file itself before moving it (in the meaning of WorkspaceEdit changes)
- to avoid reinvention of models and use `RenameResult` in handler I changed both the adapter and handler into a pattern used in `ClassRenamer`.

przepompownia:
Composing the summarized `WorkspaceEdit` instance from component instances can be controversial. It assumes (relying on `LocatedTextEditConverter::toWorkspaceEdit` implementation) that only `documentChanges` field is not empty in each of them and we can ignore other fields. 

https://github.com/phpactor/phpactor/pull/2531/files#diff-39ca8889abb439b22964fc403e2e98d48192ef428529ede2bee62f84221faedfR60-R71

lib/Extension/LanguageServerRename/Handler/FileRenameHandler.php
```diff
@@ -39,14 +45,42 @@ public function methods(): array
...
+            $count = 0;
+            $documentChanges = [];
+            try {
+                foreach ($params->files as $rename) {
+                    $locatedEditMap = LocatedTextEditsMap::create();
+                    assert($rename instanceof FileRename);
 
-                $workspaceEdits = $workspaceEdits->merge(yield $this->renamer->renameFile(TextDocumentUri::fromString($rename->oldUri), TextDocumentUri::fromString($rename->newUri)));
-            }
+                    $renameGen = $this->renamer->renameFile(
```
dantleech, line 55: what is `renameGen`?

lib/Extension/LanguageServerRename/Handler/FileRenameHandler.php
```diff
@@ -39,14 +45,42 @@ public function methods(): array
...
         return call(function () use ($params) {
-            $workspaceEdits = LocatedTextEditsMap::create();
-            foreach ($params->files as $rename) {
-                assert($rename instanceof FileRename);
+            $count = 0;
+            $documentChanges = [];
+            try {
+                foreach ($params->files as $rename) {
+                    $locatedEditMap = LocatedTextEditsMap::create();
+                    assert($rename instanceof FileRename);
```
dantleech, line 53: shouldn't need `assert` here, `$params->files` is a generic.

lib/Rename/Adapter/ClassMover/FileRenamer.php
```diff
@@ -2,90 +2,79 @@
...
-        private ClassMover $mover
+        private ClassMover $mover,
     ) {
     }
 
-    /**
-     * @return Promise<LocatedTextEditsMap>
-     */
-    public function renameFile(TextDocumentUri $from, TextDocumentUri $to): Promise
+    public function renameFile(TextDocumentUri $from, TextDocumentUri $to): Generator
```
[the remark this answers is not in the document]
przepompownia: I switched to using generator in similar way like in `ClassRenamer` (please compare what happens there) to smuggle `RenameResult` through `getReturn()` and avoid changes in models (as I mentioned above).

lib/Extension/LanguageServerRename/Handler/FileRenameHandler.php
przepompownia: I looked how `RenameFilesParams` instance is created - indeed there is rather no way for which this assertion could be needed.

lib/Rename/Adapter/ClassMover/FileRenamer.php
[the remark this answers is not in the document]
przepompownia: Using Promise would require some new class (`Phpactor\Rename\Model\RenameEdit`?) composed of `LocatedTextEditsMap` and `RenameResult`.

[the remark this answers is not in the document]
przepompownia: `RenameEdit` class couldn't simply look like
```php
<?php

namespace Phpactor\Rename\Model;

class RenameEdit
{
    public function __construct(
        public readonly LocatedTextEditsMap $locatedTextEditsMap,
        public readonly RenameResult $renameResult,
    ) {
    } 
}
```
because it'd loss the order of edits. Such class should be similar to `WorkspaceEdit` that allows ordered edits of different types.

dantleech:
I;d really need to test this against a real client implementation, I initially developed it against the CoC implementation - what implementation is there for NVIM LSP? Is it supported by VS code?

przepompownia:
Honestly, at the moment I couldn't think of anything more than that using a simple prompt with path completion to send such request (like the example above) but as a [nvim-tree](https://github.com/nvim-tree/nvim-tree.lua) user I could extend it like here 

https://github.com/przepompownia/nvim-arctgx/blob/ef3fc1f4221bd82382abf1b5ff973debb415f651/lua/arctgx/pluginConfigs/nvim-tree.lua#L88-L101

I haven't tested it on nvim-oil yet maybe because built-in nvim feature `vim.lsp.util.apply_workspace_edit` (after recent fixes in master) performs [WorkspaceEdit](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspaceEdit) correctly.

Maybe I'll find some free time to install oil.nvim and CoC in separate environments. It's possible that some other clients do not apply `RenameResult` (but perform it independently) but it seems to be not a reason do skip it in in the response from the server. 


> The will rename files request is sent from the client to the server before files are actually renamed as long as the rename is triggered from within the client either by a user action or by applying a workspace edit.

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_willRenameFiles

lib/Rename/Adapter/ClassMover/FileRenamer.php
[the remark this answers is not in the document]
przepompownia: I converted the adapter to return Promise and created `RenameEdit` - initially as iterable class.

przepompownia:
Intentionally I didn't merge the recent master yet because of #2607.

przepompownia:
- https://github.com/antosha417/nvim-lsp-file-operations with https://github.com/nvim-tree/nvim-tree.lua works for me as expected on this branch
- oil.nvim uses `vim.lsp.util.apply_workspace_edit` from Neovim (like in my example command), seems.

lib/Rename/Model/RenameEdit.php
```diff
@@ -0,0 +1,18 @@
...
+namespace Phpactor\Rename\Model;
+
+use ArrayIterator;
+
+/**
+ * @extends ArrayIterator<int, LocatedTextEditsMap|RenameResult>
+ */
+class RenameEdit extends ArrayIterator
+{
+    public function __construct(LocatedTextEditsMap|RenameResult|null ...$edits)
```
dantleech, line 12: why is this a variadic?

lib/Extension/LanguageServerRename/Util/RenameEditConverter.php
```diff
@@ -5,20 +5,46 @@
...
+use Phpactor\Rename\Model\RenameEdit;
 use Phpactor\Rename\Model\RenameResult;
 use Phpactor\LanguageServerProtocol\RenameFile;
 use Phpactor\LanguageServerProtocol\TextDocumentEdit;
 use Phpactor\LanguageServerProtocol\WorkspaceEdit;
 use Phpactor\LanguageServer\Core\Workspace\Workspace;
 use Phpactor\TextDocument\TextDocumentLocator;
 
-final class LocatedTextEditConverter
+final class RenameEditConverter
```
dantleech, line 16: rename to WorkspaceRenameEditsConverter as with the other renaming.

lib/Rename/Model/RenameEdit.php
przepompownia: Only for enforcing types without using docblock.

dantleech: Ok, let's use the docblock then :)

przepompownia: No problem, regardless, I'd like to know why you think this is probably bad practice (especially that in practice the constructor gets max two arguments).

przepompownia:
I'm confused about why phpstan is complaining about changes unrelated to this PR after merge #2614

lib/Rename/Model/RenameEdit.php
dantleech: because you need to iterate over them and check `instanceof` to get either the rename result or the text edit map. it is implementing the wrong data structure in order to avoid using a docblock.

przepompownia: - The order of _documentCnages_ matters thus I used array here (imagine [TextDocumentEdit](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentEdit) on `file1` **and then** [RenameFile](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#renameFile) from `file1` to `file2`). Assigning LocatedTextEdits and RenameResult to properties loses the (explicit) order. 
- I removed `null` when converted variadics to docblock. It was used only in tests.

dantleech: Where is this used? and why array_filter here?

https://github.com/przepompownia/phpactor/blob/1dc9c4823f550e26b843a1ad8448fb59714cb5d0/lib/Rename/Adapter/ClassMover/FileRenamer.php#L69-L72

przepompownia: > Where is this used?

`FileRenameHandler`

If `WorkspaceRenameEdits` constructor allowed null, `array_filter` were redundant also in the rest calls. `WorkspaceEdit`'s `documentChanges` doesn't allow `null` inside array so here too.

dantleech: I'm still confused. The result of the Promise in `lib/Rename/Adapter/ClassMover/FileRenamer.php` is:

```
            return new WorkspaceRenameEdits(array_filter([
                LocatedTextEditsMap::fromLocatedEdits($locatedEdits),
                new RenameResult($from, $to),
            ]));
```

i.e. why not:

```
            return new WorkspaceRenameEdits(
                LocatedTextEditsMap::fromLocatedEdits($locatedEdits),
                new RenameResult($from, $to),
            );
```

The `array_filter` can be removed, and it doesn't need to be an array?

przepompownia: > It doesn't need to be an array?
It's (similarly to WorkspaceEdit's documentChanges elements) the only way to keep the order of operations here.

dantleech: Just to clarify though, it's not needed now?

dantleech: I mean, we are also passing an aggregate of located text edits, rather than the individual edits. So it's a bit ... weird. If this is gojg to be a list of operations then we should or bably not use the LocatedTextEditsMap here, as that' not an atomic operation.

przepompownia: I agree that this separation is far from the best and even. I would like to leave some design assumptions to you, but I realize that this may mean a lot of reconstruction (and, in effect, wait possibly long time). Please feel free to replace it with something better. I'm not attached in any way the current form of solution. It would be good to have working such LSP request.

lib/Extension/LanguageServerRename/Handler/FileRenameHandler.php
```diff
@@ -39,14 +43,34 @@ public function methods(): array
...
+                    }
+                }
 
-            return $this->converter->toWorkspaceEdit($workspaceEdits);
+                return new WorkspaceEdit(documentChanges: $documentChanges);
+            } catch (CouldNotRename $error) {
+                $previous = $error->getPrevious();
+
+                $this->clientApi->window()->showMessage()->error(sprintf(
+                    $error->getMessage() . ($previous?->getTraceAsString() ?? '')
```
dantleech, line 69: is `sprintf` missing a format string? `%s %s` why do we get the trace from `$previous`?

dantleech:
I've made a PR on your PR: https://github.com/przepompownia/phpactor/pull/1 would be great if you can verify that it's all working.

lib/Extension/LanguageServerRename/Handler/FileRenameHandler.php
przepompownia: I copied catching `CouldNotRename` from `RenameHandler` to have the same way of error handling and don't know what the trace was needed for in the original place: 

https://github.com/phpactor/phpactor/blame/a23130918b61c3f45f4a403abe56b5a06be80de9/lib/Extension/LanguageServerRename/Handler/RenameHandler.php#L93

z01cbrag:
Any updates on this? I'd love to have this feature :heart:

dantleech:
Which client are you using that can take advantage of it?

dantleech:
So there's this plugin for neovim at least which integrates with neo-tree and nvim-tree: https://github.com/antosha417/nvim-lsp-file-operations will try this out soon, sorry for the delay.

z01cbrag:
There's also https://github.com/stevearc/oil.nvim, that's what I'm using and it has support for this.

Issue #2540, opened 2024-02-11
Error with const in enum (Case does not exist on enum)


Eviljeks:
Firstly, I'd like to extend my sincerest gratitude to all the authors and contributors who have dedicated their time and effort to develop and maintain this invaluable repository.
I've been utilizing a phpactor with vs code for the past few days and found an error with **const in enums** (see an image). 
<img width="732" alt="зображення" src="https://github.com/phpactor/phpactor/assets/22118652/8c16ded4-ac17-40a8-a364-a2e380f63627">
After a few hours of learning repo I guess I found a solution (pr).
It would be awesome if you have time to review or leave some other feedback about reproducing or other staff.

PR #2541, opened 2024-02-11
Fix const in enums

Eviljeks:
Fixes https://github.com/phpactor/phpactor/issues/2540

### DIFF
--- a/lib/WorseReferenceFinder/Tests/Unit/WorseReflectionDefinitionLocatorTest.php
+++ b/lib/WorseReferenceFinder/Tests/Unit/WorseReflectionDefinitionLocatorTest.php
@@ -310,9 +310,6 @@ class WorseReflectionDefinitionLocatorTest extends DefinitionLocatorTestCase
 
     public function testLocatesCase(): void
     {
-        if (!defined('T_ENUM')) {
-            $this->markTestSkipped('PHP8.1');
-        }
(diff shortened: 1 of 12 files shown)


lib/WorseReferenceFinder/Tests/Unit/WorseReflectionDefinitionLocatorTest.php
```diff
@@ -322,6 +322,20 @@ public function testLocatesCase(): void
         $this->assertTypeLocation($location->first(), 'FoobarEnum.php', 24, 33);
     }
 
+    public function testLocatesEnumConst(): void
+    {
+        if (!defined('T_ENUM')) {
```
dantleech, line 327: Can drop this as the minimum version is 8.1 (which includes enums)

dantleech (commented):
Very nice, thankyou. Just one small comment.

lib/WorseReferenceFinder/Tests/Unit/WorseReflectionDefinitionLocatorTest.php
Eviljeks: I dropped all checks for `defined('T_ENUM')`, is it ok?

dantleech:
yes perfect, thanks!

PR #2546, opened 2024-02-17
Class references: avoid redundant results if implementation implements another interface

przepompownia:


dantleech:
Thanks, can you explain why  we remove this code and if it fixes an issue which issue it fixes?

przepompownia:
As in https://github.com/phpactor/phpactor/pull/2261#issuecomment-1947592610 - look at the references to `Watcher` in `lib/Indexer/Extension/IndexerExtension.php` - before #2261 and on master branch. The removed part was used only to generate unrelated results.

przepompownia:
Example from Phpactor: 
```php
class LanguageServerWatcher implements Watcher, WatcherProcess, ListenerProviderInterface
```
```php
class LazyAggregateProvider implements ListenerProviderInterface
```
the above definitions are enough to show `LazyAggregateProvider` in the list of references to `Watcher` on `master`. I hope that the test added here covers such case.

przepompownia:
Btw. it's a side effect of investigating why the old file entry after renaming is still present in the index.

dantleech:
at some point I need to investigate exactly what PHPStorm does when finding references, there is a good chance it only returns direct symbol references, and does not take inheritance into account at all.

PR #2550, opened 2024-02-21
Run Composer post-install-cmd with the same PHP executable

MatmaRex:
As documented here:
https://getcomposer.org/doc/articles/scripts.md#executing-php-scripts

Add `@php` to the `post-install-cmd` script, so that bin/phpactor is executed using the same executable used to run Composer itself. This is important when you have several versions of PHP installed, and the default one isn't at least PHP 8.1 as required by Phpactor.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -111,7 +111,7 @@
     ],
     "scripts": {
         "post-install-cmd": [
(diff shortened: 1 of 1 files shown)


dantleech:
Ta!

PR #2552, opened 2024-02-21
Fix JSON encoding in default config

MatmaRex:
`$this->schemaPath` always contains backslashes on Windows, and presumably may contain backslashes and quotes on Linux too.


dantleech:
image

:think

dantleech:
happy to merge this, but there are CI issues, you can run `composer integrate` to run most of the checks that happen in CI (it will also fix the CS). You'll probably need to handle the case where `json_encode` returns NULL.

MatmaRex:
I haven't gotten `composer integrate` running yet, but I'll have a look at that next.

PR #2554, opened 2024-02-22
Run various PHP subprocesses using `PHP_BINARY`

MatmaRex:
This ensures that the subprocesses use the same PHP executable as the main one, which is important in case the user has multiple PHP versions installed.

It's also needed for Windows support, because it can't execute PHP scripts directly (only files with specific extensions are executable).


dantleech:
thanks!

PR #2555, opened 2024-02-22
Get PHPUnit running on Windows

MatmaRex:
With these patches, `vendor/bin/phpunit` runs to completion on my Windows machine, without crashing the test runner or hanging forever. 

```
Tests: 3550, Assertions: 6277, Errors: 188, Failures: 410, Skipped: 5, Risky: 1.
```

:)

(the number of errors varies slightly, something must be flapping)

### DIFF
--- a/lib/Amp/Tests/Process/ProcessBuilderTest.php
+++ b/lib/Amp/Tests/Process/ProcessBuilderTest.php
@@ -13,7 +13,7 @@ class ProcessBuilderTest extends TestCase
 
     public function testBuildProcess(): void
     {
-        $process = ProcessBuilder::create(['export'])->build();
+        $process = $this->makeEnvDumpProcess()->build();
         $pid = wait($process->start());
         $exitCode = wait($process->join());
         self::assertEquals(0, $exitCode);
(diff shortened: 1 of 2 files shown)


MatmaRex:
Huh, that's weird.

lib/Amp/Tests/Process/ProcessBuilderTest.php
```diff
@@ -22,28 +22,33 @@ public function testBuildProcess(): void
...
         self::assertStringContainsString('myenvvar', $out);
         self::assertStringContainsString(self::PARENT_PROCESS_ENV_VAR, $out);
+        $exitCode = wait($process->join());
+        self::assertEquals(0, $exitCode);
         putenv('FOO');
     }
+
+    private function makeEnvDumpProcess(): ProcessBuilder
+    {
+        return ProcessBuilder::create([PHP_BINARY, '-r', 'var_dump($_ENV);']);
```
dantleech, line 52: confused why this is necesary when we use `mergeParentEnv` - do processes not receive env vars on windows?

mamazu: Why don't we use the php `[getenv](https://www.php.net/manual/en/function.getenv.php)`? Since php8 it allows the parameter to be null and return all env vars.

MatmaRex: They do, but I'm afraid I don't really understand the question. ProcessBuilder (and Process) override that behavior for some reason, and this is a test for it.

MatmaRex:
I figured out why it wasn't working, apparently `$_ENV` isn't always set (https://www.php.net/manual/en/reserved.variables.environment.php#98113). Thanks for the pointer to `getenv()` @mamazu (I saw your comment before it was deleted).

MatmaRex:
I am now even more confused by the new failure, which is in a test case I didn't touch. Is this a known flaky test?

```
There was 1 failure:

1) Phpactor\Indexer\Tests\Adapter\Tolerant\TolerantIndexBuilderTest::testDoesNotRemoveExisting
Failed asserting that actual size 4 matches expected size 3.

/home/runner/work/phpactor/phpactor/lib/Indexer/Tests/Adapter/IndexBuilderTestCase.php:579
```

dantleech:
>  Is this a known flaky test?
yes it is, you can ignore it :)

lib/Amp/Tests/Process/ProcessBuilderTest.php
dantleech: ok i see, ignore me

PR #2558, opened 2024-02-24
GH-2538: Fix for diagnostic/linting concurrency.

### LINKED ISSUE #2538: phpactor is killing my cpu
Hi,
not sure what's happening, but my cpu burn if I try to use phpactor
image
this is after less than 1 minute on my file.
I'm using neovim with nvim_cmp and lspconfig. I've installed phpactor through mason and set it up using mason_lspconfig.
here is my setup handler:
 ```lua
   mason_lsp.setup_handlers {
        -- The first entry (without a key) will be the default handler
        -- and will be called for each installed server that doesn't have
        -- a dedicated handler.
        function (server_name) -- default handler (optional)
          -- local capabilities = require('cmp_nvim_lsp').default_capabilities()
          local lsp_config = require("lspconfig")
          lsp_config[server_name].setup ({
            root_dir = function(file_name)
              return lsp_config.util.find_git_ancestor(file_name)
            end,
            -- capabilities = capabilities
          })
        end
      }
    end
```
nothing fancy here
the scenario to get to the thumbnail is this.
when I open a file, one process pops at 100% on the cpu. I thought of indexing and tried to let it do it's thing with no success. it never leave. when i try to type something, everytime the completion pops up, a new process is created which also uses 100% cpu. 
so just typing a few words gets to the situation above.
and even when I quit neovim, all the processes stays here and I have to kill them manually.
Any clues on what is happening?
Tell me if you need anything else to diagnostic this
Cheers

dantleech:
Hopefully fixes #2538 

This PR will:

- Fix a wrong assumption that the text document version should b e provided on `save` by the client. We now get the last known version from the Workspace.
- Do not run the linters if the text document (which may have been queued) is older than the last linted version of that document.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ Improvements:
 
 Bug fixes:
 
+  - Fix diagnostic process concurrency and do not lint outdated files #2538
   - Fix detection of import used relatively in an annotation #2539
   - Fix PHAR crashing issue on PHP8.3 #2533
   - Fix UTF-16 conversion for LSP #2530 #2557
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
         "phpactor/amp-fswatch": "^0.3.0",
         "thecodingmachine/safe": "^1.0",
         "phpactor/phly-event-dispatcher": "^2.0.0",
-        "phpactor/language-server": "^6.1.1",
+        "phpactor/language-server": "^6.1.2",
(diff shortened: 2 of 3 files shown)


dantleech:
>  I still get several processes: one for each file I have open in nvim and another one.

it's still expected to get one per file - has this PR improved it at all? I guess even if we throttled it to one file you'd still be waiting for it to analyze the open files.

dantleech:
this has made it into master now anyway, so closing.

PR #2560, opened 2024-02-24
Bump amp/process to fix #2516

### LINKED ISSUE #2516: Crash when removing folders in sublime text
phpactor crashes including crashing other things like firefox, terminal, and also gnome shell itself. Obviously this is quiet a severe crash. 

I'm using phpactor in [Sublime Text LSP](https://github.com/sublimelsp/LSP) and can consistently reproduce the crash when removing several folders in quick succession. It's not the only context in which the crash occurs, but I can reproduce that one.

I reported the issue in the Sublime text [LSP plugin](https://github.com/sublimelsp/LSP/issues/2380), but I've narrowed the issue down to phpactor. There are additional details on the issue, but I'll add some here.

My Sublime Text lsp config, the configurations outside of "clients" don't seem to be relevant, but I'll show anyway:

```
{

    "diagnostics_delay_ms": 50,
    "diagnostics_gutter_marker": "dot",
    "log_debug": false,
    "log_server": false,
    "show_code_actions": "",
    "show_diagnostics_count_in_view_status": false,
    "show_diagnostics_in_view_status": false,
    "show_inlay_hints": false,
    "show_multiline_diagnostics_highlights": false,
    "show_view_status": false,

    "diagnostics_highlight_style": {
        "error": "",
        "warning": "",
        "info": "stippled",
        "hint": "stippled"
    },

    "clients": {
        "phpactor": {
            "command": [
                "php",
                "-d",
                "xdebug.mode=off",
                "/usr/local/bin/phpactor",
                "language-server",
            ],
            "enabled": true,
            "languageId": "php",
            "scopes": [
                "source.php",
                "embedding.php"
            ],
            "syntaxes": [
                "Packages/PHP/PHP.sublime-syntax"
            ],
        },
    },
}
```

I don't think it probably matters how I have it installed, but I have phpactor installed via composer. The `/usr/local/bin/phpactor` is symlinked to `bin/phpactor` in my cloned phpactor repository which is installed, for example, like this':

```
cd ~
git clone https://github.com/phpactor/phpactor
cd phpactor
composer install --no-dev --optimize-autoloader
cd /usr/local/bin
sudo ln -s ~/phpactor/phpactor/bin/phpactor
``` 

## To reproduce the issue

1. Add several project folders to sublime 
2. Open a php file (opening a php file initialises the phpactor language-server).
3. Remove two of the folders one after the other quickly (less than about 1 sec apart)

I use my [Sesame](https://packagecontrol.io/packages/Sesame) plugin so it's easy to remove folders quickly, but you can do it without Sesame, you just need to be quick with you mouse.

I enabled [phpactor logging](https://phpactor.readthedocs.io/en/master/development/debugging.html#logging) but there is nothing of note in the log.

**Other notes**

- Sublime 4169
- Ubuntu 22.04
- Gnome Shell 42.9
- Windowing System Wayland

In Sublime the code to remove a folder is:

```
def _remove_folder(window: Window, folder: str) -> None:
    window.run_command('remove_folder', {
        'dirs': [folder]
    })
```

So you should be able to run the command via the console if you like:

```
>>> sublime.active_window().run_command('remove_folder', {'dirs': ['/path/to/folder']})
```

Otherwise right click the folder in side bar and select Remove Folder from Project.

dantleech:
Includes newly tagged `amphp/process` thanks to @gerardroche - can  you test the fix?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ Improvements:
 
 Bug fixes:
 
+  - Upgrade `amp/process` to fix #2516 thanks to @gerardroche
   - Fix division by zero edge case
   - Fix crash if referenced file no longer exists on class rename #2518
   - Fix diagnostic process concurrency and do not lint outdated files #2538
--- a/composer.json
+++ b/composer.json
@@ -16,6 +16,7 @@
         "symfony/console": "^5.0",
         "dantleech/invoke": "^2.0",
         "phpactor/amp-fswatch": "^0.3.0",
+        "amphp/process": "^1.1.5",
         "thecodingmachine/safe": "^1.0",
         "phpactor/phly-event-dispatcher": "^2.0.0",
         "phpactor/language-server": "^6.1.1",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "cde4dde661f5fe0a177855700b7c5e42",
+    "content-hash": "a044eaa949abdb83f0b7016bce043818",
     "packages": [
         {
             "name": "amphp/amp",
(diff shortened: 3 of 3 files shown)


gerardroche:
Yep, fixed. Thanks.

PR #2561, opened 2024-02-25
Matching class names case insensitive

mamazu:
This only changes the behaviour of the ShortName search as searching with an FQN sould be case sensitive. Or what do you think?

### DIFF
--- a/lib/Indexer/Model/Query/Criteria/ShortNameBeginsWith.php
+++ b/lib/Indexer/Model/Query/Criteria/ShortNameBeginsWith.php
@@ -8,8 +8,11 @@ use Phpactor\Indexer\Model\Record\HasShortName;
 
 class ShortNameBeginsWith extends Criteria
 {
-    public function __construct(private string $name)
+    private string $name;
+
+    public function __construct(string $name)
     {
+        $this->name = strtolower($name);
     }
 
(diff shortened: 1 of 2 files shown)


lib/Indexer/Model/Query/Criteria/ShortNameBeginsWith.php
```diff
@@ -22,6 +25,6 @@ public function isSatisfiedBy(Record $record): bool
             return false;
         }
 
-        return str_starts_with($record->shortName(), $this->name);
+        return str_starts_with(strtolower($record->shortName()), $this->name);
```
przepompownia, line 28: This place is tempting to try (probably slow) fuzzy matching if `str_starts_with` didn't match:


```suggestion
              if (str_starts_with(strtolower($record->shortName()), $this->name)) {
            return true;
        }

        $regex = '/' . implode('.*', array_map(preg_quote(...), str_split($this->name))) . '/i';

        return preg_match($regex, $record->shortName()) === 1;
  
```
image

mamazu:
Closed replaced by #2562

przepompownia:
On simple benchmark in #2563 I see that using a class compound of two Criteria class is significantly slower so closing this PR seems to be good idea in favor of single class implementation.

PR #2565, opened 2024-02-26
Revert "AttributeCompletor: add snippet to class suggestion"

przepompownia:
Reverts phpactor/phpactor#2508

It's similar to 0d10530


przepompownia:
I have no idea how to parse nodes after cursor. Maybe method completion could be some hint but I didn't verify what happens on the Phpactor side and what on the client side.

PR #2568, opened 2024-02-27
Fixes to path handling on Windows

MatmaRex:
Paths don't start with '/', so use helpers for absolute/relative path, and may use both '/' and '\\', so canonicalize before comparing them.

This resolves 26 test failures on Windows.

### DIFF
--- a/lib/Filesystem/Domain/FilePath.php
+++ b/lib/Filesystem/Domain/FilePath.php
@@ -32,8 +32,9 @@ final class FilePath
     public static function fromParts(array $parts): FilePath
     {
         $path = Path::join(...$parts);
-        if (!str_starts_with($path, '/')) {
-            $path = '/'.$path;
+        if (!Path::isAbsolute($path)) {
+            // Not sure if this makes sense… Maybe it should just throw?
+            $path = Path::makeAbsolute($path, '/');
         }
 
         return self::fromString($path);
@@ -107,7 +108,7 @@ final class FilePath
 
     public function isWithin(FilePath $path): bool
     {
-        return str_starts_with($this->path(), $path->path().'/');
+        return Path::isBasePath($path->path(), $this->path());
     }
 
     public function isWithinOrSame(FilePath $path): bool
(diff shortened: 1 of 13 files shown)


lib/Filesystem/Domain/FilePath.php
```diff
@@ -32,8 +32,9 @@ public static function fromString(string $string): FilePath
     public static function fromParts(array $parts): FilePath
     {
         $path = Path::join(...$parts);
-        if (!str_starts_with($path, '/')) {
-            $path = '/'.$path;
+        if (!Path::isAbsolute($path)) {
+            // Not sure if this makes sense… Maybe it should just throw?
+            $path = Path::makeAbsolute($path, '/');
```
dantleech, line 37: i think the _idea_ is that this is _always_ absolute, so `fromParts(['home','daniel',])`

dantleech:
nice, thanks

lib/Filesystem/Domain/FilePath.php
MatmaRex: To elaborate, there are two things that confuse me:
* Where does data in this format come from? I'm not aware of any filesystem APIs that return paths like that, and I've never seen anyone use that format in their configuration. I also couldn't find what in your code generates them. I'm like 70% sure that this code path is unused apart from tests.
* How is this supposed to work with Windows paths, where absolute paths must have a drive letter (e.g. "C:") in addition to the list of path segments? The way I made it work here, this function actually returns a drive-relative path (e.g. "\Users\daniel"), not an absolute one ("C:\Users\daniel"), and the drive letter will be resolved against the current working directory when trying to read from the path. This will probably often work, but it sometimes won't.

dantleech: it's used here:

https://github.com/phpactor/phpactor/blob/master/lib/Filesystem/Adapter/Simple/SimpleFilesystem.php

i guess it won't work for Windows... but feel free to throw an exception here, if the tests pass then it's all good.

MatmaRex: Hmm, just this one method?

```php
    public function createPath(string $path): FilePath
    {
        if (Path::isRelative($path)) {
            return FilePath::fromParts([$this->path->path(), $path]);
        }

        ...
    }
```

If `$this->path->path()` is an absolute path, and it looks like it should be, then that will be fine.

PR #2572, opened 2024-02-27
Fix parsing/generating Windows-style file: URIs

MatmaRex:
This fixes features like "go to definition" on Windows.

In a file: URI, after the scheme, there are two slashes, then the 'host' (usually empty), then another slash, then the 'path'. For example:

  file:///foo/bar.php
  file:///C:/foo/bar.php

That's three leading slashes total, and the third slash is not a part of the path, but a part of the URI format itself; it just looks like that with Unix-style filesystem paths.

Unfortunately, this means you can't just prepend 'file://' to a filesystem path and get a valid file: URI. Fortunately, `TextDocumentUri::fromString(…)` accepts filesystem paths too, so document that and use it in test cases. Replace invalid URIs with two slashes in other test cases too to avoid confusion.

See: https://en.wikipedia.org/wiki/File_URI_scheme

### DIFF
--- a/lib/TextDocument/TextDocumentUri.php
+++ b/lib/TextDocument/TextDocumentUri.php
@@ -21,9 +21,12 @@ class TextDocumentUri
         if ($this->scheme === self::SCHEME_UNTITLED) {
             return sprintf('%s:%s', $this->scheme, $this->path);
         }
-        return sprintf('%s://%s', $this->scheme, $this->path);
+        return sprintf('%s:///%s', $this->scheme, ltrim($this->path, '/'));
     }
 
+    /**
+     * Construct a TextDocumentUri from a URI string or a filesystem path.
+     */
     public static function fromString(?string $uri): self
     {
         if ($uri === null || $uri === '') {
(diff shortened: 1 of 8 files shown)


MatmaRex:
This is a bit of a follow-up for #2568, after I better understood how things work.

lib/TextDocument/TextDocumentUri.php
```diff
@@ -39,35 +42,46 @@ public static function fromString(?string $uri): self
...
+                throw new InvalidUriException(sprintf(
+                    'Path for file:// must be absolute, got "%s"',
+                    $uri
+                ));
+            }
 
-        $components['path'] = Path::canonicalize($components['path']);
+            $components['path'] = Path::canonicalize($components['path']);
+
+        } else {
```
dantleech, line 65: note in general I avoid using `else` in this code base in favour of _early returns_ - I don't want to make you jump through hoops, but do you think you could refactor to remove the `else`? otherwise happy to merge it and we can avoid it in the future.

MatmaRex: Sure, I don't really agree with that blog post, but I'm happy to follow your conventions. I'll try to keep it in mind for any future patches.

dantleech:
nice work!

PR #2582, opened 2024-03-02
Remove /doc/_build from gitignore

lens0021:
It seems an outdated entry. I've searched `_build` in the code but nothing found but this gitignore.

And `make docs` builds the output at `/build/html`.


dantleech:
Thanks

PR #2593, opened 2024-03-09
Include excluded files

### LINKED ISSUE #2485: Include and exclude patterns
I am using phpactor for a symfony project and recommended settings for the indexer are the following:

```json
{
    "indexer.exclude_patterns": [
        "/vendor/**/Tests/**/*",
        "/vendor/**/tests/**/*",
        "/vendor/composer/**/*",
        "/var/cache/**/*",
        "/**/Workspace/**/*"
    ],
    "indexer.include_patterns": [
        "/var/cache/**/Symfony/Config/**/*.php",
        "/**/*.php"
    ]
}
```

However the include patterns contain a pattern for the Config Objects that is excluded by the exclude pattern and this will result in Phpactor not being able to index files.

mamazu:
Fixes #2485 

The include and exclude logic now allows for included paths inside excluded paths.
Also the tests for the file provider now also check that it returns the correct files and not just the count.

## How this worked before
Call `includes` to filter the list of files to include and then call `exclude` to remove files from that list that are excluded.

## How it works now
Build a map of paths and if they are included or not and tries to match all files against that map to see if they should be included or not.

### DIFF
--- a/lib/Filesystem/Domain/FileList.php
+++ b/lib/Filesystem/Domain/FileList.php
@@ -86,19 +86,49 @@ class FileList implements Iterator
     }
 
     /**
-     * @param string[] $globPatterns
+     * @param string[] $includePatterns
+     * @param string[] $excludePatterns
      */
-    public function excludePatterns(array $globPatterns): self
+    public function includeAndExclude(array $includePatterns, array $excludePatterns): self
     {
-        return $this->filter(function (SplFileInfo $info) use ($globPatterns) {
-            foreach ($globPatterns as $pattern) {
-                if (Glob::match($info->getPathname(), $pattern)) {
-                    return false;
+        // Building map of include paths that are sub paths of excludes so that we can include them again
+        $includedExcludes = [];
+        foreach($excludePatterns as $excludePattern) {
+            foreach($includePatterns as $includePattern) {
+                if (Glob::match($includePattern, $excludePattern)) {
+                    $includedExcludes[$excludePattern][] = $includePattern;
                 }
+            }
+        }
+
+        $this->includePatterns($includePatterns);
+
+        return $this->filter(function (SplFileInfo $info) use ($excludePatterns, $includedExcludes) {
+            foreach ($excludePatterns as $pattern) {
+                if (!Glob::match($info->getPathname(), $pattern)) {
+                    continue;
+                }
+
+                $includePatterns = $includedExcludes[$pattern] ?? [];
+                foreach($includePatterns as $includePattern) {
+                    if (Glob::match($info->getPathname(), $includePattern)) {
+                        return true;
+                    }
+                }
+                return false;
             }
 
             return true;
         });
+
+    }
+
+    /**
+     * @param string[] $globPatterns
+     */
+    public function excludePatterns(array $globPatterns): self
+    {
+        return $this->includeAndExclude([], $globPatterns);
     }
 
     /**
--- a/lib/Filesystem/Domain/FileList.php
+++ b/lib/Filesystem/Domain/FileList.php
@@ -92,6 +92,10 @@ class FileList implements Iterator
     public function includeAndExclude(array $includePatterns = [], array $excludePatterns = []): self
     {
         $inclusionMap = [];
+        if ($includePatterns === []) {
+            $inclusionMap['/**/*'] = true;
+        }
+
         foreach ($includePatterns as $includePattern) {
             $inclusionMap[$includePattern] = true;
         }
--- a/lib/Filesystem/Domain/FileList.php
+++ b/lib/Filesystem/Domain/FileList.php
@@ -89,57 +89,23 @@ class FileList implements Iterator
      * @param string[] $includePatterns
      * @param string[] $excludePatterns
      */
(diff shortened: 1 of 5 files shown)


lib/Filesystem/Domain/FileList.php
```diff
@@ -86,30 +86,50 @@ public function byExtensions(array $extensions): self
...
+            $inclusionMap[$includePattern] = true;
+        }
+        foreach ($excludePatterns as $excludePattern) {
+            $inclusionMap[$excludePattern] = false;
+        }
+
+        // Sort map by keys so that more specific paths are getting matched first
+        uksort($inclusionMap, function (string $x, string $y) {
+            $partsX = explode(DIRECTORY_SEPARATOR, $x);
+            $partsY = explode(DIRECTORY_SEPARATOR, $y);
```
dantleech, line 109: maybe rename to 1 and 2 instead of X and Y or even A and B? x+y I kinda reserved for co-ordinates.

lib/Filesystem/Domain/FileList.php
```diff
@@ -86,30 +86,50 @@ public function byExtensions(array $extensions): self
...
+            foreach ($partsX as $i => $pathPartX) {
+                if ($pathPartX === '**' || $pathPartX === '*') {
+                    return 1;
+                }
+
+                $compare = strcmp($pathPartX, $partsY[$i]);
+                if($compare !== 0) {
+                    return $compare;
+                }
+            }
```
dantleech, line 125: can you add a commenta about what this is doing? i.e. `if the segments are the same length then...` ??

lib/Filesystem/Domain/FileList.php
```diff
@@ -86,30 +86,50 @@ public function byExtensions(array $extensions): self
...
+        foreach ($excludePatterns as $excludePattern) {
+            $inclusionMap[$excludePattern] = false;
+        }
+
+        // Sort map by keys so that more specific paths are getting matched first
+        uksort($inclusionMap, function (string $x, string $y) {
+            $partsX = explode(DIRECTORY_SEPARATOR, $x);
+            $partsY = explode(DIRECTORY_SEPARATOR, $y);
+            // Longer paths should come first
+            $countDiff = -(count($partsX) <=> count($partsY));
```
dantleech, line 111: maybe remove the sign here and the brackets and return `-$countDiff`

lib/Filesystem/Domain/FileList.php
```diff
@@ -86,30 +86,50 @@ public function byExtensions(array $extensions): self
...
+                if ($pathPartX === '**' || $pathPartX === '*') {
+                    return 1;
+                }
+
+                $compare = strcmp($pathPartX, $partsY[$i]);
+                if($compare !== 0) {
+                    return $compare;
+                }
+            }
+            return 0;
```
dantleech, line 126: else they are ... exactly the same!? or how would it get here?

dantleech (commented):
Looks OK, left a couple of minor comments :+1:

lib/Filesystem/Domain/FileList.php
mamazu: Yeah, the docs also have a and b for variable names.

lib/Filesystem/Domain/FileList.php
mamazu: What exactly should I comment here. If they are the same then their order doesn't matter.

lib/Filesystem/Domain/FileList.php
mamazu: Done

lib/Filesystem/Domain/FileList.php
dantleech: explaining what the different branches mean. It's not obvious what `==='**` and `==='*'` mean and why we return. what happens with `baz*` or `*baz` or with `*/baz/*`?

lib/Filesystem/Domain/FileList.php
```diff
@@ -86,30 +86,51 @@ public function byExtensions(array $extensions): self
...
+        foreach ($excludePatterns as $excludePattern) {
+            $inclusionMap[$excludePattern] = false;
+        }
+
+        // Sort map by keys so that more specific paths are getting matched first
+        uksort($inclusionMap, function (string $a, string $b) {
+            $partsA = explode(DIRECTORY_SEPARATOR, $a);
+            $partsB = explode(DIRECTORY_SEPARATOR, $b);
+            $countDiff = count($partsA) <=> count($partsB);
+            if ($countDiff !== 0) {
```
dantleech, line 111: this is the common case or? why not `strlen` compare before exploding instead? should be more performant.

lib/Filesystem/Domain/FileList.php
```diff
@@ -86,30 +86,51 @@ public function byExtensions(array $extensions): self
...
+                // Longer paths should come first
```
dantleech, line 112: paths with more segments.

lib/Filesystem/Domain/FileList.php
mamazu: Oh yeah wildcards within a path aren't handled. That's a valid point.

Issue #2412, opened 2023-10-24
Index PHAR files


calebdw:
Hello!

Would it be possible for phpactor to index PHAR files? I was attempting to write a custom PHPStan rule and apparently everything is buried inside the PHAR which isn't indexed and therefore has no autocompletion:

https://github.com/phpstan/phpstan/issues/9819#issuecomment-1697291398

Thanks!

CalebDW:
It looks like the namespace is somewhat detected in the `use` autocomplete, but not everything is found. 


![image](https://github.com/phpactor/phpactor/assets/4176520/da9fbc1d-1ab7-4523-9721-2572720ec51f)

![image](https://github.com/phpactor/phpactor/assets/4176520/77eb05ca-1866-43c2-bd10-96a69c3215b1)

![image](https://github.com/phpactor/phpactor/assets/4176520/22a53223-ece1-4553-8494-82d2f2c713f8)


![image](https://github.com/phpactor/phpactor/assets/4176520/44b3da80-c9be-4c9f-bb24-ce8016a14a92)

dantleech:
Can Neovim/VSCode open files PHAR archives?  It might be possible be but it would be involved at this point.

dantleech:
hmm, then maybe :thinking: you could experiment by hacking Phpactor to index `.phar` files.

dantleech:
ok, it's basically either a Zip or Tar file. Experimenting locally I do not get anything immediately parseable (using the `phpactor.phar`) so I think it would need to be untarred/unzipped and then indexed. Apparently PHPStorm will happily index PHAR files.

CalebDW:
Technically the `phar` can be one of three types (however, the phar extension can convert between the formats very easily):

> Phar archives can be distributed as tar archives, zip archives, or as the custom phar file format designed specifically for the phar extension.

I believe PHPStorm does allow 'jump to definition' and the other code actions...not sure how though.

The solution for other folks has just been to extract the phar manually which would allow their lsp to index the files:
https://github.com/phpstan/phpstan/issues/2972#issuecomment-585143076

dantleech:
This is probably doable, but needs thought.

dantleech:
_should__ work #2602

calebdw:
Awesome!! It looks like it's working, thank you! I'm getting auto-completion and class/method docs


I am getting this error occasionally though:

[...]

PR #2602, opened 2024-03-16
GH-2412: Support for indexing and analyzing phars

dantleech:
Indexing and analysing PHARs is prerty simple we just need to ensure we preserve the `phar:///` scheme internally. The clients will not open `phar://` files however.

### DIFF
--- a/lib/Indexer/Tests/Unit/Adapter/Worse/IndexerClassSourceLocatorTest.php
+++ b/lib/Indexer/Tests/Unit/Adapter/Worse/IndexerClassSourceLocatorTest.php
@@ -7,6 +7,7 @@ use Phpactor\Indexer\Model\Record\ClassRecord;
 use Phpactor\TextDocument\ByteOffset;
 use Phpactor\Indexer\Adapter\Php\InMemory\InMemoryIndex;
 use Phpactor\Indexer\Adapter\Worse\IndexerClassSourceLocator;
+use Phpactor\TextDocument\TextDocumentUri;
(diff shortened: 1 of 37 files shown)


lib/Indexer/Tests/Unit/Adapter/Worse/IndexerClassSourceLocatorTest.php
```diff
@@ -52,6 +53,21 @@ public function testReturnsSourceCode(): void
...
+            ->setStart(ByteOffset::fromInt(0))
+            ->setEnd(ByteOffset::fromInt(10))
+            ->setFilePath('phar://' . __FILE__);
+
+        $index = new InMemoryIndex();
+        $index->write($record);
+        $locator = $this->createLocator($index);
+        $sourceCode = $locator->locate(Name::fromString('Foobar'));
+        $this->assertEquals(Path::canonicalize(__FILE__), $sourceCode->uri()?->path());
+    }
```
dantleech, line 69: is this testing anything?

lib/TextDocument/Exception/TextDocumentNotFound.php
```diff
@@ -12,6 +12,6 @@ public static function fromUri(TextDocumentUri $uri): self
         return new self(sprintf(
             'Text document "%s" not found',
             $uri
-        ));
+        ).(new \Exception())->getTraceAsString());
```
dantleech, line 15: revert

PR #2605, opened 2024-03-17
Remove deprecated composer 1 methods

mamazu:
* Explicitly documenting composer 2 as a dependency (as we already use the `InstalledVersions` class in the project somewhere else
* Replacing `PackageVersions\Version` with the not deprecated version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ Requirements
 ------------
 
 - PHP 8.1+
-- [Composer](https://getcomposer.org/)
+- [Composer 2](https://getcomposer.org/)
 
 Project Recommendations
 -----------------------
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -30,7 +30,7 @@ Requirements
 ------------
 
 - PHP 8.1+
-- [Composer](https://getcomposer.org/)
+- [Composer 2](https://getcomposer.org/)
```
dantleech, line 33: i don't think we need to specify `2`, you need composer. nobody should be using v1.

mamazu: Agreed. Removed.

dantleech:
:bow:

PR #2606, opened 2024-03-17
Adding more types in the FileFinder

mamazu:
Todo maybe: This class does a lot of copying with file path arrays maybe it would make more sense to refactor this to either append the arrays or to copy only once with the spread operator.

Either:
```php
$filePaths = [];
$this->parentFilePaths($reflection, $filePaths);
$this->interfaceFilePaths($reflection, $filePaths);

return FileList::fromFilePaths($filePaths);
````
or: 
```php
return FileList::fromFilePaths([
    ...$filePaths,
    ...$this->parentFilePaths($reflection),
    ...$this->interfaceFilePaths($reflection),
]);
````

### DIFF
--- a/lib/Extension/ClassMover/Application/Finder/FileFinder.php
+++ b/lib/Extension/ClassMover/Application/Finder/FileFinder.php
@@ -57,7 +57,7 @@ class FileFinder
         return $this->pathsFromReflectionClass($reflection, $private);
     }
 
-    private function pathsFromReflectionClass(ReflectionClass $reflection, bool $private)
+    private function pathsFromReflectionClass(ReflectionClass $reflection, bool $private): FileList
     {
         $path = $reflection->sourceCode()->uri()?->path();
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: lib/Extension/ClassMover/Application/Finder/FileFinder.php [hunk 1]
```diff
@@ -78,35 +80,68 @@ private function pathsFromReflectionClass(ReflectionClass $reflection, bool $pri
...
-        $filePaths = $filesystem->fileList()->existing()->phpFiles();
-        return $filePaths;
+        return $filesystem->fileList()->existing()->phpFiles();
     }
 
-    private function parentFilePaths(ReflectionClass $reflection, $filePaths)
+    /**
+     * @param array<string> $filePaths
+     *
+     * @return array<string>
```
dantleech, line 91: this was `array<string|null>`? is this definitely not handled? we could also skip if it's NULL for whatever reason.

mamazu: Yeah seeing that this never crashes it seems like it's never null but yeah, skipping those would probably be the better option.

dantleech:
thanks!

Issue #2612, opened 2024-03-21
Implement contracts inside anonymous classes


przepompownia:
For example in outdated https://github.com/phpactor/language-server/blob/c46e883649806613e649957b264c7e751bb66e20/bin/proxy#L48 after changing to 

```php
$formatter = new class() implements MessageFormatter {

};
```
we could expect `Implement contracts` code action.

mamazu:
On it.

mamazu:
Ref #2613

PR #2613, opened 2024-03-22
First try at implementing it

mamazu:


lib/WorseReflection/Core/Inference/Resolver/ObjectCreationExpressionResolver.php
```diff
@@ -27,12 +28,13 @@ public function __construct(private GenericMapResolver $resolver)
     public function resolve(NodeContextResolver $resolver, Frame $frame, Node $node): NodeContext
     {
         assert($node instanceof ObjectCreationExpression);
-        if (false === $node->classTypeDesignator instanceof Node) {
-            throw new CouldNotResolveNode(sprintf('Could not create object from "%s"', get_class($node)));
-        }
 
+        $className = $node->classTypeDesignator;
+        if (false === $className instanceof Node) {
```
dantleech, line 33: can this be explicit? in what case should it be an anonymous class?

lib/WorseReflection/Core/Reflection/Collection/ClassLikeReflectionMemberCollection.php
```diff
@@ -68,9 +69,11 @@ final class ClassLikeReflectionMemberCollection extends AbstractReflectionCollec
 
     public static function fromClassMemberDeclarations(
         ServiceLocator $serviceLocator,
-        ClassDeclaration $class,
+        ClassDeclaration|ObjectCreationExpression $class,
         ReflectionClass $reflectionClass
     ): self {
+        assert($class->classMembers !== null, 'ObjectCreationExpression does not contain anonymous class');
```
dantleech, line 75: what could it contain? should be an exception if it should never happen (and it probably can) or the error needs to be handled without crashing (maybe via. `NotFound` exception)

lib/WorseReflection/Core/Inference/Resolver/ObjectCreationExpressionResolver.php
mamazu: I don't think so. From what I have seen is that in the case of an anonymous class it is a Token pointing to something. (When I checked it it was something inside the anon class :shrug: ) As I see it there is no good way to detect an anonymous class. But that's a parser issue.

lib/WorseReflection/Core/Reflection/Collection/ClassLikeReflectionMemberCollection.php
mamazu: Yes, should probably assert that this is a `ClassMembers` variable. This happens when the code comes across an ObjectCreationExpression that does not contain an anonymous class.

lib/WorseReflection/Core/Reflection/Collection/ReflectionClassLikeCollection.php
```diff
@@ -55,6 +56,12 @@ public static function fromNode(ServiceLocator $serviceLocator, TextDocument $so
 
             if ($child instanceof ClassDeclaration) {
                 $items[(string) $child->getNamespacedName()] = new ReflectionClass($serviceLocator, $source, $child, $visited);
+                continue;
+            }
+
+            if ($child instanceof ObjectCreationExpression) {
+                // TODO: come up with a good way to generate a class name
```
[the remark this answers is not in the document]
dantleech: PHPStan does it like this: https://github.com/dantleech/phpstan-src/blob/f7b89d47cfa1fa636fb571b4315841a9af0e976f/src/Broker/AnonymousClassNameHelper.php#L22-L25

but I don't think this is the correct place to do it.

We can create the reflection class in the resolver?

lib/WorseReflection/Core/Inference/Resolver/ObjectCreationExpressionResolver.php
```diff
@@ -74,4 +76,20 @@ private function resolveClassType(NodeContextResolver $resolver, Frame $frame, O
...
+
+    private function resolveAnonymousClass(NodeContextResolver $resolver, Node $node): NodeContext
+    {
+        return NodeContextFactory::create(
+            'testing',
+            $node->getStartPosition(),
+            $node->getEndPosition(),
+            [
+                'symbol_type' => Symbol::CLASS_,
+                'type' => TypeFactory::fromStringWithReflector(
```
dantleech, line 88: why `fromString`?

lib/WorseReflection/Core/Inference/Resolver/ObjectCreationExpressionResolver.php
dantleech: so at least if it's a Token?

mamazu: Yes. But I don't know if that's always the case.

lib/WorseReflection/Core/Reflection/Collection/ReflectionClassLikeCollection.php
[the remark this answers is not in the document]
mamazu: Yeah, technically using the line number is not 100% garanteed to be unique but who has two anon classes on the same line. :D

But what you mean with "creating the reflection class in the resolver"? I wanted to try to avoid making a `ReflectionAnonymousClass` class as anon classes should be the same except for their weird name.

lib/WorseReflection/Core/Inference/Resolver/ObjectCreationExpressionResolver.php
mamazu: Copied that over from how the class definition resolver handles it.

lib/WorseReflection/Core/Inference/Resolver/ObjectCreationExpressionResolver.php
dantleech: if it's the case so far then let's go with that :)

lib/WorseReflection/Core/Inference/Resolver/ObjectCreationExpressionResolver.php
dantleech: ok, well TypeFactory::reflectedClass` or whatever but this is where the `ReflectionClass` should be created I think, not in teh collection

lib/WorseReflection/Core/Inference/Resolver/ObjectCreationExpressionResolver.php
mamazu: Replaced it with a check for `Token`

lib/WorseReflection/Core/Inference/Resolver/ObjectCreationExpressionResolver.php
mamazu: Yes using `Typefactory::reflectedClass` is better.

przepompownia:
From #2612 perspective this seems to be one step ahead but:

image

mamazu:
What were you trying to do? Generate a class?

przepompownia:
"Implement contracts" like in #2612

lib/WorseReflection/Core/Reflection/Collection/ReflectionClassLikeCollection.php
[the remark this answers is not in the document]
dantleech: I don't think it needs refactoring and would also fix the issue Thomas raised below, but would need to show an example.

mamazu:
Now it should work (at least the test for it is green) and it looks like the rest broke.

przepompownia:
Works for me:

image

lib/CodeBuilder/Adapter/TolerantParser/Updater/ClassMethodUpdater.php
```diff
@@ -13,19 +14,20 @@
...
-    * @return ClassMembersNode|TraitMembers|EnumMembers
-    */
-    public function memberDeclarationsNode(ClassLike $classNode)
+    public function memberDeclarationsNode(ClassLike|ObjectCreationExpression $classNode): ClassMembersNode|TraitMembers|EnumMembers
     {
-        if ($classNode instanceof ClassDeclaration) {
-            return $classNode->classMembers;
+        if ($classNode instanceof ClassDeclaration || $classNode instanceof ObjectCreationExpression) {
+            $classNode = $classNode->classMembers;
+            Assert::isInstanceOf( $classNode, ClassMembersNode::class);
```
dantleech, line 28: is it expected that this is NOT a ClassMembersNode? this will cause a runtime exception, which is probably not what we want.

if it should never happen then this should be an exception and I'd prefer not to use Assertion libraries.

lib/WorseReflection/Core/Util/NodeUtil.php
```diff
@@ -96,13 +96,16 @@ public static function nodeContainerClassLikeDeclaration(Node $node): ?Node
...
     /**
-     * @param Token|Node|mixed $nodeOrToken
+     * @param Token|Node|ObjectCreationExpression|mixed $nodeOrToken
      */
     public static function nameFromTokenOrNode(Node $node, $nodeOrToken): string
     {
         if ($nodeOrToken instanceof Token) {
             return (string)$nodeOrToken->getText($node->getFileContents());
         }
+        if ($nodeOrToken instanceof ObjectCreationExpression) {
```
dantleech, line 106: this is method is explicitly about getting the `name` frm a Node or Token, it's not about annonymous classes. you can add a separate method or probably there is something better.

lib/WorseReflection/Bridge/TolerantParser/Reflection/ReflectionClass.php
```diff
@@ -249,9 +255,12 @@ public function traits(): ReflectionTraitCollection
 
     public function memberListPosition(): ByteOffsetRange
     {
+        $classMembers = $this->node->classMembers;
+        assert($classMembers instanceof ClassMembersNode, 'ObjectCreationExpression does not contain anonymous class');
```
dantleech, line 259: presumably thid can happen and this will crash?

lib/WorseReflection/Bridge/TolerantParser/Reflection/TraitImport/TraitImports.php
```diff
@@ -103,8 +106,12 @@ private function __construct(array $declarations)
         }
     }
 
-    public static function forClassDeclaration(ClassDeclaration $classDeclaration): self
+    public static function forClassDeclaration(ClassDeclaration|ObjectCreationExpression $classDeclaration): self
     {
+        if (!$classDeclaration->classMembers instanceof ClassMembersNode) {
+            throw new NotFound('ObjectCreationExpression does not contain anonymous class');
```
dantleech, line 112: why not just default to an empty array?

lib/WorseReflection/Core/Inference/Resolver/ObjectCreationExpressionResolver.php
```diff
@@ -74,4 +77,20 @@ private function resolveClassType(NodeContextResolver $resolver, Frame $frame, O
         );
         return new GenericClassType($resolver->reflector(), $classType->name(), $templateMap->toArguments());
     }
+
+    private function resolveAnonymousClass(NodeContextResolver $resolver, Node $node): NodeContext
+    {
+        return NodeContextFactory::create(
+            'testing',
```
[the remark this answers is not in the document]
dantleech: it's an arbitrary identifier it can be the anonymous class "name"

lib/WorseReflection/Core/Inference/Resolver/ObjectCreationExpressionResolver.php
```diff
@@ -74,4 +78,17 @@ private function resolveClassType(NodeContextResolver $resolver, Frame $frame, O
...
+            $node->getStartPosition(),
+            $node->getEndPosition(),
+            [
+                'symbol_type' => Symbol::CLASS_,
+                'type' => TypeFactory::reflectedClass($resolver->reflector(), NodeUtil::nameFromTokenOrNode($node, $node))
```
dantleech, line 90: again, i'm not sure how this works, and assume it works based on the anonymous class previously being parsed? we should avoid depending on the cache.

see https://github.com/Roave/BetterReflection/pull/300

probably the reflector should support accepting the `Node` instead of a "name".

lib/WorseReflection/Core/Reflection/Collection/ClassLikeReflectionMemberCollection.php
```diff
@@ -68,9 +71,13 @@ final class ClassLikeReflectionMemberCollection extends AbstractReflectionCollec
 
     public static function fromClassMemberDeclarations(
         ServiceLocator $serviceLocator,
-        ClassDeclaration $class,
+        ClassDeclaration|ObjectCreationExpression $class,
         ReflectionClass $reflectionClass
     ): self {
+        if (!$class->classMembers instanceof ClassMembersNode) {
+            throw new NotFound('ObjectCreationExpression does not contain anonymous class');
```
dantleech, line 78: can default to empty array?

lib/WorseReflection/Core/Reflection/Collection/ReflectionClassLikeCollection.php
```diff
@@ -55,6 +57,11 @@ public static function fromNode(ServiceLocator $serviceLocator, TextDocument $so
 
             if ($child instanceof ClassDeclaration) {
                 $items[(string) $child->getNamespacedName()] = new ReflectionClass($serviceLocator, $source, $child, $visited);
+                continue;
+            }
+
+            if ($child instanceof ObjectCreationExpression && !($child->classTypeDesignator instanceof Node)) {
```
dantleech, line 63: this may cause unexpected issues. "reflect classes in" reflects locatable symbols and also it _should_ not need to deeply traverse the AST.

we could consider removing this and instead adding support to `$reflector->reflect{Node,AnnonymousClass,Object}($node)` for example.

lib/CodeBuilder/Adapter/TolerantParser/Updater/ClassMethodUpdater.php
mamazu: Yeah, the problem in this case is that it could also be a "normal" Object creation with a class name which would then crash here. But in this case I'm not sure what the best way to go about this is.

lib/WorseReflection/Bridge/TolerantParser/Reflection/ReflectionClass.php
```diff
@@ -57,13 +59,17 @@ class ReflectionClass extends AbstractReflectionClass implements CoreReflectionC
...
-        private ClassDeclaration $node,
-        private array $visited = []
+        private ClassDeclaration|ObjectCreationExpression $node,
+        private array $visited = [],
     ) {
     }
 
     public function isAbstract(): bool
     {
+        if ($this->node instanceof ObjectCreationExpression) {
```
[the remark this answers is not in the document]
mamazu: Yes, that's a good idea. I've done that.

lib/WorseReflection/Bridge/TolerantParser/Reflection/ReflectionMethod.php
```diff
@@ -60,7 +62,11 @@ public function nameRange(): ByteOffsetRange
 
     public function declaringClass(): ReflectionClassLike
     {
-        $classDeclaration = $this->node->getFirstAncestor(ClassLike::class);
+        $classDeclaration = $this->node->getFirstAncestor(ClassLike::class, ObjectCreationExpression::class);
+        if ($classDeclaration instanceof ObjectCreationExpression) {
+            return $this->class ?? $this->serviceLocator->reflector()
+                ->reflectClassLike(NodeUtil::nameFromTokenOrNode($classDeclaration, $classDeclaration));
```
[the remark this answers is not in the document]
mamazu: I don't know how the reflector does it. But I assume it cashes the classes that it has seen yes. Where would the logic be for finding it properly? Do we need to index anonymous classes for that?

lib/WorseReflection/Core/Reflection/Collection/ReflectionClassLikeCollection.php
mamazu: The problem is this approach won't find all classes anyways. Because we actually have to traverse into class definitions as well. Otherwise we would miss anonymous classes inside other classes. But I don't have an idea how to do improve performance here.

dantleech: yes, they need probably to be interpreted inline as types as with the above comment. doing this will mean that "implement contracts" will not automatically work, but at the same time it's only working because of the cache effect in this PR

mamazu: I just saw that `reflectClassesIn` should probably better be named `reflectClassLikeIn` 'cause it also returns interfaces enums etc. which are completely irrelevant for the "Implement contracts". But that's not a problem of this PR.

However I get what you're saying, I'm not sure if its based on cache in this situation because the source code builder reflects the same structure again and should also build the same list of `ClassLike` nodes. On the other hand I'm quite familiar with the concept of inline types and how I would go about and implement this. Do we already have an example for this?

przepompownia:
At the moment I have no occasion to work with anonymous classes, but have merged the recent version of this PR into my experimental branch and can report possible side effect.

dantleech:
just to be clear, it's unlikely to be merged with it's current implementation. but hopefully I can find some time to look more into it.

przepompownia:
Regardless of merging into master, this change stopped working after merge  de10e2220ebe5676f2f2396a9303830c8ebdbd1f here (incompatible declarartion of `reflectClassesIn` in lib/WorseReflection/Bridge/TolerantParser/Reflector/TolerantSourceCodeReflector.php).

mamazu:
Should work again now.

przepompownia:
Works again, thank you

mamazu:
I had to rebase this because I messed up the merging of this branch.

PR #2614, opened 2024-03-24
Adding better types

mamazu:
Using phpactor to generate return types for phpactor code.

What is in here:
* Specifying types and reducing the amount of "fromUnknown" calls since now we know what types there are.
* Removing deprecated `isClass` and `isInterface` calls from the code and replaced them with instanceof calls.
* Moving one of the tests from array returns to Generator (and include the tests).

### DIFF
--- a/lib/Filesystem/Domain/FilePath.php
+++ b/lib/Filesystem/Domain/FilePath.php
@@ -48,7 +48,7 @@ final class FilePath
         return self::fromString((string) $fileInfo);
     }
 
-    public static function fromUnknown(FilePath|string $path): FilePath
+    public static function fromFilePathOrString(FilePath|string $path): FilePath
     {
         if ($path instanceof FilePath) {
             return $path;
--- a/lib/Filesystem/Domain/FilePath.php
+++ b/lib/Filesystem/Domain/FilePath.php
@@ -29,6 +29,9 @@ final class FilePath
         return new self($textDocumentUri);
     }
 
+    /**
+     * @param array<string> $parts
+     */
     public static function fromParts(array $parts): FilePath
     {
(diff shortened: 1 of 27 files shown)


lib/Filesystem/Domain/MappedFilesystemRegistry.php
```diff
@@ -35,7 +35,6 @@ public function has(string $name): bool
         return isset($this->filesystems[$name]);
     }
 
-    /** @return array<string> */
```
mamazu, line 38: Moved to the interface.

lib/Filesystem/Domain/FilePath.php
```diff
@@ -54,19 +57,6 @@ public static function fromUnknown($path): FilePath
...
-        }
-
-        if ($path instanceof SplFileInfo) {
-            return self::fromSplFileInfo($path);
-        }
-
-        throw new RuntimeException(sprintf(
-            'Do not know how to create FilePath from "%s"',
-            is_object($path) ? get_class($path) : gettype($path)
-        ));
```
mamazu, line 69: Those were only used in tests. The rest of the code already uses the correct function.

dantleech: they were used by tests though, removing this changes the purpose of the function (`fromUnknown`)

dantleech: can rename to `fromFilePathOrString` for clarity

mamazu: If possible I would like to remove those kinds of methods entirely and reduce the amount of "unknown"s from phpactor. :D

dantleech:
thanks

Issue #2618, opened 2024-03-27
`$this` incorrectly detected inside object creation expression


przepompownia:
![image](https://github.com/phpactor/phpactor/assets/11404453/20383f72-fc84-404a-93d9-399a0d1cca01)

The case
```php
        return new Success((function () {
            $classes = $this;
        })());
```
exists for example in `lib/CodeTransform/Adapter/WorseReflection/Transformer/ImplementContracts.php`. 

It's a regression after #2469 @mamazu 

Perhaps similar caution will be needed in #2613

mamazu:
Could you maybe come up with a test case for that and I'll have a look at it.

przepompownia:
https://github.com/phpactor/phpactor/pull/2619

przepompownia:
I also looked into `FunctionLikeWalker` but don't have idea how to distinguish whether created object belongs to named or anonymous class.

mamazu:
The distinguishing feature is that the `typeDescrimitator` of the `ObjectCreationExpression` is a Token and not a `Node`. That's what we are using in the other PR.

mamazu:
I've created a fix pr: @2620

I still hold the opinion that the Parser should return a more coherent structure but oh well.

przepompownia:
Fixed by #2620

PR #2620, opened 2024-03-28
`$this` type inside object creation

mamazu:
That's an easy fix. If we're inside an `ObjectCreationExpression` and it has a name, then just continue looking up the tree.

### DIFF
--- a/lib/WorseReflection/Core/Inference/Walker/FunctionLikeWalker.php
+++ b/lib/WorseReflection/Core/Inference/Walker/FunctionLikeWalker.php
@@ -76,6 +76,17 @@ class FunctionLikeWalker implements Walker
             ObjectCreationExpression::class, // For Inline classes
         );
 
+        while ($classNode instanceof ObjectCreationExpression && $classNode->classTypeDesignator instanceof Node) {
+            // If we are here we found a normal ObjectCreationExpression like: new A(); and this is not useful and we continue traversing
+            $classNode = $classNode->getFirstAncestor(
+                ClassDeclaration::class,
+                InterfaceDeclaration::class,
+                TraitDeclaration::class,
+                EnumDeclaration::class,
+                ObjectCreationExpression::class, // For Inline classes
+            );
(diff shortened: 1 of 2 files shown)


przepompownia:
Seems to work inside `ImplementContracts` :tada: 

btw. on that class this is not the last issue - it's a good implicit test case :wink:

lib/WorseReflection/Core/Inference/Walker/FunctionLikeWalker.php
```diff
@@ -76,6 +76,17 @@ private function walkFunctionLike(FrameResolver $resolver, Frame $frame, Functio
             ObjectCreationExpression::class, // For Inline classes
         );
 
+        while ($classNode instanceof ObjectCreationExpression && $classNode->classTypeDesignator instanceof Node) {
```
dantleech, line 79: while?

lib/WorseReflection/Tests/Integration/Core/Inference/FrameWalker/FunctionLikeWalkerTest.php
```diff
@@ -34,6 +34,29 @@ public function hello()
             $this->assertEquals(false, $frame->locals()->byName('this')->first()->isProperty());
         }];
 
+        yield 'It returns this with correct type' => [
```
dantleech, line 37: ... in an anonymous function

dantleech:
is this a regression or has it _never_ worked?

dantleech:
In which MR was the regression introduced?

mamazu:
https://github.com/phpactor/phpactor/pull/2469

przepompownia:
> is this a regression or has it _never_ worked?

It's a regression, as I mentioned in https://github.com/phpactor/phpactor/issues/2618#issue-2211939576 (this PR should contain `Fixes #2618` in description).

przepompownia:
Thank you

PR #2622, opened 2024-03-29
KeywordCompletor: complete `return` and `yield`

przepompownia:


przepompownia:
I initially wanted to add more keywords (`abstract` and `final` too) here but they seem to need a separate attention, while this PR can already be reviewed and possibly merged. 

In other words, let do it separately.

lib/Completion/Bridge/TolerantParser/CompletionContext.php
```diff
@@ -264,6 +264,11 @@ public static function anonymousUse(Node $node): bool
 
     public static function methodName(Node $node): bool
     {
+        // TODO better distinguish empty body from empty method parameter list
```
dantleech, line 267: can we remove TODO and instead document what this condition is for? TODO will never get done.

dantleech:
in general I'm nervous about adding these completions as the mechanism for determining if they should be provided or not is not good vs. the effort of typing a few well known characters.

image

image

probably more instances where the results are not as wanted.

przepompownia:
> in general I'm nervous about adding these completions as the mechanism for determining if they should be provided or not is not good vs. the effort of typing a few well known characters.
> 
> probably more instances where the keywords are either not provided or provided in the wrong place.

I will try to look at these cases again. 

I share your annoyance with such cases, but on the other hand, I'm even more bothered by the lack of additions in expected places. Habits of using lua-language-server raise my expectations here too.

przepompownia:
At the moment I reproduced only the second case (in real use) - the first does not work after a few guesses.

lib/Completion/Bridge/TolerantParser/CompletionContext.php
przepompownia: Done :wink:

przepompownia:
image
This condition checks nodes on truncated document (up to cursor).

przepompownia:
https://github.com/phpactor/phpactor/blob/b890cfbaec6996ee98696c2bfe4a5e0cdc790401/lib/Completion/Bridge/TolerantParser/CompletionContext.php#L208-L210

I noticed that commenting the above condition does not affect any test result. Where it could be needed?

przepompownia:
image
```
dap> $nodeBeforeOffset === $node->parent->parent
true
```
Isn't it a bug?

przepompownia:
Now I think we need a separate method in `CompletionContext` to determine whether the current node is in CompoundStatement within some method or function declaration.

przepompownia:
Now it seems to be ready to review again. 

Some keywords like `break`, `continue` needs additional conditions. Maybe I'll do it here, as well as it can be done later.

`instanceof` seems to be more separate case. It follows expression and does not start any statement.

przepompownia:
> in general I'm nervous about adding these completions as the mechanism for determining if they should be provided or not is not good vs. the effort of typing a few well known characters.
> 
> image
> 
> image
> 
> probably more instances where the keywords are either not provided or provided in the wrong place.

I tried to set better priority for keywords:

https://github.com/phpactor/phpactor/pull/2622/files#diff-2ba75b9069ec2e0eca8655a5fa214ecb0050bf8656e6c1842d2be83b91d31f27R136

przepompownia:
Experience with this PR shows that there can still exist lots of cases to consider to be statement context or not. Since implementing the new method in `CompletionContext` adding any next case became simple - especially not mixed with other contexts.

IMHO we could already merge this PR as is and possibly support new cases in the future. Before possible merge please report any annoying cases that I haven't found yet.

przepompownia:
In practice I merge this and other PRs to one experimental branch and see what happens (independently of tests). For example I found

image
image
image
image

przepompownia:
Fixed case like below:

Screenshot From 2025-01-02 20-34-02

mamazu:
The wrong suggestions also happen when there is no invalid syntax. Like in this fully defined match:
image

przepompownia:
Yes, I caught it too and still have no time to focus on it.

przepompownia:
Your example shows allowed behavior like
```php
match(match(1) {1 => 1}) {1 => 1}
```
I meant rather an example like
image

dantleech:
Again, I am nervous about that PR - it adds lots of conditionals to make up for the fact that the parser breaks in strange ways. It would need constant maintainence  and will likely never behave properly. The last one I merged introduced false positives  which I'd like to revert (`const` and `match` IIRC). If I merge it I will need to test it and I'm unlikely to find the time to do that.

I'd rather have no suggestions than wrong ones.

przepompownia:
Feel free to close this PR if you think so (and no one lets us know they need it). I have some use for it and I want to continue this experiment even on a branch detached from any PR.

przepompownia:
Test failure: #3007

PR #2623, opened 2024-03-29
Making the diagnostics looks good

mamazu:
Now they're also properly displayed in Neovim with a strike through.

image


lib/Extension/LanguageServerWorseReflection/DiagnosticProvider/WorseDiagnosticProvider.php
```diff
@@ -26,15 +29,22 @@ public function provideDiagnostics(TextDocumentItem $textDocument, CancellationT
...
                 $lspDiagnostic = ProtocolFactory::diagnostic($range, $diagnostic->message());
                 $lspDiagnostic->severity = self::toLspSeverity($diagnostic->severity());
                 $lspDiagnostic->source = 'phpactor';
-                $lspDiagnostics[] = $lspDiagnostic;
 
+                if ($diagnostic instanceof DeprecatedUsageDiagnostic) {
+                    $lspDiagnostic->tags[] = DiagnosticTag::DEPRECATED;
+                }
+
+                if ($diagnostic instanceof UnusedImportDiagnostic) {
```
dantleech, line 43: these "tags" should rather be added to the Diagnostic class, there can be many sources of "unnecessary" code

mamazu:
Maybe an interface would have been better that way you don't need to change the implementation of Diagnostics that don't return anything. But both are fine to me.

dantleech:
thanks

PR #2627, opened 2024-03-31
Fix renaming attributed class members

### LINKED ISSUE #2577: Cannot rename property or method if its definition contains any attribute
```php
<?php

namespace Arctgx\PhpactorAttrCompletion;

use Override;

class Details
{
    #[Override]
    private $foo;

    #[ReturnTypeWillChange]
    public function getFoo()
    {
        return $this->foo;
    }
}
```
Trying to rename `$foo` in the above example I get `LSP Message (phpactor) Could not find corresponding reference to member name "foo" in document "file:///home/arctgx/dev/arctgx/php/phpactor-attr-completion/src/Details.php" at offset 84`. The same exception I get when trying to rename method with attribute. 

https://github.com/phpactor/phpactor/blob/5da647cd525f04c41f2ded6678f44844c58114c8/lib/Rename/Adapter/ReferenceFinder/AbstractReferenceRenamer.php#L81-L89

The example is simplified as much as possible. In practice it was ORM column attribute.

przepompownia:
Fix #2577

### DIFF
--- a/lib/Rename/Tests/Cases/member_renamer/constant_declaration_public/ClassOne.ph
+++ b/lib/Rename/Tests/Cases/member_renamer/constant_declaration_public/ClassOne.ph
@@ -1,6 +1,11 @@
 <?php
 
+use JetBrains\PhpStorm\Deprecated;
+
 class ClassOne
 {
     public const FOO = 'bar';
+
+    #[Deprecated]
+    public const MOO = 'bar', ZOO = 'bar';
 }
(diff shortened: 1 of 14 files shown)


dantleech:
is this a parser bug though? and if not does the same apply to properties?

przepompownia:
It seems to be ready to review for class members.

dantleech (approved):

dantleech:
thanks!

przepompownia:
Thank you especially for [the hint](https://github.com/phpactor/phpactor/issues/2577#issuecomment-2027025787) - it significantly reduced investigating.

PR #2629, opened 2024-04-03
Index attribute flags

### LINKED ISSUE #2199: Attribute completion: restrict result list by target, if present

przepompownia:
Fix #2199

### DIFF
--- a/lib/Indexer/Adapter/Php/Serialized/FileRepository.php
+++ b/lib/Indexer/Adapter/Php/Serialized/FileRepository.php
@@ -17,7 +17,7 @@ class FileRepository
     /**
      * Increment this number each time there is a B/C break in the index.
      */
-    private const VERSION = 2;
+    private const VERSION = 3;
 
     /**
      * Flush to the filesystem after BATCH_SIZE updates
--- a/lib/Indexer/Adapter/Tolerant/Indexer/ClassDeclarationIndexer.php
+++ b/lib/Indexer/Adapter/Tolerant/Indexer/ClassDeclarationIndexer.php
@@ -5,6 +5,9 @@ namespace Phpactor\Indexer\Adapter\Tolerant\Indexer;
 use Microsoft\PhpParser\MissingToken;
 use Microsoft\PhpParser\Node;
 use Microsoft\PhpParser\Node\Attribute;
+use Microsoft\PhpParser\Node\Expression\ArgumentExpression;
+use Microsoft\PhpParser\Node\Expression\BinaryExpression;
+use Microsoft\PhpParser\Node\Expression\ScopedPropertyAccessExpression;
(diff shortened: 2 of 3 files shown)


przepompownia:
Any idea how to better read target names from an attribute? At the moment it digs into `Attribute` node to find `ScopedPropertyAccessExpression` nodes and get their text representations.

lib/Completion/Bridge/TolerantParser/ChainTolerantCompletor.php
```diff
@@ -29,11 +29,18 @@ public function __construct(private array $tolerantCompletors, Parser $parser =
...
+        $node = $this->parser
+            ->parseSourceFile(substr_replace((string) $source, ' ', $truncatedLength, 0))
+            ->getDescendantNodeAtPosition(
+                // the parser requires the byte offset, not the char offset
+                $truncatedLength,
+            );
+
+        if ($node->getEndPosition() > $truncatedLength) {
+            $node = $this->parser->parseSourceFile($truncatedSource)->getDescendantNodeAtPosition($truncatedLength);
+        }
```
przepompownia, line 43: @dantleech @mamazu  It's hacky and probably worth a better replacement, but passes all current tests. 

- on truncated source cannot determine attribute context (whether it's attached to class, method, etc.) because the relevant information is after the cursor
- without truncation: the data set `does not use variables declared after offset a` in `WorseParameterCompletorTest` (and not only) does not pass. 
- inserting additional space after the cursor before parsing does not break any existing behavior but seems to help to get the correct node if there are no white space after the cursor.

przepompownia: I didn't found any issue related to this change, using this branch in normal work.

lib/Completion/Bridge/TolerantParser/ChainTolerantCompletor.php
```diff
@@ -29,11 +29,18 @@ public function __construct(private array $tolerantCompletors, ?Parser $parser =
...
```
dantleech: what does this achieve?

dantleech:
let's merge it after the release I'll make today then I'll be able to evaluate it.

dantleech:
:boom:

przepompownia:
> 💥

In other words, let's see what else can explode :stuck_out_tongue: Thanks!

PR #2642, opened 2024-04-18
Renaming the reflector method

mamazu:
Renaming: SourceCodeReflector::reflectClassesIn => SourceCodeReflector::reflectClassLikesIn

The old method name would imply that it only returns classes which it doesn't to so renaming makes the intention clearer and opens up the possibility to create a function which only returns classes as well.

### DIFF
--- a/lib/CodeBuilder/Adapter/WorseReflection/WorseBuilderFactory.php
+++ b/lib/CodeBuilder/Adapter/WorseReflection/WorseBuilderFactory.php
@@ -37,7 +37,7 @@ class WorseBuilderFactory implements BuilderFactory
                 ->build();
         }
 
-        $classes = $this->reflector->reflectClassesIn($source);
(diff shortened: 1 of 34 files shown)


dantleech:
ir's not intended to add a function to return classes, as you get a collection and can filter it. 

The PHP API doesn't have the concept of ClassLike. `ReflectionClass` is for all "class likes", `class_exists` is for all class likes, `::class` is for all class-likes ....

mamazu:
Fair enough, I thought that it could have performance benefits if we could be more precise in specifying what we want. Like for example in the "Implement contracts" doesn't need to reflect on anything other than classes.

dantleech:
It's lazy, so extremely marginal benefit as you always need to parse the entire AST :)

mamazu:
True, but then I'll close this.

PR #2645, opened 2024-04-22
LanguageServerCodeTransformExtension: allow query class new variants

przepompownia:


przepompownia:
I didn't found any similar method on the LSP side. If such one exist, please close this PR.

Up until now, I was manually creating a list of variants for each project and had overlooked that Phpactor provides more generators than it did about four years ago.

image

dantleech:
There's a code action for creating new classes? What's the use case here?

przepompownia:
No, there is LSP command `create_class` used in the client-side example.

przepompownia:
https://github.com/phpactor/phpactor/blob/b890cfbaec6996ee98696c2bfe4a5e0cdc790401/lib/Extension/LanguageServerCodeTransform/LspCommand/CreateClassCommand.php#L24-L26

dantleech:
there _is_ a code action to create classes. create an empty file and invoke code actions on it. You will get to select from the variants

przepompownia:
I found `CreateClassProvider`. Is that what you had in mind? It turns out that I use both ways to create new class depending on the situation. 

Stubbornly, I probably can change the client code to create empty buffer and call the mentioned code action from it (maybe the code will become much simpler) but what do we need `CreateClassCommand` for?

przepompownia:
I see: `CreateClassCommand` is simply used by `CreateClassProvider`.

przepompownia:
Probably my direct call to `create_class` comes from some documentation snippet that is no longer available but I cannot find any trace of it. It seems that there is no sense to add off-spec query to support the way that can duplicate some built-in client behavior.

przepompownia:
In practice the code action is not available just after creating document at the client side. I tried to defer execution in the client side ways (autocommands in Neovim) but it seems that I need some information from the server about such availability.

przepompownia:
Deferring the call after `textDocument/didOpen` client notification seems to be enough for me.

https://github.com/przepompownia/nvim-arctgx/blob/1837299e9863b5b24360df601087de6e088ad6e7/lua/arctgx/lsp/serverConfigs/phpactor.lua#L3-L21

Issue #2332, opened 2023-08-06
Issue in VSCode: DiagnosticsParser.php, Could not decode expected PHPStan JSON string


ckilb:
New installation of PHPActor in VSCode breaks with:

```
[Info  - 14:06:32] Connection to server got closed. Server will restart.
Starting language server, use -vvv for verbose output

In DiagnosticsParser.php line 46:
                                                    
  Could not decode expected PHPStan JSON string ""  
```

Any idea how to fix this?

dantleech:
Is this a regression with today's release?

ckilb:
Closing this.
PHPActor by default takes PHPStan from the project dir which - in my case - includes a binary for another platform.

dantleech:
ok, it probably shouldn't crash however

leo-unglaub:
I just ran into the same issue while beeing on 1869936f5bd01515ddc8ada0928e9458dfd9a6e1. This causes

```
Starting language server, use -vvv for verbose output
In DiagnosticsParser.php line 46:
                                                    
  Could not decode expected PHPStan JSON string ""  
                                                    
language-server [--address ADDRESS] [--no-loop]

[Info  - 15:33:44] Connection to server got closed. Server will restart.
[Error - 15:33:44] Request textDocument/codeAction failed.
```

This causes the server to crash and then it triggers phpactor/phpactor/issues/2358.

dantleech:
Whatever the cause it shouldn't crash the server

scippio:
I have same error: https://github.com/larastan/larastan/issues/1978

mamazu:
This has been broken for some time now (I just never bothered fixing it). The fix I suggested only deals with empty outputs. Should we also just quietly fail with an error message and then continue in case phpstan sends invalid JSON?

scippio:
The interesting thing is, when I run it manually exactly like phpactor on some file I get valid JSON:
`
/usr/bin/php /home/scippio/.../vendor/bin/phpstan analyse --no-progress --error-format=json /tmp/phpstanlspXyM1S
`
``` json
{
    "totals": {
        "errors": 0,
        "file_errors": 13
    },
    "files": {
        "/tmp/phpstanlspXyM1S": {
            "errors": 13,
            "messages": [
                {
                    "message": "Attribute class Vyuldashev\\LaravelOpenApi\\Attributes\\PathItem does not exist.",
                    "line": 30,
                    "ignorable": true,
                    "identifier": "attribute.notFound"
                },
[...]

PR #2681, opened 2024-06-24
Skip empty phpstan outputs

mamazu:
If phpstan produces an empty stdout we don't need to parse it.

refs #2332


dantleech:
why does phpstan return empty?

mamazu:
The code seems to allow all exit codes less than or equal to 1. Maybe we have missed an edge case where phpstan doesn't produce any output.

dantleech:
thanks!

PR #2694, opened 2024-07-14
Adding support for explanatory text in throw

mamazu:
Before when a method had this signiture the help text after the throw was not visible in the hover overlay:
image

Now:
image


dantleech:
Nice

PR #2695, opened 2024-07-15
Completion context: improve recognizing expressions outside class member body

przepompownia:
Problem: cannot complete expressions for attribute parameters value when the attribute is attached to some non-promoted property, class or outside class.

The current version does not fix cases when the completed word has zero length.


dantleech:
Not sure what the impact will be, but merging to find out

przepompownia:
Let's look at the example before (no suggestions):
image
and now:
image

przepompownia:
I need to remind myself (maybe by looking into still open PR's) how to suggest before typing in such contexts.

PR #2701, opened 2024-07-20
fix autocompletion after php tag

### LINKED ISSUE #2689: Auto completion doesn't suggest the basics on the first try
Hey here,
I am trying to set up phpactor with monaco-editor but having some issues regarding the auto-completions.
CleanShot 2024-07-08 at 21 04 52
If you see from the video, it doesn't suggest more auto completions when the file is empty. but as soon as I enter something, even a character like `1` or the first line, it starts working fine. and then again if I make the file empty, it doesn't suggest more.

saeedvaziry:
fixes #2689


dantleech:
thanks :)

PR #2702, opened 2024-07-21
Imporve formatting for override method

dantleech:
- Do not show FQN (use short name)
- Sort alphabetically
- Distinguish between static and instance methods


mamazu:
Could we also move those to the end of the quickfix list? Because in some cases this has replaced quickfixes like "import missing class" or "remove unused class" from the first spot.

Issue #2730, opened 2024-09-02
False positive for missing method param


nsrosenqvist:
This code snippet here results in `Method "array" is missing @param $input` even though it is specified in the PHPDoc. It likely has something to do with the definition syntax that PHPStan supports.

```php
declare(strict_types=1);

namespace App;

class Asd
{
    /**
     * @param array{0: 'w', 1: 'o', 2: 'r', 3: 'l', 4: 'd'} $input
     *
     * @return 'world'
     */
    public function hello(array $input): string
    {
        return implode('', $input);
    }
}
```

Possibly related to https://github.com/phpactor/phpactor/issues/2264

dantleech:
probably a docblock parser issue

nsrosenqvist:
> probably a docblock parser issue

I just submitted a PR based on the changes you made previously in https://github.com/phpactor/phpactor/issues/2264 that passes the tests, but saw now that you've recently added a commit as well. Let me know if you want me to make any changes: https://github.com/phpactor/phpactor/pull/2732

dantleech:
oh, thanks for the PR - unfortunate timinng as I randomly felt like fixing it and merged before seeing your PR. It fixes your specific issue but perhaps not the test that you added.

nsrosenqvist:
@dantleech I'm just grateful that you responded to the issue so quickly :) I'll see if any of the changes can provide any value to the test suite at least. I'll close the PR and pull the latest changes

nsrosenqvist:
@dantleech Added now a PR with just the part of the test suite that your commit didn't include. Hope that it can provide some value

ADmad:
I am using the latest `2024.06.30.0` release and I see errors like this for every argument typed as `array` 
```
phpactor: Method "fooMethod" is missing @param $bar
```
I expect now error to be shown since the `array` type is already specified.

dantleech:
@ADmad that warning is telling you to specify the types of the array, `array` is implicitly `array<mixed>` which is almost always not what you want. e.g. list of strings is `list<string>`, or a disctionary mapping strings to strings is `array<string,string`.

You can now disable that warning if you really want to though as of https://github.com/phpactor/phpactor/pull/2781

ADmad:
@dantleech The error message needs be improved here as it gives no indication that it's expecting more detailed typing for the array.

I am aware of the array type declaration but there are occasions where it's not really necessary or needs to be put of for a later date. So the ability to disable it is welcome.

[...]

PR #2732, opened 2024-09-02
Support array shapes with string literals in param annotations

nsrosenqvist:
Fixes https://github.com/phpactor/phpactor/issues/2730


nsrosenqvist:
Literally fixed at the same time by @dantleech

PR #2733, opened 2024-09-02
Add diagnostics for array shape string literals

nsrosenqvist:
Will squash if accepted so that only the Diagnostics changes are reflected in the merge. Adds to https://github.com/phpactor/phpactor/issues/2730


dantleech:
all good, thanks :)

PR #2746, opened 2024-09-13
Moving the `RecordReferenceEnhancer`

mamazu:
Moving the `RecordReferenceEnhancer (interface)` next to the implementation (NullEnhancer):

```php
//From
use Phpactor\Indexer\Model\RecordReferenceEnhancer;
// To
use Phpactor\Indexer\Model\RecordReferenceEnhancer\RecordReferenceEnhancer;
```

Or was that supposed to be all the interfaces in the `Model` directory? And the folders are just for the implementations?


dantleech:
Yes folders are for implementations :)

mamazu:
Yeah, I sometimes do pr's without thinking it through. :D

Issue #2749, opened 2024-09-30
PHP-CS-Fixer fails to start with non-default bindir


jtojnar:
I am using Phpactor 2024.06.30.0 as LSP in Sublime Text and get an error about PHP-CS-Fixer failing on startup:

```
Exception [Phpactor\Extension\LanguageServerPhpCsFixer\Exception\PhpCsFixerError] php-cs-fixer exited with code '1'; cmd: '/nix/store/9bfaprpy0vid9pc9z2a0jdlzp93vrq97-php-with-extensions-8.2.24/bin/php' '/home/jtojnar/Projects/simplepie/vendor/bin/php-cs-fixer' 'fix' '--diff' '--dry-run' '-'; stderr: ''; stdout: 'Could not open input file: /home/jtojnar/Projects/simplepie/vendor/bin/php-cs-fixer' at /nix/store/a8vk7c5q8srixv4lc4l79ajy151qmxsc-phpactor-2024.06.30.0/share/php/phpactor/lib/Extension/LanguageServerPhpCsFixer/Model/PhpCsFixerProcess.php#57", 'data': "#0 [internal function]: Phpactor\Extension\LanguageServerPhpCsFixer\Model\PhpCsFixerProcess->Phpactor\Extension\LanguageServerPhpCsFixer\Model\{closure}()
#1 /nix/store/a8vk7c5q8srixv4lc4l79ajy151qmxsc-phpactor-2024.06.30.0/share/php/phpactor/vendor/amphp/amp/lib/Coroutine.php(118): Generator->send(1)
#2 /nix/store/a8vk7c5q8srixv4lc4l79ajy151qmxsc-phpactor-2024.06.30.0/share/php/phpactor/vendor/amphp/amp/lib/Internal/ResolutionQueue.php(70): Amp\Coroutine->Amp\{closure}(NULL, 1)
#3 /nix/store/a8vk7c5q8srixv4lc4l79ajy151qmxsc-phpactor-2024.06.30.0/share/php/phpactor/vendor/amphp/amp/lib/Internal/Placeholder.php(149): Amp\Internal\ResolutionQueue->__invoke(NULL, 1)
#4 /nix/store/a8vk7c5q8srixv4lc4l79ajy151qmxsc-phpactor-2024.06.30.0/share/php/phpactor/vendor/amphp/amp/lib/Deferred.php(53): Amp\Promise@anonymous->resolve(1)
[...]

PR #2753, opened 2024-10-03
Adding `composer_bin_dir` path expander

mamazu:
However, when I run phpactor locally with this enabled it doesn't load anymore. It's stuck in some kind of cycle of allocating stuff (at least that's what I can tell from the `strace`).

And when trying to debug it with xdebug it freezes on a `proc_close` call.

Closes https://github.com/phpactor/phpactor/issues/2749

### DIFF
--- a/lib/ComposerInspector/ComposerInspector.php
+++ b/lib/ComposerInspector/ComposerInspector.php
@@ -43,6 +43,11 @@ final class ComposerInspector
             return;
         }
 
+        /** @var array{
+         *   packages?: array<array{name: string, version: string}>,
+         *   "packages-dev"?: array<array{name: string, version: string}>
+         * } $lockContent
+         */
         $lockContent = $this->parseFile($this->lockFile);
         foreach ($lockContent['packages'] ?? [] as $pkg) {
             $this->packages[(string)$pkg['name']] = $this->forVersion($pkg['name'], $pkg['version'], false);
@@ -51,6 +56,7 @@ final class ComposerInspector
             $this->packages[(string)$pkg['name']] = $this->forVersion($pkg['name'], $pkg['version'], true);
         }
 
+        /** @var array{"bin-dir"?:string} $composerContent */
         $composerContent = $this->parseFile($this->composerFile);
         $this->vendorBinDir = $composerContent['bin-dir'] ?? self::DEFAULT_BIN_DIR;
 
@@ -60,7 +66,7 @@ final class ComposerInspector
     /** @return array<string, mixed> */
     private function parseFile(string $fileName): array
     {
-        $contents = file_get_contents($fileName);
+        $contents = @file_get_contents($fileName);
         if (false === $contents) {
             return [];
         }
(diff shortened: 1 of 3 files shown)


mamazu:
Found the problem, you can't have a service tagged as a value expander and depend a service that resolves paths. But I don't know how to solve that.

lib/ComposerInspector/ComposerInspector.php
```diff
@@ -7,57 +7,73 @@
...
+        $this->readFiles();
         if (!isset($this->packages[$name])) {
             return null;
         }
 
         return $this->packages[$name];
     }
 
-    private function readFile(): void
+    public function getBinDir(): string
```
dantleech, line 37: can rename to `binDir`

lib/ComposerInspector/ComposerInspector.php
```diff
@@ -7,57 +7,73 @@
...
+    {
+        $this->readFiles();
+        return $this->binDir;
+    }
+
+    private function readFiles(): void
     {
-        if ($this->loaded) {
+        if ($this->loaded || !file_exists($this->lockFile) || !file_exists($this->composerFile)) {
+            $this->loaded = true;
```
dantleech, line 46: both files are now optional...

dantleech:
you can use a lazy/callback expander as follows:

```diff
diff --git a/lib/Extension/ComposerInspector/ComposerInspectorExtension.php b/lib/Extension/ComposerInspector/ComposerInspectorExtension.php
index f289fa3e..3abef5ea 100644
--- a/lib/Extension/ComposerInspector/ComposerInspectorExtension.php
+++ b/lib/Extension/ComposerInspector/ComposerInspectorExtension.php
@@ -7,6 +7,7 @@ use Phpactor\Container\Container;
 use Phpactor\Container\ContainerBuilder;
 use Phpactor\Container\Extension;
 use Phpactor\Extension\FilePathResolver\FilePathResolverExtension;
+use Phpactor\FilePathResolver\Expander\CallbackExpander;
 use Phpactor\FilePathResolver\Expander\ValueExpander;
 use Phpactor\FilePathResolver\PathResolver;
 use Phpactor\MapResolver\Resolver;
@@ -24,8 +25,10 @@ class ComposerInspectorExtension implements Extension
         });
 
         $container->register('composer.bin_path_expander', function (Container $container) {
-            $path = $container->get(ComposerInspector::class)->getBinDir();
-            return new ValueExpander('composer_bin_dir', $path);
+            return new CallbackExpander('composer_bin_dir', function () use ($container) {
+                $path = $container->get(ComposerInspector::class)->getBinDir();
+                return $path;
+            });
         }, [ FilePathResolverExtension::TAG_EXPANDER => [] ]);
     }
```

lib/ComposerInspector/ComposerInspector.php
mamazu: Yes, before the composer file was optional and the lock file should also be optional, because you don't have to install the dependencies to use this feature.

dantleech: I mean you could have a JSON file but not a lock file or vice versa potentially but this requires thatbyiu have both. We should allow it to work with only a JSON file

dantleech:
thanks!

PR #2756, opened 2024-10-08
Making `CallbackExpander` types explict as doctype

mamazu:
The benefits are:
* Phpstan is yelling at you if you use the wrong return type of your closure at the place you miss used it (preventing runtime errors (and the documentation shows it).
* We don't need to check the type if we make the file type safe then php will crash with a type error
* Less code

### DIFF
--- a/lib/FilePathResolver/Tests/Unit/Expander/CallbackExpanderTest.php
+++ b/lib/FilePathResolver/Tests/Unit/Expander/CallbackExpanderTest.php
@@ -4,7 +4,6 @@ namespace Phpactor\FilePathResolver\Tests\Unit\Expander;
 
 use Phpactor\FilePathResolver\Expander;
 use Phpactor\FilePathResolver\Expander\CallbackExpander;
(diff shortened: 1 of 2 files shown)


lib/FilePathResolver/Tests/Unit/Expander/CallbackExpanderTest.php
```diff
@@ -19,14 +18,4 @@ public function testExpandsCallbackValue(): void
     {
         $this->assertEquals('bar', $this->expand('%foo%'));
     }
-
-    public function testThrowsExeceptionWhenCallbackReturnsNonString(): void
```
mamazu, line 23: We could replace this test with a `TypeError` exception. But at this point we're testing php works and phpstan is not going to like this.

dantleech:
Phpunit is failing...

mamazu:
Looks like the "Upload to historian" part doesn't work. In the master it uses some secret api that I don't have access to.

Issue #2706, opened 2024-07-26
Generic wildcard not supported


dantleech:
![image](https://github.com/user-attachments/assets/0039b68c-f975-4481-925c-1e53d678aede)

dantleech:
at this point it may be worth re-evaluating the PHPStan docblock parser, the original reason for writing an own parser was due to real-time performance, but as I recall the difference may now be negligible.

bart-jaskulski:
Looked up quick setup from `phpstan/phpdoc-parser` with **very** basic benchmark based on existing code and it seems promising enough to incorporate PHPStan parser into Phpactor.

```
with PHP version 8.3.6, xdebug ✔, opcache ❌

\Phpactor\DocblockParser\Tests\Benchmark\PhpactorParserBench

    benchParse..............................I32 - Mo0.046ms (±2.22%)
    benchAssert.............................I9 - Mo15.323ms (±1.88%)

\Phpactor\DocblockParser\Tests\Benchmark\PhpstanParserBench

    benchParse..............................I32 - Mo0.032ms (±3.53%)
    benchAssert.............................I9 - Mo7.105ms (±1.59%)
```

This should be a good point for issues relating missing docblock implementations (like #2641), as the heavy work would be offloaded to PHPStan maintainers.

<details>
<summary>benchmark class</summary>

```php
<?php

namespace Phpactor\DocblockParser\Tests\Benchmark;

use PHPStan\PhpDocParser\Ast\PhpDoc\ParamTagValueNode;
use PHPStan\PhpDocParser\Ast\PhpDoc\PhpDocNode;
use PHPStan\PhpDocParser\Ast\Type\IdentifierTypeNode;
use PHPStan\PhpDocParser\Lexer\Lexer;
use PHPStan\PhpDocParser\Parser\ConstExprParser;
use PHPStan\PhpDocParser\Parser\PhpDocParser;
use PHPStan\PhpDocParser\Parser\TokenIterator;
use PHPStan\PhpDocParser\Parser\TypeParser;

class PhpstanParserBench extends AbstractParserBenchCase
{
    private $parser;

    private $lexer;

    public function setUp(): void
    {
        $this->lexer = new Lexer();
        $constExprParser = new ConstExprParser();
        $typeParser = new TypeParser($constExprParser);
        $this->parser = new PhpDocParser($typeParser, $constExprParser);
    }

    public function parse(string $doc): void
    {
[...]

PR #2759, opened 2024-10-11
gh-2706: Switch to PHPStan docblock parser

bart-jaskulski:
This is a draft for incorporating PHPStan docblock parser with the premise of better stability and performance.


lib/WorseReflection/Bridge/Phpactor/DocblockParser/PHPStanDocblockParserFactory.php
```diff
@@ -0,0 +1,42 @@
...
+{
+    private PhpDocParser $parser;
+
+    public function __construct(
+        private Reflector $reflector,
+        private Lexer $lexer = new Lexer(),
+    ) {
+        $constExprParser = new ConstExprParser();
+        $typeParser = new TypeParser($constExprParser);
+        $this->parser = new PhpDocParser($typeParser, $constExprParser);
```
bart-jaskulski, line 26: This ofc should be moved to constructor

lib/WorseReflection/Bridge/Phpactor/DocblockParser/PHPStanParsedDocblock.php
```diff
@@ -0,0 +1,321 @@
...
+    public function isDefined(): bool
+    {
+        return true;
+    }
+
+    public function properties(ReflectionClassLike $declaringClass): CoreReflectionPropertyCollection
+    {
+        $properties = [];
+        // merge read/write virtual properties
+        foreach ($this->node->getPropertyTagValues() as $propertyTag) {
```
bart-jaskulski, line 185: Here only `@property` tag is listed, without `@property-read` and `@property-write`, but I'm not sure if Phpactor has support for those

lib/WorseReflection/Bridge/Phpactor/DocblockParser/PHPStanTypeConverter.php
```diff
@@ -0,0 +1,259 @@
...
+            $type instanceof ArrayShapeNode => $this->convertArrayShape($type),
+            $type instanceof ArrayTypeNode => new ArrayType($this->convert($type->type)),
+            $type instanceof CallableTypeNode => $this->convertCallable($type),
+            $type instanceof ConditionalTypeForParameterNode => $this->convertConditional($type),
+            $type instanceof ConstTypeNode => $this->convertConstant($type),
+            $type instanceof GenericTypeNode => $this->convertGeneric($type),
+            $type instanceof IdentifierTypeNode => $this->convertSimpleIdentifier($type->name),
+            $type instanceof IntersectionTypeNode => $this->convertIntersection($type),
+            $type instanceof NullableTypeNode => new NullableType($this->convert($type->type)),
+            // todo: ObjectShape
```
bart-jaskulski, line 86: I'm not sure if it's ok to create this along this PR, but IMHO it would be easy to implement some additional types, which are missing now, like object shape or `non-empty-string`, etc. following https://phpstan.org/writing-php-code/phpdoc-types

lib/WorseReflection/Bridge/Phpactor/DocblockParser/PHPStanTypeConverter.php
```diff
@@ -0,0 +1,259 @@
...
+    private function convertCallable(CallableTypeNode $callableNode): Type
+    {
+        $parameters = array_map(
+            fn (CallableTypeParameterNode $type) => $this->convert($type->type),
+            $callableNode->parameters
+        );
+
+        $type = $this->convert($callableNode->returnType);
+
+        if (((string) $callableNode->identifier) === 'Closure') {
```
bart-jaskulski, line 219: This may be unnecessary, would require a test case

lib/WorseReflection/Tests/Unit/Bridge/Phpactor/DocblockParser/PHPStanDocblockParserFactoryTest.php
```diff
@@ -0,0 +1,442 @@
...
+use Phpactor\WorseReflection\Core\Type\StringType;
+use Phpactor\WorseReflection\Core\Type\UnionType;
+use Phpactor\WorseReflection\Core\Type\VoidType;
+use Phpactor\WorseReflection\Reflector;
+use Phpactor\WorseReflection\ReflectorBuilder;
+use Phpactor\WorseReflection\Tests\Integration\IntegrationTestCase;
+use Prophecy\Argument;
+use Prophecy\PhpUnit\ProphecyTrait;
+
+class PHPStanDocblockParserFactoryTest extends IntegrationTestCase
```
bart-jaskulski, line 39: This is nearly a raw copy of `DocblockParserFactoryTest`

lib/WorseReflection/Tests/Unit/Bridge/Phpactor/DocblockParser/PHPStanDocblockParserFactoryTest.php
```diff
@@ -0,0 +1,442 @@
...
+            new StringType()
+        ];
+
+        yield [
+            '/** @return int */',
+            new IntType()
+        ];
+
+        yield [
+            '/** @return integer */',
```
bart-jaskulski, line 75: This is not handled by original `DockblockParserFactory`

lib/WorseReflection/Tests/Unit/Bridge/Phpactor/DocblockParser/PHPStanDocblockParserFactoryTest.php
```diff
@@ -0,0 +1,442 @@
...
+        ];
+
+        yield 'arrayshape with keys' => [
+            '/** @return array{foo:int,bar:string} */',
+            'array{foo:int,bar:string}',
+        ];
+
+        yield 'parenthesized' => [
+            '/** @return null|(callable(int):string)|string|int */',
+            'null|(callable(int): string)|string|int',
```
bart-jaskulski, line 188: Parsing with PHPStan docparser returns middle callable without parenthesis. I don't know if it's acceptable

lib/WorseReflection/Bridge/Phpactor/DocblockParser/PHPStanTypeConverter.php
dantleech: yes, but I would aim for the minimum required for green tests before adding more stuff.

lib/WorseReflection/Tests/Unit/Bridge/Phpactor/DocblockParser/PHPStanDocblockParserFactoryTest.php
```diff
@@ -0,0 +1,442 @@
...
+        ];
+
+        yield 'list' => [
+            '/** @return list */',
+            TypeFactory::list(),
+        ];
+
+        yield 'list with type' => [
+            '/** @return list<string> */',
+            TypeFactory::list(TypeFactory::string()),
```
bart-jaskulski, line 216: Parsing with `DocblockParserFactory` returns `array<int,string>` from `list<string>` expression. With `PHPStanDocblockParser` it becomes `string[]`. Both are semantically correct, but I don't know if it's okay in case of Phpactor

lib/WorseReflection/Tests/Unit/Bridge/Phpactor/DocblockParser/PHPStanDocblockParserFactoryTest.php
```diff
@@ -0,0 +1,442 @@
...
+        ];
+
+        yield 'psalm prefix' => [
+            '/** @psalm-return int */',
+            TypeFactory::int(),
+        ];
+
+        yield 'conditional type' => [
+            '/** @return ($foo is true ? string : int) */',
+            TypeFactory::parenthesized(
```
bart-jaskulski, line 240: Again, PHPStan return it correctly, but without parenthesis

dantleech:
some before/after benchmarks:
- `./bin/phpactor worse:analyse lib/WorseReflection`
- `./vendor/bin/phpbench run lib/WorseReflection --variant=PhpUnitReflectClassBench` (maybe see if any othre benchmarks are sensitive to this change)

bart-jaskulski:
Besides `DocblockParserFactory`, there's also use of raw `Phpactor\DocblockParser\DocblockParser` in `Phpactor\CodeTransform\Adapter\DocblockParser\ParserDocblockUpdater`. This also might be replaced with phpstan docblock parser, but as here is crafting docblock comments for actions like _Add implements tags_, etc and it may be a bit more complex, as phpstan/docblock-parser has good support for printing whole docblock, but not so great for _manufacturing_ them.

What would be your take on that?

And additionally, what are the best ways to test the Phpactor after those changes, specifically, where this phpdoc parsing is used? Apart from phpunit tests, ofc

dantleech:
> This also might be replaced with phpstan docblock parser

we can leave that as is for now and focus on worse-reflection which is where most of the value comes in. We can then look at if/how we can replace DocblcokParser elsewhere (one of it's design principles was that it be "lossless" so you can reconsitute the docblock from the AST - but PHPStan may also fulfil that requirement now - but not sure).

the best way to test phpactor is to run the tests? or?

lib/WorseReflection/Bridge/Phpactor/DocblockParser/PHPStanDocblockParserFactory.php
```diff
@@ -0,0 +1,42 @@
...
+use Phpactor\WorseReflection\Core\DocBlock\DocBlockFactory;
+use Phpactor\WorseReflection\Core\Reflection\ReflectionScope;
+use Phpactor\WorseReflection\Reflector;
+use PHPStan\PhpDocParser\Lexer\Lexer;
+use PHPStan\PhpDocParser\Parser\ConstExprParser;
+use PHPStan\PhpDocParser\Parser\PhpDocParser;
+use PHPStan\PhpDocParser\Parser\TokenIterator;
+use PHPStan\PhpDocParser\Parser\TypeParser;
+
+class PHPStanDocblockParserFactory implements DocBlockFactory
```
dantleech, line 16: this would now be a _bridge_ to PHPStan, so the namespace for this code should be `...\Bridge\PHPStan\`

lib/WorseReflection/Bridge/Phpactor/DocblockParser/PHPStanDocblockParserFactory.php
```diff
@@ -0,0 +1,42 @@
...
+        $typeParser = new TypeParser($constExprParser);
+        $this->parser = new PhpDocParser($typeParser, $constExprParser);
+    }
+
+    public function create(string $docblock, ReflectionScope $scope): DocBlock
+    {
+        if (trim($docblock) === '') {
+            return new PlainDocblock();
+        }
+
```
dantleech, line 34: note that you _will_ need to filter for "supported tags" by regex before proceeding to parse as with the previous factory (maybe it can be done with a decorator). Otherwise we parse every single docblock whether we extract info from it or not.

bart-jaskulski:
I've been looking at benchmarks from CI:

commit from master: https://github.com/phpactor/phpactor/actions/runs/11244200976/job/31261739177#step:5:84
this PR: https://github.com/phpactor/phpactor/actions/runs/11289848501/job/31400507757?pr=2759#step:5:84

There's a slight improvement, but negligible, may be only a variance between tests

lib/WorseReflection/Bridge/Phpactor/DocblockParser/PHPStanDocblockParserFactory.php
bart-jaskulski: By supported you mean those, which has according `Type` classes in worse reflection? I've ditched this, because I thought those are tags supported by doc parser

dantleech:
yeah, i'd run them locally in anycase, you can get wildly different results in GL runners.

FWIW this was my original attempt to use phpstan from > 6 years ago:

https://github.com/phpactor/worse-reflection/pull/28

(at the time it was replacing a reg-based solution).

lib/WorseReflection/Bridge/Phpactor/DocblockParser/PHPStanDocblockParserFactory.php
dantleech: i mean:
```
    const SUPPORTED_TAGS = [
        'property',
        'var',
        'param',
        'return',
        'method',
        'type',
        'deprecated',
        'extends',
        'implements',
        'template',
        'template-covariant',
        'template-extends',
        'mixin',
        'throws',
        'assert',
    ];
```
are the types that are analysed, if there is a docblock that has none of those then we don't need to parse it.

lib/WorseReflection/Bridge/Phpactor/DocblockParser/PHPStanParsedDocblock.php
```diff
@@ -0,0 +1,321 @@
...
+        );
+    }
+
+    public function returnType(): Type
+    {
+        foreach ($this->node->getReturnTagValues() as $tag) {
+            return $this->convertType($tag->type);
+        }
+
+        foreach ($this->node->getReturnTagValues('@psalm-return') as $tag) {
```
bart-jaskulski, line 164: This should be more like https://github.com/phpstan/phpstan-src/blob/eed9282f97246c03ec4374549b7f3e1712c48f06/src/PhpDoc/PhpDocNodeResolver.php#L444 to actually support different prefixes

bart-jaskulski:
I did some benchmarking on my (low-end) laptop (intel core i7-1065G7, integrated graphics, 8GB ram), and then numbers this time lost its magic. Below is the median for both branches from 3 runs:

| **Benchmark**                                        | **master** | **gh-2706** | **Diff** |
|------------------------------------------------------|------------|-------------|----------|
| **PhpUnitReflectClassBench: test_case**              | 15.628     | 15.948      | 2.05%    |
| **PhpUnitReflectClassBench: methods_and_properties** | 54.07      | 55.30       | 2.27%    |
| **PhpUnitReflectClassBench: method_frames**          | 2091.34    | 2100.60     | 0.44%    |
| **SelfReflectClassBench: methods_and_properties**    | 0.77       | 0.81        | 5.19%    |
| **SelfReflectClassBench: frames**                    | 22.04      | 22.60       | 2.54%    |
| **YiiBench: benchMembers**                           | 92.720     | 96.856      | 4.46%    |
| **ReflectionStubsBench: test_classes_and_methods**   | 4.971      | 5.314       | 6.89%    |
| **ReflectPropertyBench: property**                   | 1.38       | 1.46        | 5.80%    |
| **ReflectPropertyBench: property_return_type**       | 2.62       | 2.69        | 2.67%    |
| **ReflectMethodBench: method**                       | 1.50       | 1.65        | 10.00%   |
| **ReflectMethodBench: method_return_type**           | 2.68       | 2.84        | 5.97%    |
| **ReflectMethodBench: inferred_return_type**         | 1.92       | 2.02        | 5.21%    |
| **DiagnosticsBench: lots_of_missing_methods**        | 61.176     | 64.378      | 5.23%    |
| **DiagnosticsBench: lots_of_new_generic**            | 23.622     | 24.610      | 4.18%    |
| **DiagnosticsBench: lots_of_new_objects**            | 21.250     | 21.360      | 0.52%    |
| **DiagnosticsBench: method_chain.test**              | 26.188     | 27.051      | 3.30%    |
| **DiagnosticsBench: phpstan.test**                   | 2010       | 2123        | 5.62%    |
| **AnalyserBench: benchAnalyse**                      | 59.801     | 65.816      | 10.06%   |

dantleech:
hm ok, well it's not the end of the world necessarily if we can make up the performance elsewhere (phpactor can be so much faster than it is), we can also make this a strategy and merge with it being disabled by default

lib/WorseReflection/Bridge/Phpactor/DocblockParser/PHPStanDocblockParserFactory.php
bart-jaskulski: moved to separate namespace, yet for now I left the (redundant) PHPStan prefix in class names (easier during testing and switching with original parser)

lib/WorseReflection/Tests/Unit/Bridge/Phpactor/DocblockParser/PHPStanDocblockParserFactoryTest.php
bart-jaskulski: ConditionalTypes are now wrapped in ParenthesizedType by default with PHPStan parser, but this seems fake to me

lib/WorseReflection/Bridge/Phpactor/DocblockParser/PHPStanDocblockParserFactory.php
bart-jaskulski: added

bart-jaskulski:
Please, review the [benchmark table](https://github.com/phpactor/phpactor/pull/2759#issuecomment-2407649243) again, as I have mistakenly benchmarked the branches without actual changes :facepalm:

I've edited the table to avoid confusion, and now it includes the values where tested code is **actually** using PHPStanDocblockParserFactory (benchmarks before still used original parser, what makes it even more bizzare, that results were slower).

Anyway, this again shows more potential.

Please, note that at current stage I've hardcoded PHPStan parser, yet closer to the actual merge, I will introduce strategy you mentioned (after solving failing test cases)

dantleech:
Yes, it's tricky :) in any case adding the filtering would have helped, but his looks encouraging

lib/WorseReflection/Bridge/PHPStan/DocblockParser/PHPStanDocblockParserFactory.php
```diff
@@ -0,0 +1,75 @@
...
+            return new PlainDocblock();
+        }
+
+        // if no supported tags in the docblock, do not parse it
+        if (0 === preg_match(
+            sprintf('{@((psalm|phpstan|phan)-)?(%s)}', implode('|', self::SUPPORTED_TAGS)),
+            $docblock,
+            $matches
+        )) {
+            return new PlainDocblock($docblock);
```
dantleech, line 60: we should extract this "filter" to e.g. `Phpactor\WorseReflection\Core\DocBloc` so that it's used by both implementations (assuming we will run them side-by-side via. a strategy initially)

bart-jaskulski: Good :+1: . I was thinking about the same, but didn't want to introduce that prematurely

bart-jaskulski:
I'd like to ask for help about specific cases and guidance in solving the issues.

1. https://github.com/phpactor/phpactor/actions/runs/11327207256/job/31497789474?pr=2759#step:5:400
PHPStan doesn't wrap types in parenthesizes. Is it acceptable?

2. https://github.com/phpactor/phpactor/actions/runs/11327207256/job/31497789474?pr=2759#step:5:394
Invalid virtual properties (missing type or property name) are ignored in PHPStan parser. Invalid in current implementation.

3. https://github.com/phpactor/phpactor/actions/runs/11327207256/job/31497789474?pr=2759#step:5:372
This test seems a bit odd to me -- how `(string|int)|int` is even valid typing? Doesn't it reduce to `string|int` even without any assertions?

4. https://github.com/phpactor/phpactor/actions/runs/11327207256/job/31497789474?pr=2759#step:5:326
https://github.com/phpactor/phpactor/actions/runs/11327207256/job/31497789474?pr=2759#step:5:348
I've got some issues about generics, but only in those two cases. TBH, I don't know what is the difference between native parser and phpstan implementation, but I surely miss something. Any guidance appreciated.

dantleech:
> PHPStan doesn't wrap types in parenthesizes. Is it acceptable?

yes

> Invalid virtual properties (missing type or property name) are ignored in PHPStan parser. Invalid in current implementation.

fine to ignore invalid ones i guess

> This test seems a bit odd to me -- how (string|int)|int is even valid typing? Doesn't it reduce to string|int even without any assertions?

it's fine if PHPStan reduces the type

> I've got some issues about generics, but only in those two cases. TBH, I don't know what is the difference between native parser and phpstan implementation, but I surely miss something. Any guidance appreciated.

```diff
diff --git a/lib/WorseReflection/Tests/Inference/generics/class_implements_multiple1.test b/lib/WorseReflection/Tests/Inference/generics/class_implements_multiple1.test
index e07e6e36..c8238f5d 100644
--- a/lib/WorseReflection/Tests/Inference/generics/class_implements_multiple1.test
+++ b/lib/WorseReflection/Tests/Inference/generics/class_implements_multiple1.test
@@ -23,7 +23,8 @@ interface Second {
 }
 
 /**
- * @implements Second<Baz>, First<Boo>
+ * @implements Second<Baz>
+ * @implements First<Boo>
  */
 class Foo implements Second, First
 {
```

for the second with `@psalm-assert` it looks like we're only looking for `@phpstan-assert`?

bart-jaskulski:
FYI, I'll get back to this PR next week

lib/WorseReflection/Bridge/PHPStan/DocblockParser/PHPStanDocblockParserFactory.php
bart-jaskulski: I've settled with decorator because I did not fully understand how could I handle this with just static method, but if you have any suggestion, please do.

lib/WorseReflection/Bridge/TolerantParser/Diagnostics/DocblockMissingExtendsTagProvider.php
```diff
@@ -138,7 +138,7 @@ class Foobar extends NeedGeneric
             assertion: function (Diagnostics $diagnostics): void {
                 Assert::assertCount(1, $diagnostics);
                 Assert::assertEquals(
-                    'Generic tag `@extends NeedGeneric` should be compatible with `@extends NeedGeneric<int>`',
+                    'Missing generic tag `@extends NeedGeneric<int>`',
```
bart-jaskulski, line 141: Another change between native and phpstan parser directly in test suite. As phpstan recognizes `@extends` tag without generic as invalid, the most simple solution is to annotate it with missing generic tag message

lib/WorseReflection/Bridge/TolerantParser/Diagnostics/UnusedImportProvider.php
```diff
@@ -462,18 +469,24 @@ public function name(): string
         return 'unused_import';
     }
 
-    private function extractDocblockNames(Docblock $docblock, NodeContextResolver $resolver, Node $node): void
+    private function extractDocblockNames(DocBlock $docblock, NodeContextResolver $resolver, Node $node): void
```
bart-jaskulski, line 472: This re-implementation of extracting docblock types relies on parsed types, instead of directly on docblock AST, yet it may pose issues with native parser

lib/WorseReflection/Core/DocBlock/DocBlock.php
```diff
@@ -11,6 +11,8 @@
 
 interface DocBlock
 {
+    /** public function types(): Types; */
```
bart-jaskulski, line 14: I wasn't sure if it's okay to introduce new method in interface, so I've put this as comment for now

lib/WorseReflection/Core/ServiceLocator.php
```diff
@@ -87,7 +89,14 @@ public function __construct(
         );
 
         $this->sourceLocator = $sourceLocator;
-        $this->docblockFactory = new DocblockParserFactory($this->reflector);
+
+        if (!true) {
```
bart-jaskulski, line 93: `ServiceLocator` creates parser factory directly in constructor, without access to global configuration (I was thinking of exposing chosen parser to config). I would appreciate any suggestions for making this strategy work. Maybe introduce strategy service, which would be responsible for creating actual factory?

lib/WorseReflection/Tests/Inference/generics/class_implements_multiple1.test
```diff
@@ -23,7 +23,8 @@ interface Second {
 }
 
 /**
- * @implements Second<Baz>, First<Boo>
+ * @implements Second<Baz>
+ * @implements First<Boo>
```
bart-jaskulski, line 27: Incompatible with native parser

lib/WorseReflection/Tests/Inference/type/parenthesized.test
```diff
@@ -3,10 +3,10 @@
 /** @var (string|int)|int */
 $foo;
 
-wrAssertType('(string|int)|int', $foo);
+wrAssertType('string|int', $foo);
```
bart-jaskulski, line 6: Incompatible with native parser

lib/WorseReflection/Tests/Integration/Bridge/TolerantParser/Reflection/ReflectionClassTest.php
```diff
@@ -1105,7 +1105,7 @@ class Class1
         ,
             'Class1',
             function (ReflectionClass $class): void {
-                $this->assertEquals(2, $class->properties()->count());
+                $this->assertEquals(0, $class->properties()->count());
```
bart-jaskulski, line 1108: Incompatible with native parser

lib/WorseReflection/Tests/Integration/Core/Inference/NodeContextResolverTest.php
```diff
@@ -1175,7 +1175,7 @@ private function resolveNodeAtOffset(
         $nameResolver = new NodeToTypeConverter($reflector, $this->logger());
         $resolver = new NodeContextResolver(
             $reflector,
-            new DocblockParserFactory($reflector),
+            new PHPStanDocblockParserFactory($reflector),
```
bart-jaskulski, line 1178: Parser used in tests is now hardcoded. I have no good idea, those parses could be iterated to execute test suite with both of them

lib/WorseReflection/Bridge/Phpactor/DocblockParser/PHPStanParsedDocblock.php
dantleech: Phpactor supports them AFAIK but only in the sense that it interprets them as `@property`

lib/WorseReflection/Core/ServiceLocator.php
dantleech: hmm, as we need the Reflector to construct it it will either have to be a `boolean` flag to the `ServiceLocator` or a factory so that the factory can be passed in.

The factory is the better choice architecturally, but tbh a boolean flag is also fine.

lib/WorseReflection/Tests/Integration/Core/Inference/NodeContextResolverTest.php
dantleech: This is the weaker set of tests, the `Inference/SelfTest.php` is much easier to manage and if it passes the chances are we're good. So this could remain with the original parser for now.

dantleech:
thanks for the work so far, I'll have a closer look at this soon and get a better perspective on how to get the build green.

lib/WorseReflection/Bridge/TolerantParser/Diagnostics/MissingReturnTypeProvider.php
```diff
@@ -144,11 +162,16 @@ public function exit(NodeContextResolver $resolver, Frame $frame, Node $node): i
             return;
         }
 
-        if ($method->docblock()->returnType()->isMixed()) {
+        $docblockReturn = $method->docblock()->returnType();
+        if ($docblockReturn->isMixed()) {
+            return;
+        }
+
+        if ($docblockReturn->isDefined()) {
```
bart-jaskulski, line 170: There are `MissingReturnTypeProvider` and `DocblockMissingReturnTypeProvider`. It seems, even on `master` branch declaring return type only in docblock leaves with false-positive diagnostic about missing return type. I've added this check, but it may be better to merge both classes -- despite the change I think it is out of the scope of this PR, but without this new test case is failing. Anyway, up to you if this should be split into own PR.

lib/WorseReflection/ReflectorBuilder.php
```diff
@@ -36,6 +36,8 @@ final class ReflectorBuilder
 
     private bool $enableContextualSourceLocation = false;
 
+    private bool $usePhpStanDocblock = true;
```
bart-jaskulski, line 39: Although I've moved phpstan parser flag away from `ServiceLocator`, it still is hardcoded, as `ReflectorBuilder` is always instantiated with static method, with no access to global config or so.

bart-jaskulski:
As `phpstan/phpdoc-parser` v2 just dropped, I've upgraded the library.

Besides, when using this branch on daily basis I've noticed two issues, which are now covered with test case and appropriate fix (missing return and unused import for array types).

lib/WorseReflection/ReflectorBuilder.php
bart-jaskulski: Nevermind. Added as config parameter inside `WorseReflectionExtension` for key `worse_reflection.phpdoc_parser` (options are "native" and "phpstan").

lib/WorseReflection/Bridge/Phpactor/DocblockParser/PHPStanParsedDocblock.php
bart-jaskulski: Added `@property-read` and `@property-write` to parsed property types

bart-jaskulski:
> Nice, so what is left from your perspective? I guess we still need to run the tests with both implementations (something I could look at).

Yeah, running tests for both implementation would be crucial -- any help appreciated. Besides coming up with some sort of consensus about discrepancies in tests, like rendered doc formatting or changes which are not reflected in native parser. I've annotated all such cases in this PR comments; it would also be obvious when running the test suite against the original parser.

> presumably we can fully replace the original parser with that option in the long run (althoug the API isn't as nice)

API for creating phpdoc tags in phpstan is indeed more _object-oriented_ (:boom:) and burdensome to actually craft anything, but it would be possible to make such replacement, following the example from their readme with `Printer`: https://github.com/phpstan/phpdoc-parser?tab=readme-ov-file#format-preserving-printer

PR #2760, opened 2024-10-12
PHP8,.4  Compat

dantleech:


lib/CodeTransform/Adapter/WorseReflection/Refactor/WorseReplaceQualifierWithImport.php
```diff
@@ -23,7 +23,7 @@ public function __construct(
         private Reflector $reflector,
         private BuilderFactory $factory,
         private Updater $updater,
-        Parser $parser = null
+        ?Parser $parser = null
```
mamazu, line 26: If we upgraded to php8.1 This would not have to be null and could be promoted to this:
```php
private Parser $parser = new Parser(),
```

Issue #2754, opened 2024-10-04
Phpactor not recognising standard php functions


bchubb-web:
I use the phpactor lsp inside of neovim, and codebase-wise it works perfectly, i much prefer it to intelephense, but recently it has stopped recognising all standard php functions, i experienced this with intelephense when i used stubs to add wordpress functions, but i have no special config in my lsp settings, does anyone know how to resolve this?

GosuTeacher:
I've taken a bit of time to check this out as it was bugging me also and found out that after the PHAR indexing the path scheme logic is changed and generates `file:///` scheme instead of `phar://` for the stubs. (When the project is loaded from phar, the phar is somewhere in the middle and the scheme remains as a file, the uri results in `file:///` and it can't be loaded)

From what I've checked, TextDocumentLocator works like this:
1. scheme is already set to one of the supported (phar, file) (not the case for the jetbrains stubs as they are set as canonical absolute path)
2. checks if the path ends with `phar`(extension) to set the scheme for the path  (not the case for the jetbrains stubs as they are set as canonical absolute path)

The second issue that impacts this is that the project is delivered in a `phar` file, if you unwrap the `phpactor.phar` and change the bash file that executes the app, or short-circuit it to bin/phpactor directly, this will work

So, we have a few solutions for this:
1. The easiest one and I think the one that's recommended until this issue is fixed is to change the `https://phpactor.readthedocs.io/en/master/reference/configuration.html#worse-reflection-stub-dir` stub dir config (either at the system/user/project level) like this: `https://phpactor.readthedocs.io/en/master/usage/configuration.html` either to the stubs delivered with the phpactor project, or you can install the `jetbrains/phpstorm-stubs` via composer and point to that location(I'll try to create a PoC for this today)
2. Change the scheme for the config in the app entry point based on how you interact with phpactor (via phar or bin/phpactor) - check if `$phpactorBin` is a phar archive or not. In my opinion this is the worse way to fix it
3. Change the way the scheme is parsed in `TextDocumentLocator`, but this may not be so reliable. e.g. check if the path contains the `.phar/` string and set the scheme as phar

Hope this helps. Good luck!

charjr:
Thank you @GosuTeacher for the clarification, and for the temporary solution!

This issue is still present in the latest release: [2024.11.05.0](https://github.com/phpactor/phpactor/releases/tag/2024.11.05.0)

dantleech:
[...]

PR #2774, opened 2024-11-08
gh-2754: Preserve PHAR scheme when indexing PHAR stubs

dantleech:


MaG21:
I'm really rooting for this PR, I'm going crazy with these false positives
image

Issue #2728, opened 2024-08-30
Unable to disable "Class name should probably be..."


3ynm:
In my project, migration files include date prefixes, causing them not to exactly match the class names. This triggers a warning: `Class name should probably be....`

I do not intend to change the class names to remove the prefixes, so I'd like an option to disable this warning. However, it seems there is no configuration available to do so.

If disabling the warning is not possible, I would prefer that the diagnostics to only highlight the specific offending line, rather than marking the entire class as problematic, as seen on the following image.

![annoying_diagnostics](https://github.com/user-attachments/assets/c2cfebe0-f8c8-4dbf-9828-3f7529cb8694)

mamazu:
I mean there is an option to disable the diagnostic provider for all transformer based diagnostics, but this will also disable way more than you want.
```json
"language_server.diagnostic_providers": [
   "code-action"
   "create-class",
   "implement-contracts"
],
```
as far as I know there is no way to disable this individual one.

3ynm:
I think the diagnostic would be fine if it didn't mark the whole class as offending

dantleech:
you should now be able to disable any diagnostic by it's displayed code in the LSP as of #2781

dantleech:
fwiw I also updated the diagnostic to only highlight the name: https://github.com/phpactor/phpactor/pull/2783?new_mergebox=false

PR #2777, opened 2024-11-15
Add ability to disable specific diagnostics

bart-jaskulski:
Minimal POC for #2728. Currently doesn't work with WorseReflection diagnostics (e.g. missing return).

### DIFF
--- a/lib/Extension/LanguageServer/LanguageServerExtension.php
+++ b/lib/Extension/LanguageServer/LanguageServerExtension.php
@@ -654,6 +654,7 @@ class LanguageServerExtension implements Extension
      */
     private function collectDiagnosticProviders(Container $container, ?bool $outsourced): array
     {
+        /** @var DiagnosticsProvider[] */
         $providers = [];
         foreach ($container->getServiceIdsForTag(self::TAG_DIAGNOSTICS_PROVIDER) as $serviceId => $attrs) {
             Assert::isArray($attrs, 'Attributes must be an array, got "%s"');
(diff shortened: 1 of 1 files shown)


bart-jaskulski:
With this, you can add to your phpactor config:

```json
{
  "language_server.disabled_diagnostics": [
    "fix_namespace_class_name"
  ]
}
```

bart-jaskulski:
I don't know if there's a reliable way to discover existing diagnostics identifiers right now

dantleech:
actually think I was rather hasty there.

this disables diagnostic _providers_ not diagnostics.

bart-jaskulski:
Well, I could marked it as draft, as I planned on adding tests and other providers to exclude (like mentioned, this did not work with WorseRef extension).

Tbh, there's no way to disable specific diagnostic, only it's provider. I assumed it was fine (it's how rules mechanism works e.g. in PhpCodeSniffer), but I can also refactor it to add identifiers for diagnostics and a way for filtering them.

Besides, for what I understand, currently there are two levels of diagnostics providers:
1. Diagnostic group, like `code-action`
2. Diagnostic provider, like namespace fix

And atm you can allow for specific groups to be enabled. As it is a broad scope, it would be nice too to use deny list for unwanted providers (or specific diagnostics).

Happy to further discuss it, as it may be my overengineering of related issue problem.

dantleech:
I've had a "quick" (heh, hours) go at adding diagnostic codes to all diagnostics:

https://github.com/phpactor/phpactor/pull/2781

I think this is the way forward as it's a pretty generic solution... needs a little bit of reflection and change in the language-server package.

PR #2778, opened 2024-11-15
fix: fix render of nullable type

jeremyFreeAgent:
I've noticed that strange behavior when running the `class:transform` cli on interfaces:

```
 interface Animal
 {
-    public function jump(): ?Foo;
+    public function jump(): ? ?Foo;
 }
 ```
 
 That PR should fix that.


mamazu:
Hey, thank you for the contribution. However it seems like the issue is the underlying logic that updates the methods. If you're curious feel free to check out the linked PR.

jeremyFreeAgent:
The problem is resolved. Thanks!

PR #2779, opened 2024-11-16
Fixing method updating when return type is nullable.

mamazu:
Replaces #2778

When updating a method's return type the `TextEdit` updates only the type list which is correct unless there is a nullable token at the beginning:

```
public function test() ?Testing;
                        ^     ^
```
This is the part that was being replaced. This PR will also include the question mark operator.

### DIFF
--- a/lib/CodeBuilder/Adapter/TolerantParser/Updater/AbstractMethodUpdater.php
+++ b/lib/CodeBuilder/Adapter/TolerantParser/Updater/AbstractMethodUpdater.php
@@ -219,21 +219,23 @@ abstract class AbstractMethodUpdater
             return;
         }
 
-        $returnType = (string) $this->renderer->render($returnType->type());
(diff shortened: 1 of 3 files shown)


lib/CodeBuilder/Adapter/TolerantParser/Updater/AbstractMethodUpdater.php
```diff
@@ -219,21 +219,23 @@ private function updateOrAddReturnType(Edits $edits, ReturnType $returnType, Met
             return;
         }
 
-        $returnType = (string) $this->renderer->render($returnType->type());
+        $returnType = trim((string) $this->renderer->render($returnType->type()));
+        if ($returnType === '') {
+            return;
+        }
```
mamazu, line 225: Do we also want to remove the return type when phpactor renders it as empty?

dantleech:
:bow:

jeremyFreeAgent:
Thanks!

PR #2781, opened 2024-11-16
Gh 2729: Diagnostic codes and disabling

dantleech:
gh-2781: show diagnostic codes and allow them to be ignored

- Show provider ID as code if no code present
- Add diagnostic codes to all WR diagnostics
- Always show a diagnostic code
- Remove redundant `name` method on diagnostic provider
- Add `language_server.diagnostic_ignore_cores` option.

image
image

### DIFF
--- a/lib/Extension/LanguageServer/DiagnosticProvider/AggregateDiagnosticsProvider.php
+++ b/lib/Extension/LanguageServer/DiagnosticProvider/AggregateDiagnosticsProvider.php
@@ -1,0 +1,86 @@
+<?php
+
+namespace Phpactor\Extension\LanguageServer\DiagnosticProvider;
+
+use Amp\CancellationToken;
+use Amp\Promise;
+use Phpactor\LanguageServerProtocol\Diagnostic;
+use Phpactor\LanguageServerProtocol\TextDocumentItem;
+use Phpactor\LanguageServer\Core\Diagnostics\DiagnosticsProvider;
+use Psr\Log\LoggerInterface;
+use function Amp\call;
+use Throwable;
+
+class AggregateDiagnosticsProvider implements DiagnosticsProvider
+{
+    /**
+     * @var array<DiagnosticsProvider>
+     */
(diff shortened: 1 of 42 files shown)


lib/Extension/LanguageServer/DiagnosticProvider/AggregateDiagnosticsProvider.php
```diff
@@ -0,0 +1,84 @@
...
+use Amp\CancellationToken;
+use Amp\Promise;
+use Phpactor\LanguageServerProtocol\Diagnostic;
+use Phpactor\LanguageServerProtocol\TextDocumentItem;
+use Phpactor\LanguageServer\Core\Diagnostics\DiagnosticsProvider;
+use Psr\Log\LoggerInterface;
+use function Amp\call;
+use Throwable;
+
+class AggregateDiagnosticsProvider implements DiagnosticsProvider
```
dantleech, line 14: overrideen this class from the language-server package, could port it back up

lib/Extension/LanguageServerWorseReflection/DiagnosticProvider/WorseDiagnosticProvider.php
```diff
@@ -94,4 +98,26 @@ private static function toLspTags(array $tags): array
             $tags
         );
     }
+
+    private function diagnosticToCode(Diagnostic $diagnostic): string
```
dantleech, line 102: instead of guessing the name we could set it explicitly. the reason I didn't was because of the existing `DiagnosticProvider#name` method - this method however could probably simply be removed.

mamazu: Does this also relate to this: #2074 ?

dantleech: Nope

PR #2786, opened 2024-11-20
using Enums for NamesearcherTypes

mamazu:


dantleech:
are we sure this has a neutral impact on memory and perforamnce?

mamazu:
I did some sandboxed profiling and it looks like the enums memory footprint is smaller (the longer the string, the more memory you save with enum). The same with comparing, comparing enums is faster than comparing strings.

But don't we have benchmarks for searching in the test suite.

<details>
  <summary>Code used for profiling</summary>
```php
      <?php

    const SIZE = 100_000;

    function generateRandomString($length = 5) {
        $characters = '0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ';
        $charactersLength = strlen($characters);
        $randomString = '00000';

        for ($i = 0; $i < $length; $i++) {
            $randomString[$i]= $characters[random_int(0, $charactersLength - 1)];
        }

        return $randomString;
    }

    enum Hello: string {
        case YYj1eiLFXX = 'YYj1e';
        case OcTFWw4eYW = 'ocTFW';
        case ZvBIAgeXlC = 'ZvBIA';
        case fRnKFkBc29 = 'fRnKF';
        case vmZGlm3bpv = 'vmZGl';
        case wcNXjjkiA9 = 'wcNXj';
        case EDUSZG2Iyi = '3DUSZ';
        case ZZTJjfyxty = '2ZTJj';
        case DhXexs0kGL = 'DhXex';
        case Wuvn5u4cV9 = 'Wuvn5';
    }

    function strings() {
        $elements = [];
        $before=memory_get_usage();
        for ($i = 0; $i < SIZE; $i++) {
            $elements[] = generateRandomString();
        }
        $after=memory_get_usage();
        return ($after - $before)/1000;
    }
    echo 'Strings:' .strings()." MB\n";

    function enums() {
        $cases = Hello::cases();
        $casesCount = count($cases);
        $e2 = [];
        $before=memory_get_usage();
        for ($i = 0; $i < SIZE; $i++) {

            $e2[] = $cases[rand(0, $casesCount-1)];
        }
        $after = memory_get_usage();

        return  ($after - $before)/1000;
    }
    echo 'Enums:' .enums()." MB\n";

    echo '=========================='.PHP_EOL;
    // --- Speed
    function stringSpeed() {
        $elements = [];
        $before=microtime(true);
        $dups = 0;
        for ($i = 0; $i < SIZE; $i++) {
            if(generateRandomString() === '2ZTJj'){
                $dups++;
            }
        }
        $after=microtime(true);
        echo "Duplicates ". $dups.PHP_EOL;

        return $after - $before;
    }
    echo 'Strings:' .stringSpeed()." seconds\n";

    function enumsSpeed() {
        $cases = Hello::cases();
        $casesCount = count($cases);
        $before=microtime(true);
        $dups = 0;
        for ($i = 0; $i < SIZE; $i++) {
            if ($cases[rand(0, $casesCount-1)] === Hello::wcNXjjkiA9) {
                $dups++;
            }
        }
        $after = microtime(true);
        echo "Duplicates ". $dups.PHP_EOL;

        return  $after - $before;
    }
    echo 'Enums:' .enumsSpeed()." seconds\n";
```
</details>

mamazu:
Did some more profiling and as expected serializing and deserializing from a string is more expensive with an enum but I don't know what the io vs comparision rate is in phpactor.

PR #2792, opened 2024-11-24
Add Windows CI steps

zobo:


dantleech:
Looks like we'd need to track down whatever is blocking forever

zobo:
Yes. It doesn't block locally. Will see what I can find. 

Also, a lot of tests fail due to different line endings. I can fix some, but an easy fix would be to use a line ending independent assertEquals. How do you feel about that?

zobo:
I've made it so that the windows job does not fail the workflow (as per [GitHub docs](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error)). I'll wait for https://github.com/phpactor/test-utils/pull/13 to be merged and then see how for we are. A lot of tests can pass if the work dir is checked out with LF instead of CRLF, but a bunch of tests just need more work... E.g.: phpstan seems to behave differently on Windows...

zobo:
@dantleech rebased ... let me know what you think about this patch. I can keep updating it for now and try to fix more tests, or we can merge it as is...

PR #2793, opened 2024-11-26
Error on method extraction with whitespace

bart-jaskulski:
Signed-off-by: Bart Jaskulski <bjaskulski@protonmail.com>


bart-jaskulski:
If there's a whitespace inside if braces included in selection, method extractions goes awry. Current output is:

```php
<?php

class Foo {

    public function bar() {
        $foo = $bar = 1;
        if ( $this->newMethod($foo, $bar);) {
            return true;
        }

        return false;
    }
    private function newMethod($foo, int $bar)
    {
        $foo && $bar 
    }

}
```

PR #2795, opened 2024-12-01
Support multiple indecies in TolerantIndexBuilder

mamazu:
The indexer had a constructor argument that is the `Index`. This is suboptimal if we want to support multiple indices. So now it's an argument when trying to index a file (or list of files).


dantleech:
Can you describe the you envision for this feature? Also beware, as is evident from other PRs, I haven't really got the time or motivation to deep dive into complex topics with Phpactor.

mamazu:
Well my main goal is to reduce the dependency on the index as a constructor argument and move it to the methods that actually need it. Furthermore I'm trying to split up the `IndexAgentBuilder` as it currently is also very intertwined with the `Index`. But as for the steps needed to take, I'm not sure myself yet.

dantleech:
So I don't think this is necessarily the way to go, I need to know what this will look like before knowing how to refactor it.

dantleech:
I've dumped some research here: https://github.com/phpactor/phpactor/blob/645e8e8fd167352a96e8184a2554f14601d245da/doc/adr/0004-indexer.md

mamazu:
Looks like a good idea. In the end I was thinking of the same thing. But I wasn't sure where to best place to put the distinction which index to use was.

But I think the lower the better. So that the amount of classes having to actually deal with an index is as small as possible.

Issue #2414, opened 2023-10-25
Add support for Evaluatable Expression


zobo:
This service helps a vscode (php) debugger to better understand what part of the code can be evaluated - usually in a "hover" context.

If there is code like  `$this->arr['foo'][$xx['bar']]->buz[777]` hovering over different parts should tell the debugger what par of the code to evaluate. For example:
- `$xx` if hovering over `$xx`
- `$xx['bar'] if hovering over `bar`
- `$this->arr` if hovering over `arr`
- `$this->arr['foo'][$xx['bar']]->buz` if hovering over `buz`
- ...

The implementation consists of two parts - registering a provider in VSCode extension and handing the processing of to LS over a non standard LSP call.
https://github.com/zobo/vscode-php-intellisense/pull/64/files#diff-04bba6a35cad1c794cbbe677678a51de13441b7a6ee8592b7b50be1f05c6f626R146

https://github.com/zobo/php-language-server/pull/58/files

Functionality wise it's very similar to the `textDocument/hover` action.

dantleech:
Phpactor basically already evaluates expressions e.g. hover over `$a` in `$a = 1+5 / 2` but happy for this to be added.

zobo:
Sorry, wasn't explicit that I intend to work on this. 😀 
And as you said, there's probably already everything needed in there.

PR #2805, opened 2024-12-25
feat: evaluatable expressions

zobo:
This implements LSP side of (debug) evaluatable expressions for vscode. There is no official LSP message for this, so I implemented this as an experimental feature. I did the same php intellisense.

I'm still a bit lost in the source, so please take a look and give me some pointers how to refactor this to be more in line with the existing code.
- Should I move the `Tests` under `LanguageServerCompletion`?
- Is the semi auto generated message class under `Protocol` namespace Ok?
- Is there any utility function I missed that could replace `byteOffsetRangeForNode`
- Am I even doing this correctly by using the parser directly? I just want to find the variable that can be evaluated under the cursor

### DIFF
--- a/lib/Extension/LanguageServerEvaluatableExpression/Handler/EvaluatableExpressionHandler.php
+++ b/lib/Extension/LanguageServerEvaluatableExpression/Handler/EvaluatableExpressionHandler.php
@@ -8,7 +8,6 @@ use Microsoft\PhpParser\Parser;
 use Microsoft\PhpParser\Node;
 use Microsoft\PhpParser\Token;
 use Phpactor\Extension\LanguageServerBridge\Converter\PositionConverter;
-use Phpactor\Extension\LanguageServerBridge\Converter\RangeConverter;
 use Phpactor\Extension\LanguageServerEvaluatableExpression\Protocol\EvaluatableExpression;
 use Phpactor\LanguageServerProtocol\Position;
 use Phpactor\LanguageServerProtocol\ServerCapabilities;
@@ -17,8 +16,6 @@ use Phpactor\LanguageServer\Core\Handler\CanRegisterCapabilities;
 use Phpactor\LanguageServer\Core\Handler\Handler;
 use Phpactor\LanguageServer\Core\Workspace\Workspace;
 use Phpactor\TextDocument\TextDocumentBuilder;
-use Phpactor\WorseReflection\Core\Util\NodeUtil;
-use Phpactor\TextDocument\ByteOffsetRange;
 use Phpactor\LanguageServerProtocol\Range;
 
 class EvaluatableExpressionHandler implements Handler, CanRegisterCapabilities
@@ -72,25 +69,28 @@ class EvaluatableExpressionHandler implements Handler, CanRegisterCapabilities
     private function nodeToEvaluatable(Node $node): ?EvaluatableExpression
     {
         if ($node instanceof Node\Parameter) {
-            return
-                new EvaluatableExpression(
-                    expression: $node->variableName->getText($node->getFileContents()),
-                    range: $this->byteOffsetRangeForNode($node->variableName, $node),
-                );
+            return $this->evaluatableExpressionForNode($node->variableName, $node);
         }
         if (
             $node instanceof Node\Expression\Variable ||
             $node instanceof Node\Expression\SubscriptExpression ||
-            $node instanceof Node\Expression\MemberAccessExpression ||
-            (($node2 = $node->getFirstAncestor(Node\Expression\SubscriptExpression::class)) && ($node = $node2))
+            $node instanceof Node\Expression\MemberAccessExpression
         ) {
-            return
-                new EvaluatableExpression(
-                    expression: $node->getText(),
-                    range: $this->byteOffsetRangeForNode($node, $node),
-                );
+            return $this->evaluatableExpressionForNode($node, $node);
+        }
+        if ($node2 = $node->getFirstAncestor(Node\Expression\SubscriptExpression::class)) {
+            return $this->evaluatableExpressionForNode($node2, $node2);
         }
         return null;
+    }
+
+    private function evaluatableExpressionForNode(Node|Token $token, Node $textNode): EvaluatableExpression
+    {
+        return
+            new EvaluatableExpression(
+                expression: (string)$token->getText($textNode->getFileContents()),
+                range: $this->byteOffsetRangeForNode($token, $textNode),
+            );
     }
 
     /**
@@ -98,8 +98,9 @@ class EvaluatableExpressionHandler implements Handler, CanRegisterCapabilities
      */
     private function byteOffsetRangeForNode(Node|Token $token, Node $textNode): Range
     {
-        // Note: Cloud have usexd NodeUtil::byteOffsetRangeForNode but it's limited to Variable
-        $boRange = ByteOffsetRange::fromInts($token->getStartPosition(), $token->getEndPosition());
-        return RangeConverter::toLspRange($boRange, $textNode->getFileContents());
+        return new Range(
+            PositionConverter::intByteOffsetToPosition($token->getStartPosition(), $textNode->getFileContents()),
+            PositionConverter::intByteOffsetToPosition($token->getEndPosition(), $textNode->getFileContents()),
+        );
     }
 }
--- a/lib/Extension/LanguageServerEvaluatableExpression/Handler/EvaluatableExpressionHandler.php
+++ b/lib/Extension/LanguageServerEvaluatableExpression/Handler/EvaluatableExpressionHandler.php
@@ -2,7 +2,7 @@
 
 namespace Phpactor\Extension\LanguageServerEvaluatableExpression\Handler;
 
-use function Amp\call;
+use Amp\Success;
 use Amp\Promise;
 use Microsoft\PhpParser\Parser;
 use Microsoft\PhpParser\Node;
@@ -43,25 +43,23 @@ class EvaluatableExpressionHandler implements Handler, CanRegisterCapabilities
         TextDocumentIdentifier $textDocument,
         Position $position
     ): Promise {
-        return call(function () use ($textDocument, $position) {
-            $document = $this->workspace->get($textDocument->uri);
-            $offset = PositionConverter::positionToByteOffset($position, $document->text);
-            $document = TextDocumentBuilder::create($document->text)
-                ->uri($document->uri)
-                ->language('php')
-                ->build();
+        $document = $this->workspace->get($textDocument->uri);
+        $offset = PositionConverter::positionToByteOffset($position, $document->text);
+        $document = TextDocumentBuilder::create($document->text)
+            ->uri($document->uri)
+            ->language('php')
+            ->build();
 
-            $char = substr($document, $offset->toInt(), 1);
+        $char = substr($document, $offset->toInt(), 1);
 
-            // do not provide evaluatable for whitespace
-            if (trim($char) == '') {
-                return null;
-            }
+        // do not provide evaluatable for whitespace
+        if (trim($char) == '') {
+            return new Success(null);
+        }
 
-            $rootNode = $this->parser->parseSourceFile((string) $document, $document->uri()?->__toString());
-            $node = $rootNode->getDescendantNodeAtPosition($offset->toInt());
-            return $this->nodeToEvaluatable($node);
-        });
+        $rootNode = $this->parser->parseSourceFile((string) $document, $document->uri()?->__toString());
+        $node = $rootNode->getDescendantNodeAtPosition($offset->toInt());
+        return new Success($this->nodeToEvaluatable($node));
     }
 
     public function registerCapabiltiies(ServerCapabilities $capabilities): void
--- a/lib/Extension/LanguageServerEvaluatableExpression/Handler/EvaluatableExpressionHandler.php
+++ b/lib/Extension/LanguageServerEvaluatableExpression/Handler/EvaluatableExpressionHandler.php
@@ -60,8 +60,7 @@ class EvaluatableExpressionHandler implements Handler, CanRegisterCapabilities
 
             $rootNode = $this->parser->parseSourceFile((string) $document, $document->uri()?->__toString());
             $node = $rootNode->getDescendantNodeAtPosition($offset->toInt());
-            $ret = $this->nodeToEvaluatable($node);
-            return $ret;
+            return $this->nodeToEvaluatable($node);
         });
     }
 
--- a/lib/Extension/LanguageServerEvaluatableExpression/Handler/EvaluatableExpressionHandler.php
+++ b/lib/Extension/LanguageServerEvaluatableExpression/Handler/EvaluatableExpressionHandler.php
@@ -1,0 +1,115 @@
+<?php
+
+namespace Phpactor\Extension\LanguageServerEvaluatableExpression\Handler;
+
+use function Amp\call;
+use Amp\Promise;
+use Microsoft\PhpParser\Parser;
+use Microsoft\PhpParser\Node;
+use Phpactor\Extension\LanguageServerBridge\Converter\PositionConverter;
+use Phpactor\Extension\LanguageServerBridge\Converter\RangeConverter;
+use Phpactor\Extension\LanguageServerEvaluatableExpression\Protocol\EvaluatableExpression;
+use Phpactor\LanguageServerProtocol\Position;
+use Phpactor\LanguageServerProtocol\ServerCapabilities;
+use Phpactor\LanguageServerProtocol\TextDocumentIdentifier;
+use Phpactor\LanguageServer\Core\Handler\CanRegisterCapabilities;
+use Phpactor\LanguageServer\Core\Handler\Handler;
+use Phpactor\LanguageServer\Core\Workspace\Workspace;
+use Phpactor\ObjectRenderer\Model\ObjectRenderer;
+use Phpactor\TextDocument\TextDocumentBuilder;
+use Phpactor\WorseReflection\Core\Util\NodeUtil;
+use Phpactor\TextDocument\ByteOffsetRange;
+use Phpactor\LanguageServerProtocol\Range;
+
+class EvaluatableExpressionHandler implements Handler, CanRegisterCapabilities
+{
+    public function __construct(
+        private Workspace $workspace,
+        private Parser $parser,
+    ) {
+    }
+
+    public function methods(): array
+    {
+        return [
+            'textDocument/xevaluatableExpression' => 'xevaluatableExpression',
+        ];
+    }
+
+    /**
+     * @return Promise<EvaluatableExpression|null>
+     */
+    public function xevaluatableExpression(
+        TextDocumentIdentifier $textDocument,
+        Position $position
+    ): Promise {
+        return call(function () use ($textDocument, $position) {
+            $document = $this->workspace->get($textDocument->uri);
+            $offset = PositionConverter::positionToByteOffset($position, $document->text);
+            $document = TextDocumentBuilder::create($document->text)
+                ->uri($document->uri)
+                ->language('php')
+                ->build();
+
+            $char = substr($document, $offset->toInt(), 1);
+
+            // do not provide evaluatable for whitespace
+            if (trim($char) == '') {
+                return null;
+            }
+
+            $rootNode = $this->parser->parseSourceFile((string) $document, $document->uri()?->__toString());
+            $node = $rootNode->getDescendantNodeAtPosition($offset->toInt());
+            if ($node) {
+                $ret = $this->nodeToEvaluatable($node);
+                return $ret;
+            }
+            return null;
+        });
+    }
+
+    public function registerCapabiltiies(ServerCapabilities $capabilities): void
+    {
+        // TODO
+        $capabilities->experimental['xevaluatableExpressionProvider'] = true;
+    }
+
+    /**
+     * @param Node $node
+     * @return EvaluatableExpression|null
+     */
+    private function nodeToEvaluatable(Node $node): ?EvaluatableExpression
+    {
+        if ($node instanceof Node\Parameter) {
+            return
+                new EvaluatableExpression(
+                    expression: $node->variableName->getText($node->getFileContents()),
+                    range: $this->byteOffsetRangeForNode($node),
+                );
+        }
+        if (
+            $node instanceof Node\Parameter ||
+            $node instanceof Node\Expression\Variable ||
+            $node instanceof Node\Expression\SubscriptExpression ||
+            $node instanceof Node\Expression\MemberAccessExpression ||
+            (($node2 = $node->getFirstAncestor(Node\Expression\SubscriptExpression::class)) && ($node = $node2))
+        ) {
+            return
+                new EvaluatableExpression(
+                    expression: $node->getText($node->getFileContents()),
+                    range: $this->byteOffsetRangeForNode($node),
+                );
+    }
+        return null;
+    }
+
+    /**
+     * Converts Microsoft PhpParser Node to LSP Range.
+     */
+    private function byteOffsetRangeForNode(Node $node): Range
+    {
+        // Note: Cloud have usexd NodeUtil::byteOffsetRangeForNode but it's limited to Variable
+        $boRange = ByteOffsetRange::fromInts($node->getStartPosition(), $node->getEndPosition());
+        return RangeConverter::toLspRange($boRange, $node->getFileContents());
+    }    
+}
--- a/lib/Extension/LanguageServerEvaluatableExpression/Handler/EvaluatableExpressionHandler.php
+++ b/lib/Extension/LanguageServerEvaluatableExpression/Handler/EvaluatableExpressionHandler.php
@@ -64,6 +64,8 @@ class EvaluatableExpressionHandler implements Handler, CanRegisterCapabilities
 
     public function registerCapabiltiies(ServerCapabilities $capabilities): void
     {
+        $capabilities->experimental ??= [];
+        // @phpstan-ignore offsetAccess.nonOffsetAccessible
         $capabilities->experimental['xevaluatableExpressionProvider'] = true;
     }
 
--- a/lib/Extension/LanguageServerEvaluatableExpression/Handler/EvaluatableExpressionHandler.php
+++ b/lib/Extension/LanguageServerEvaluatableExpression/Handler/EvaluatableExpressionHandler.php
@@ -15,7 +15,6 @@ use Phpactor\LanguageServerProtocol\TextDocumentIdentifier;
 use Phpactor\LanguageServer\Core\Handler\CanRegisterCapabilities;
 use Phpactor\LanguageServer\Core\Handler\Handler;
 use Phpactor\LanguageServer\Core\Workspace\Workspace;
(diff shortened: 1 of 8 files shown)


zobo:
Implements #2414

zobo:
What do I do with phpstan? Thers some errors that don't make much sense.. Ignore all?
Wthat's the command to update the baseline?

lib/Extension/LanguageServerEvaluatableExpression/Handler/EvaluatableExpressionHandler.php
```diff
@@ -0,0 +1,106 @@
...
+                    expression: $node->variableName->getText($node->getFileContents()),
+                    range: $this->byteOffsetRangeForNode($node->variableName, $node),
+                );
+        }
+        if (
+            $node instanceof Node\Expression\Variable ||
+            $node instanceof Node\Expression\SubscriptExpression ||
+            $node instanceof Node\Expression\MemberAccessExpression ||
+            (($node2 = $node->getFirstAncestor(Node\Expression\SubscriptExpression::class)) && ($node = $node2))
+        ) {
```
dantleech, line 87: maybe extract `$this->isXXXExpression` and avoid assigning in a conditional expression?

lib/Extension/LanguageServerEvaluatableExpression/Handler/EvaluatableExpressionHandler.php
```diff
@@ -0,0 +1,106 @@
...
+
+    /**
+     * @return Promise<EvaluatableExpression|null>
+     */
+    public function xevaluatableExpression(
+        TextDocumentIdentifier $textDocument,
+        Position $position
+    ): Promise {
+        return call(function () use ($textDocument, $position) {
+            $document = $this->workspace->get($textDocument->uri);
```
dantleech, line 47: Here I think PHPStan _would_ understand this if the `function()` was a generator, but it's not.

You can fix this by not using `call()` and returnjing `new Success($result)` instead:
```diff
diff --git a/lib/Extension/LanguageServerEvaluatableExpression/Handler/EvaluatableExpressionHandler.php b/lib/Extension/LanguageServerEvaluatableExpression/Handler/Evaluata>
index f675ef07..df0d8c55 100644
--- a/lib/Extension/LanguageServerEvaluatableExpression/Handler/EvaluatableExpressionHandler.php
+++ b/lib/Extension/LanguageServerEvaluatableExpression/Handler/EvaluatableExpressionHandler.php
@@ -2,6 +2,7 @@

 namespace Phpactor\Extension\LanguageServerEvaluatableExpression\Handler;

+use Amp\Success;
 use function Amp\call;
 use Amp\Promise;
 use Microsoft\PhpParser\Parser;
@@ -43,26 +44,23 @@ class EvaluatableExpressionHandler implements Handler, CanRegisterCapabilities
         TextDocumentIdentifier $textDocument,
         Position $position
     ): Promise {
-        return call(function () use ($textDocument, $position) {
-            $document = $this->workspace->get($textDocument->uri);
-            $offset = PositionConverter::positionToByteOffset($position, $document->text);
-            $document = TextDocumentBuilder::create($document->text)
-                ->uri($document->uri)
-                ->language('php')
-                ->build();
+        $document = $this->workspace->get($textDocument->uri);
+        $offset = PositionConverter::positionToByteOffset($position, $document->text);
+        $document = TextDocumentBuilder::create($document->text)
+            ->uri($document->uri)
+            ->language('php')
+            ->build();

-            $char = substr($document, $offset->toInt(), 1);
+        $char = substr($document, $offset->toInt(), 1);

-            // do not provide evaluatable for whitespace
-            if (trim($char) == '') {
-                return null;
-            }
+        // do not provide evaluatable for whitespace
+        if (trim($char) == '') {
+            return new Success(null);
+        }

-            $rootNode = $this->parser->parseSourceFile((string) $document, $document->uri()?->__toString());
-            $node = $rootNode->getDescendantNodeAtPosition($offset->toInt());
-            $ret = $this->nodeToEvaluatable($node);
-            return $ret;
-        });
+        $rootNode = $this->parser->parseSourceFile((string) $document, $document->uri()?->__toString());
+        $node = $rootNode->getDescendantNodeAtPosition($offset->toInt());
+        return new Success($this->nodeToEvaluatable($node));
     }
```

lib/Extension/LanguageServerEvaluatableExpression/Handler/EvaluatableExpressionHandler.php
```diff
@@ -0,0 +1,106 @@
...
+
+            // do not provide evaluatable for whitespace
+            if (trim($char) == '') {
+                return null;
+            }
+
+            $rootNode = $this->parser->parseSourceFile((string) $document, $document->uri()?->__toString());
+            $node = $rootNode->getDescendantNodeAtPosition($offset->toInt());
+            $ret = $this->nodeToEvaluatable($node);
+            return $ret;
```
dantleech, line 64: could remove intermediate variable `$ret`.

lib/Extension/LanguageServerEvaluatableExpression/Handler/EvaluatableExpressionHandler.php
```diff
@@ -0,0 +1,106 @@
...
+        });
+    }
+
+    public function registerCapabiltiies(ServerCapabilities $capabilities): void
+    {
+        $capabilities->experimental['xevaluatableExpressionProvider'] = true;
```
dantleech, line 70: I'm not sure why LSP defines this as `any` (i.e. `mixed`) is it guaranteed to be an array at this point? should we throw an exception or ignore it if it's not?

lib/Extension/LanguageServerEvaluatableExpression/Handler/EvaluatableExpressionHandler.php
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    private function nodeToEvaluatable(Node $node): ?EvaluatableExpression
+    {
+        if ($node instanceof Node\Parameter) {
+            return
+                new EvaluatableExpression(
+                    expression: $node->variableName->getText($node->getFileContents()),
+                    range: $this->byteOffsetRangeForNode($node->variableName, $node),
```
dantleech, line 79: as below maybe extract an method here describing what this does and avoid the inline assignment. to avoid the `getFileContents()` `string|null` error just cast it to a string: `getText((string)$node->getFileContents())`

lib/Extension/LanguageServerEvaluatableExpression/Handler/EvaluatableExpressionHandler.php
zobo: This was a style mostly used in php-intellisense. I'll split up the code and make it more readable.

lib/Extension/LanguageServerEvaluatableExpression/LanguageServerEvaluatableExpressionExtension.php
```diff
@@ -0,0 +1,28 @@
...
+use Phpactor\Container\Extension;
+use Phpactor\MapResolver\Resolver;
+
+class LanguageServerEvaluatableExpressionExtension implements Extension
+{
+    public function load(ContainerBuilder $container): void
+    {
+        $container->register('language_server_completion.handler.evaluatable_expression', function (Container $container) {
+            return new EvaluatableExpressionHandler(
+                $container->get(LanguageServerExtension::SERVICE_SESSION_WORKSPACE),
```
dantleech, line 19: you can do `$container->expect(LanguageServerExtension::SERVICE_SESSION_WORKSPACE, Workspace::class)` to make phpstan happy.

lib/Extension/LanguageServerEvaluatableExpression/LanguageServerEvaluatableExpressionExtension.php
```diff
@@ -0,0 +1,28 @@
...
+                $container->get(WorseReflectionExtension::SERVICE_PARSER),
```
dantleech, line 20: same

lib/Extension/LanguageServerEvaluatableExpression/Handler/EvaluatableExpressionHandler.php
zobo: Yes, the async `call()` is a leftover from the `HoverHanlder` that I used as a starting point. But even there, phpstan does not process it correctly, because there is an exception in `phpstan-baseline.neon`

```
		-
			message: "#^Method Phpactor\\\\Extension\\\\LanguageServerHover\\\\Handler\\\\HoverHandler\\:\\:hover\\(\\) should return Amp\\\\Promise\\<Phpactor\\\\LanguageServerProtocol\\\\Hover\\|null\\> but returns Amp\\\\Promise\\<mixed\\>\\.$#"
			count: 1
			path: lib/Extension/LanguageServerHover/Handler/HoverHandler.php
```

I'm not sure about all the architecture, but even though `xevaluatableExpression` does not have any async code in it, leaving it inside `call()` allows the `Amp` loop to process other messages before spending CPU in this call... I could be wrong though...

lib/Extension/LanguageServerEvaluatableExpression/Protocol/EvaluatableExpression.php
```diff
@@ -0,0 +1,96 @@
...
+            if (!is_array($value)) {
+                continue;
+            }
+
+            if (empty($map[$key]['names'])) {
+                continue;
+            }
+
+            if ($map[$key]['iterable']) {
+                $value = array_map(function ($object) use ($map, $key, $allowUnknownKeys) {
```
dantleech, line 59: as this was generated code it can be simplified and the above condition _is_ always false. I would simplify this:

```
fromArray(array $array, bool $whatever) {
    return new self($array['one'], $array['two'], $array['three']);
}
```

and add some simple guard conditions to make phpstan happy.

lib/Extension/LanguageServerEvaluatableExpression/Handler/EvaluatableExpressionHandler.php
zobo: Yes. Debugging leftover.

phpstan-baseline.neon
```diff
@@ -3675,6 +3675,16 @@ parameters:
...
+		-
+			message: "#^Cannot access property \\$result on Phpactor\\\\LanguageServer\\\\Core\\\\Rpc\\\\ResponseMessage\\|null\\.$#"
+			count: 1
+			path: lib/Extension/LanguageServerEvaluatableExpression/Tests/EvaluatableExpressionHandlerTest.php
+
+		-
+			message: "#^Parameter \\#1 \\$response of method Phpactor\\\\LanguageServer\\\\Test\\\\LanguageServerTester\\:\\:assertSuccess\\(\\) expects Phpactor\\\\LanguageServer\\\\Core\\\\Rpc\\\\ResponseMessage, Phpactor\\\\LanguageServer\\\\Core\\\\Rpc\\\\ResponseMessage\\|null given\\.$#"
+			count: 1
+			path: lib/Extension/LanguageServerEvaluatableExpression/Tests/EvaluatableExpressionHandlerTest.php
+
```
dantleech, line 3687: the baseline is only allowed to get _smaller_ both of these cases should be handled. the fix for this is just:

```diff
diff --git a/lib/Extension/LanguageServerEvaluatableExpression/Tests/EvaluatableExpressionHandlerTest.php b/lib/Extension/LanguageServerEvaluatableExpression/Tests/Evaluata>
index f0d865ea..133801fd 100644
--- a/lib/Extension/LanguageServerEvaluatableExpression/Tests/EvaluatableExpressionHandlerTest.php
+++ b/lib/Extension/LanguageServerEvaluatableExpression/Tests/EvaluatableExpressionHandlerTest.php
@@ -30,6 +30,7 @@ class EvaluatableExpressionHandlerTest extends IntegrationTestCase
             'textDocument' => new TextDocumentIdentifier(self::PATH),
             'position' => PositionConverter::byteOffsetToPosition(ByteOffset::fromInt((int)$offset), $text)
         ]);
+        self::assertNotNull($response);
         $tester->assertSuccess($response);
         $result = $response->result;
         $this->assertInstanceOf(EvaluatableExpression::class, $result);
```

lib/Extension/LanguageServerEvaluatableExpression/Handler/EvaluatableExpressionHandler.php
zobo: It worked, but I do agree this could be a problem. It's defined as `any` so people can add in it whatever they want. TypeScript compiler/linter still [complains](https://github.com/phpactor/vscode-phpactor/pull/167/files#diff-04bba6a35cad1c794cbbe677678a51de13441b7a6ee8592b7b50be1f05c6f626R62) though. I need to investigate code patterns from Microsoft there...

What would be a good solution here?
```php
$capabilities->experimental ??= [];
$capabilities->experimental['xevaluatableExpressionProvider'] = true;
```

When the `InitializeMiddleware` calls the `registerCapabiltiies` method, `$capabilities->experimental` is currently `null`.

lib/Extension/LanguageServerEvaluatableExpression/LanguageServerEvaluatableExpressionExtension.php
zobo: I tried `expect()`, but there was a runtime error because the `SERVICE_PARSER` was actually the cached version of the parser. I like and want it to be cached, but I wondered this was the right way to do this... Will take another look.

lib/Extension/LanguageServerEvaluatableExpression/LanguageServerEvaluatableExpressionExtension.php
zobo: ok

phpstan-baseline.neon
zobo: Nice. Will clean up today as I get a chance. Thanks!

lib/Extension/LanguageServerEvaluatableExpression/Protocol/EvaluatableExpression.php
zobo: Yeah. Wanted to use the generator... That one is also out of date and I had to patch it first.. Will submit a PR there.

Perhaps you're right. I'll just clean up and hand code the minimum.

dantleech:
> Should I move the Tests under LanguageServerCompletion?

The test is in the right place, but it technically should not extend the `IntegrationTestCase` of another module. In this case the requirements are pretty simple and can extend the PHPUnit `TestCase` anyhow:

```
diff --git a/lib/Extension/LanguageServerEvaluatableExpression/Tests/EvaluatableExpressionHandlerTest.php b/lib/Extension/LanguageServerEvaluatableExpression/Tests/Evaluata>
index f0d865ea..165e81ef 100644
--- a/lib/Extension/LanguageServerEvaluatableExpression/Tests/EvaluatableExpressionHandlerTest.php
+++ b/lib/Extension/LanguageServerEvaluatableExpression/Tests/EvaluatableExpressionHandlerTest.php
@@ -3,14 +3,18 @@
 namespace Phpactor\Extension\LanguageServerEvaluatableExpression\Tests;

 use Generator;
+use Microsoft\PhpParser\Parser;
 use Phpactor\Extension\LanguageServerBridge\Converter\PositionConverter;
+use Phpactor\Extension\LanguageServerEvaluatableExpression\Handler\EvaluatableExpressionHandler;
 use Phpactor\Extension\LanguageServerEvaluatableExpression\Protocol\EvaluatableExpression;
 use Phpactor\LanguageServerProtocol\TextDocumentIdentifier;
-use Phpactor\Extension\LanguageServerCompletion\Tests\IntegrationTestCase;
+use Phpactor\LanguageServer\LanguageServerTesterBuilder;
+use Phpactor\LanguageServer\Test\LanguageServerTester;
 use Phpactor\TestUtils\ExtractOffset;
+use Phpactor\TestUtils\PHPUnit\TestCase;
 use Phpactor\TextDocument\ByteOffset;

-class EvaluatableExpressionHandlerTest extends IntegrationTestCase
+class EvaluatableExpressionHandlerTest extends TestCase
 {
     private const PATH = 'file:///hello';

@@ -30,6 +34,7 @@ class EvaluatableExpressionHandlerTest extends IntegrationTestCase
             'textDocument' => new TextDocumentIdentifier(self::PATH),
             'position' => PositionConverter::byteOffsetToPosition(ByteOffset::fromInt((int)$offset), $text)
         ]);
+        self::assertNotNull($response);
         $tester->assertSuccess($response);
         $result = $response->result;
         $this->assertInstanceOf(EvaluatableExpression::class, $result);
@@ -57,4 +62,11 @@ class EvaluatableExpressionHandlerTest extends IntegrationTestCase
             '<?php function test(Sometype <>$a<>rg<>) {};',
         ];
     }
+
+    protected function createTester(): LanguageServerTester
+    {
+        $tester = LanguageServerTesterBuilder::create();
+        $tester->addHandler(new EvaluatableExpressionHandler($tester->workspace(), new Parser()));
+        return $tester->build();
+    }
 }
```

> Is the semi auto generated message class under Protocol namespace Ok?

as noted we can simplify it. there's no point in keeping the over-generalized generated code

> Is there any utility function I missed that could replace byteOffsetRangeForNode

`NodeUtil::byteOffsetRangeFromNode` should be changed to accept `Node` - but I'd also question if we need to go through Phpactor's representation here, could bypass it and:

```
        return new Range(
            PositionConverter::byteOffsetToPosition($node->getStartPosition(), $text),
            PositionConverter::byteOffsetToPosition($range->getEndPosition(), $text),
        );
``` 

> Am I even doing this correctly by using the parser directly? I just want to find the variable that can be evaluated under the cursor

it's generally better to use the parser only if no static analysis is required.

lib/Extension/LanguageServerEvaluatableExpression/Handler/EvaluatableExpressionHandler.php
dantleech: basically the request comes it, it gets routed to here and even if there was some "other" work to do it makes no difference if this is executed now or on the next tick, so I think executing it directly is fine.

lib/Extension/LanguageServerEvaluatableExpression/LanguageServerEvaluatableExpressionExtension.php
dantleech: the 2nd argument has no effect on the service that is returned, it doesn't even type check IIRC (although it probably should)

lib/Extension/LanguageServerEvaluatableExpression/Handler/EvaluatableExpressionHandler.php
dantleech: hm. probably we should make an opinion and initiailize the `expreimental` field as an `array` ~or better a class with `public $xevluatableExpressionProvider`)~. The problem is that the `ServerCapabiltiies` class is instantiated in the Language Server `InitializeMiddleware`, so we can either:

- Take ownership of the `InitializeMiddleware` from the Phpactor Language Server package (i.e. copy and paste the class and change it)
- Introduce a factory for `ServerCapabilities` in the Language Server package.
- Take an opinion and unconditionally initialize it as an array in that class.

Alternatively we can do what you suggest but also throw an exception if it's neither `null` nor `array`, but it's not a real solution.

dantleech: On second thoughts, those are only partial solutions because `ServerCapabiltiies#experimental` will always be `mixed` statically. although you could more legitimately ignore the error then.

for now maybe just do what you suggest and either ignore the phpstan error or a type assertion.

lib/Extension/LanguageServerEvaluatableExpression/LanguageServerEvaluatableExpressionExtension.php
zobo: Well, actually: https://github.com/phpactor/container/blob/143bbd987da798f2d7e5cedafd35b49d5051e167/lib/PhpactorContainer.php#L166 .
But I think I made a mistake with namespaces, so that's why it failed.

lib/Extension/LanguageServerEvaluatableExpression/Handler/EvaluatableExpressionHandler.php
zobo: Reduced the code a bit... I tried to be smart with `match` but it's not allowed to do an assignment inside a match case...

zobo:
> `NodeUtil::byteOffsetRangeFromNode` should be changed to accept `Node` - but I'd also question if we need to go through Phpactor's representation here, could bypass it and:
> 
> ```
>         return new Range(
>             PositionConverter::byteOffsetToPosition($node->getStartPosition(), $text),
>             PositionConverter::byteOffsetToPosition($range->getEndPosition(), $text),
>         );
> ```

I did it like this, but honestly before it made more sense... Object translation is always problematic :)

zobo:
> The test is in the right place, but it technically should not extend the `IntegrationTestCase` of another module. In this case the requirements are pretty simple and can extend the PHPUnit `TestCase` anyhow:
> 
> ```
> diff --git a/lib/Extension/LanguageServerEvaluatableExpression/Tests/EvaluatableExpressionHandlerTest.php b/lib/Extension/LanguageServerEvaluatableExpression/Tests/Evaluata>
> index f0d865ea..165e81ef 100644
> --- a/lib/Extension/LanguageServerEvaluatableExpression/Tests/EvaluatableExpressionHandlerTest.php
> +++ b/lib/Extension/LanguageServerEvaluatableExpression/Tests/EvaluatableExpressionHandlerTest.php
> @@ -3,14 +3,18 @@
>  namespace Phpactor\Extension\LanguageServerEvaluatableExpression\Tests;
> 
>  use Generator;
> +use Microsoft\PhpParser\Parser;
>  use Phpactor\Extension\LanguageServerBridge\Converter\PositionConverter;
> +use Phpactor\Extension\LanguageServerEvaluatableExpression\Handler\EvaluatableExpressionHandler;
>  use Phpactor\Extension\LanguageServerEvaluatableExpression\Protocol\EvaluatableExpression;
>  use Phpactor\LanguageServerProtocol\TextDocumentIdentifier;
> -use Phpactor\Extension\LanguageServerCompletion\Tests\IntegrationTestCase;
> +use Phpactor\LanguageServer\LanguageServerTesterBuilder;
> +use Phpactor\LanguageServer\Test\LanguageServerTester;
>  use Phpactor\TestUtils\ExtractOffset;
> +use Phpactor\TestUtils\PHPUnit\TestCase;
>  use Phpactor\TextDocument\ByteOffset;
> 
> -class EvaluatableExpressionHandlerTest extends IntegrationTestCase
> +class EvaluatableExpressionHandlerTest extends TestCase
>  {
>      private const PATH = 'file:///hello';
> 
> @@ -30,6 +34,7 @@ class EvaluatableExpressionHandlerTest extends IntegrationTestCase
>              'textDocument' => new TextDocumentIdentifier(self::PATH),
>              'position' => PositionConverter::byteOffsetToPosition(ByteOffset::fromInt((int)$offset), $text)
>          ]);
> +        self::assertNotNull($response);
>          $tester->assertSuccess($response);
>          $result = $response->result;
>          $this->assertInstanceOf(EvaluatableExpression::class, $result);
> @@ -57,4 +62,11 @@ class EvaluatableExpressionHandlerTest extends IntegrationTestCase
>              '<?php function test(Sometype <>$a<>rg<>) {};',
>          ];
>      }
> +
> +    protected function createTester(): LanguageServerTester
> +    {
> +        $tester = LanguageServerTesterBuilder::create();
> +        $tester->addHandler(new EvaluatableExpressionHandler($tester->workspace(), new Parser()));
> +        return $tester->build();
> +    }
>  }
> ```

I can do it like this, but I don't like the fact that this way the test does not run through `LanguageServerEvaluatableExpressionExtension`. I tried setting up `LanguageServerTesterBuilder` by somehow using `PhpactorContainer::fromExtensions` but didn't manage so far...

lib/Extension/LanguageServerEvaluatableExpression/Protocol/EvaluatableExpression.php
zobo: I don't think `fromArray` is ever used, as this class is just used in LS->IDE direction... anyway...

dantleech:
> I can do it like this...but I don't like the fact...

It means spinning up all the dependencies, in this case I think the instantiation is simple enough to not worry about an end-to-end. Lot's of things in Phpactor are not e2e tested, making that easier might be something for the future.

dantleech:
> I did it like this, but honestly before it made more sense

Yeah whatever you prefer, feel free to change the signature for the until method

zobo:
> > I can do it like this...but I don't like the fact...
> 
> It means spinning up all the dependencies, in this case I think the instantiation is simple enough to not worry about an end-to-end. Lot's of things in Phpactor are not e2e tested, making that easier might be something for the future.

Ok. Simplified the test. Just saying it helped a lot when I was figuring out the code and made mistakes in the Extension :)

zobo:
> > I did it like this, but honestly before it made more sense
> 
> Yeah whatever you prefer, feel free to change the signature for the util method

I'll leave it in this extension because Microsoft uses Node and Token. These don't share the same base class even though they have (mostly) compatible methods - at least for what I need here.

I'll still change the signature of `NodeUtil::byteOffsetRangeForNode` simply because it makes sense.

zobo:
That's it :)

zobo:
One design question though: I intend to also implement the inline debug values message. These two are vscode specific and both debugger related. Would it make sense to combine the two under a single Debuggger Extension?

dantleech:
> One design question though

Yeah I think it makes sense. Only issue would be if it could/would conflict with another client or unrelated debug feature. Maybe "VscodeDebugExtension"?

lib/Extension/LanguageServerEvaluatableExpression/Handler/EvaluatableExpressionHandler.php
```diff
@@ -0,0 +1,106 @@
...
+        Position $position
+    ): Promise {
+        $document = $this->workspace->get($textDocument->uri);
+        $offset = PositionConverter::positionToByteOffset($position, $document->text);
+        $document = TextDocumentBuilder::create($document->text)
+            ->uri($document->uri)
+            ->language('php')
+            ->build();
+
+        $char = substr($document, $offset->toInt(), 1);
```
dantleech, line 50: technically this should be `mb_substr` but for the purpose it doesn't matter

lib/Extension/LanguageServerEvaluatableExpression/LanguageServerEvaluatableExpressionExtension.php
```diff
@@ -0,0 +1,30 @@
...
+use Phpactor\Container\Extension;
+use Phpactor\MapResolver\Resolver;
+use Microsoft\PhpParser\Parser;
+use Phpactor\LanguageServer\Core\Workspace\Workspace;
+
+class LanguageServerEvaluatableExpressionExtension implements Extension
+{
+    public function load(ContainerBuilder $container): void
+    {
+        $container->register('language_server_completion.handler.evaluatable_expression', function (Container $container) {
```
dantleech, line 19: `language_server_evaluatable_expression.handler` (`<module_name>.<the thing>`)

dantleech:
thanks, i fixed the service name and updated the changelog in master :bow:

PR #2807, opened 2024-12-26
Removing unused classes

mamazu:
They are not used anywhere so if the pipeline is green, then there is no reason to keep them around.


dantleech:
:bow:

mamazu:
If you're curious I'm using the `tomasvotruba/class-leak` package that shows unused classes, interfaces etc.

PR #2815, opened 2025-01-04
fix(completion): magic method names displayed inside argument list

przepompownia:


przepompownia:
I fixed this bug at work on #2622 but it turns to be independent on that PR and can be fixed separately. 

Screenshot From 2025-01-02 20-34-02

mamazu:
I've just checked it out and yes, it doesn't suggest magic methods anymore but it still suggests keywords.
image

przepompownia:
I caught it when made the recent fix for #2622 and fixed (https://github.com/phpactor/phpactor/pull/2622/commits/0250c05579cf330bbae998461d944ac8a7605a12#diff-2ba75b9069ec2e0eca8655a5fa214ecb0050bf8656e6c1842d2be83b91d31f27R93) but still cannot reproduce it here.

przepompownia:
That fix assumes that these two words make no sense inside `StatementNode`.

Even if we found test data for that, let keep this PR as small as possible and fix that separately - I keep https://github.com/phpactor/phpactor/pull/2795#issuecomment-2510945959 in mind.

mamazu:
Yes, it's already better than before and I think this should be merged.

But in this case it doesn't make sense to have a `StatementNode` in the parameter definition. Maybe an `ExpressionNode` for default values.

przepompownia:
Thank you! I'll try to extract the function|const case fix too, assuming I'll have a reproducible test case.

dantleech:
Note that this is all very unstable, adding one thing may break another as it's all just working around the parser being bad at invalid ASTs

przepompownia:
More, the logic needed to recognize the current context needs to be a bit hacky - at least I have no better ideas than some of uses in #2622 

That's why we need more and more test cases.

lib/Completion/Bridge/TolerantParser/CompletionContext.php
```diff
@@ -177,12 +177,16 @@ public static function classMembersBody(?Node $node): bool
             return false;
         }
 
-        $nodeBeforeOffset = NodeUtil::firstDescendantNodeBeforeOffset($node->getRoot(), $node->parent->getStartPosition());
-
         if ($node instanceof Variable) {
             return false;
         }
 
+        if ($node->parent instanceof MethodDeclaration && $node->openBrace instanceof MissingToken) {
```
dantleech, line 184: this causes lots of warnings:
```
[ERROR][2025-01-27 09:45:46] .../vim/lsp/rpc.lua:770	"rpc"	"/home/daniel/.local/bin/phpactor"	"stderr"	"Warning: Undefined property: Microsoft\\PhpParser\\Node\\DelimitedList\\ParameterDeclarationList::$openBrace in /home/daniel/www/phpactor/phpactor/lib/Completion/Bridge/TolerantParser/CompletionContext.php on line 184\n"
[ERROR][2025-01-27 09:45:56] .../vim/lsp/rpc.lua:770	"rpc"	"/home/daniel/.local/bin/phpactor"	"stderr"	"Warning: Undefined property: Microsoft\\PhpParser\\Node\\DelimitedList\\ParameterDeclarationList::$openBrace in /home/daniel/www/phpactor/phpactor/lib/Completion/Bridge/TolerantParser/CompletionContext.php on line 184\n"
[ERROR][2025-01-27 09:45:56] .../vim/lsp/rpc.lua:770	"rpc"	"/home/daniel/.local/bin/phpactor"	"stderr"	"Warning: Undefined property: Microsoft\\PhpParser\\Node\\DelimitedList\\ParameterDeclarationList::$openBrace in /home/daniel/www/phpactor/phpactor/lib/Completion/Bridge/TolerantParser/CompletionContext.php on line 184\n"

```

przepompownia: #2832 should precise the type of `$node` in that case.

PR #2816, opened 2025-01-04
Multi-index / Workspace Folders etc.

dantleech:
Initial research into "multi index" and then how we probably don't even need that.

### DIFF
--- a/doc/adr/0004-indexer.md
+++ b/doc/adr/0004-indexer.md
@@ -1,0 +1,123 @@
+New Indexer
+===========
+
+Supporting "mutliple indexes" has been a feature request for some time, but
+why?
+
+- **Workspace Folders**: Supporting multiple workspace folders - particularly
+  useful for editors such as VSCode and Sublime Text.
+- **Index Size**: Being able to reuse f.e. the same `phpstorm-stubs` index on
+  all projects, as currently they are reindexed each time.
+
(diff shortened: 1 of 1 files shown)


mamazu:
Yes this makes the most sense.
Having the stubs and the vendor directory marked as readonly indexes by default might also open up for optimizations (eg. no dirty tracking etc)

zobo:
There's also an easier, far more ugly solution: Spawn one instance of Phpactor per workspace https://github.com/zobo/vscode-php-intellisense/pull/34 

Not recommending, just saying...

doc/adr/0004-indexer.md
```diff
@@ -0,0 +1,123 @@
...
+[Worspace Folders Blog Post](https://scalameta.org/metals/blog/2023/07/17/workspace-folders/).
+
+In Metals each endpoint first calls `getServiceFor` which returns an "service"
+configured for the **workspace containing the text document**:
+
+```
+  override def completion(
+      params: CompletionParams
+  ): CompletableFuture[CompletionList] =
+    getServiceFor(params.getTextDocument.getUri).completion(params)
```
mamazu, line 51: I think the problem with this approach does not allow cross workspace completions. Which means that would be incompatible with splitting off the stubs into it's own index.

PR #2819, opened 2025-01-04
KeywordCompletor: support `match` and possible other keyword expressions

przepompownia:


przepompownia:
It's yet another side effect of working on #2622 that could be applied separately.

dantleech:
it's not immediately clear what this PR does... in what way does it support match and throw expressions?

przepompownia:
Isn't `KeywordCompletor` enough in the title? 

I guess `CHANGELOG.md` should be updated too in this case.

dantleech:
> Isn't KeywordCompletor enough in the title?
not really - for example:

> This MR will fix ensure that keywords are suggested in completion results,

or

> This MR will ensure that match and throw keywords are recognised as _expressions_ and allow subsequent expression completions.

or whatever - I had no idea this had anything to do with snippets :slightly_smiling_face:

przepompownia:
> > Isn't KeywordCompletor enough in the title?
> 
> not really - for example:
> 
> > This MR will ensure that the `match` and `throw` keywords are suggested in completion results,
> 
> or
> 
> > This MR will ensure that match and throw keywords are recognised as _expressions_ and allow subsequent expression completions.
> 
> or whatever - I had no idea this had anything to do with snippets 🙂 basically I don't want to have to reverse engineer (and guess) what the intent of the PR is from looking at the code.

I appreciate this idea and try to keep it in mind and apply, but it is not always clear for me what user the commit is addressed for, especially when I look at the previous commit titles from the master branch :wink:. On the other side my (still poor) English may cause additional confusion - please don't hesitate to correct my possibly strange language constructions.

dantleech:
>  but it is not always clear for me what user the commit is addressed for

well you can assume that they're aimed at me or otherwise the person that's reviewing the PR :slightly_smiling_face:  My own PRs often don't feature extensive descriptions but I'm the one merging and reviewing them so have the maintainer's privilege there.

If my PR breaks Phpactor it's my fault - if your MR breaks Phpactor it's also my fault :smile: so any help to make my life easier is appreciated.

dantleech:
so with this MR using an open parenthesis `(` results in `throw` and `match` being suggested.

przepompownia:
Yes, I see that the case of cursor placed right after open parenthesis `(` is often not directly recognized by tolerant parser and requires to be expressed in an indirect, a bit hacky way. Thus the need to be guarded by enough tests.

PR #2824, opened 2025-01-13
gh-2811: PHP8.4: New without parenthesis

### LINKED ISSUE #2811: `new MyClass()->method()` without parentheses breaks parsing
I checked if tolerant parser has at least PR supporting it but no one seems to be interested in such support.

https://wiki.php.net/rfc/new_without_parentheses

dantleech:
Fixes #2811

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Changelog
 
 Features:
 
+  - Support new expression without parenthesis #2811
   - Support vscode evaluatable expressions #2905 @zobo
 
 ## 2024-11-28.1
--- a/composer.lock
+++ b/composer.lock
@@ -1816,12 +1816,12 @@
             "source": {
                 "type": "git",
                 "url": "https://github.com/phpactor/tolerant-php-parser.git",
-                "reference": "3bd51bd63a2aa47e9c686c21b8833eb67d2c4808"
+                "reference": "8a42bdf799fd8868c85f6c6926014fefa5454ffc"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/phpactor/tolerant-php-parser/zipball/3bd51bd63a2aa47e9c686c21b8833eb67d2c4808",
-                "reference": "3bd51bd63a2aa47e9c686c21b8833eb67d2c4808",
+                "url": "https://api.github.com/repos/phpactor/tolerant-php-parser/zipball/8a42bdf799fd8868c85f6c6926014fefa5454ffc",
+                "reference": "8a42bdf799fd8868c85f6c6926014fefa5454ffc",
                 "shasum": ""
             },
             "require": {
@@ -1854,7 +1854,7 @@
             "support": {
                 "source": "https://github.com/phpactor/tolerant-php-parser/tree/main"
             },
-            "time": "2024-09-03T16:49:23+00:00"
+            "time": "2025-01-13T20:19:07+00:00"
         },
         {
             "name": "psr/container",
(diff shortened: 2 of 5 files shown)


downhiller:
Am I being dumb dan?  (Almost certainly).  I've deleted my phpactor diretory, and its cache.  Re-run `git clone`, `composer install`, `phpactor index:build -reset`.  But in neovim using new without parentheses still comes up with an error ("Function foo not found") on the first method call after instantiation?

Thank you!

dantleech:
you'll need to make sure you're on the `master` branch:

```
new Foo()
  ->bar()
  ->baz();
```

works for me (even on an 8.3 runtime, although I get parse errors from PHP itself)

downhiller:
I am - checked with `git branch`, but `master` is the default when cloning from your repo anyway.  Can you think of anything else it could be?  Sorry!

dantleech:
If you'd like to investigate copy this test case:

lib/WorseReflection/Tests/Inference/new/new-no-parenthesis.test

and put your class structure and method call chain in there so that:

```
wrAssertType('string', new Message()->setWarning()->addText('blah')->getHtml());
```

then run
```
./vendor/bin/phpunit lib/WorseReflection/tests/Inference
```

to see if it works and if not you can make a PR

downhiller:
That does appear to pass.  Though I don't understand, to me it looks like all it does is asserts that the return type of getHtml is string? 🤷‍♂️

Here's some very similar code in a new file with nothing else, showing the failure that I believe I'm having.

image

dantleech:
well, if it asserts that it's a string then it's working. Are you sure these are Phpactor errors and not PHPStan ones? which PHP runtime are you using on your host?

downhiller:
PHP 8.4.3

I _was_ pretty sure that this message was coming from phpactor, I'll feel even more of a plum than normal if it's coming from phpstan.  I _think_ my phpstan only does anything when I manually run it from the commandline, and I _think_ these errors were coming up before I even installed phpstan quite recently.

downhiller:
If I add a line, `thisDoesntExist();` to that file, and run phpstan on it, it correctly fingers that function as being missing, but doesn't say anything about `setWarning()`, so I'm 99% sure the message isn't coming from phpstan.

downhiller:
> well, if it asserts that it's a string then it's working. Are you sure these are Phpactor errors and not PHPStan ones (assuming you have phpstan installed and enabled the phpactor integration)? which PHP runtime are you using on your host?

I don't follow that logic?  All that means is that the code works in php (which I know it does), not that phpactor is happy with it and finding nothing to complain about? 🤷‍♂️

dantleech:
> that the code works in php
no, it's testing that _phpactor_ evaluates the code statically, Phpactor is the interpreter here :slightly_smiling_face:  so if it determines that it's a string, then it correctly parsed and understood the methods. But it isn't to say there isn't a bug. Introducing a namespace has broken things historically, so you might try that.

downhiller:
Of course, I understand.  I knew it was an internal function, but didn't quite process what was going on.

I don't know if this is useful, but if I run `:lua vim.diagnostic.open_float()` I get a minute amount more information:

    Diagnostics:
    1. Function "setWarning" not found [worse.unresolved_name]
   
Not sure if that clarifies where the message is coming from or not!

downhiller:
Still having this issue, any further thoughts Dan?  Thank you!

downhiller:
OK, finally had time to look into this more.  The issue is that I have two versions of phpactor installed.  The one I've installed manually (dev-master) is NOT what neovim is loading in through Mason, which is stuck on 2024.11.28.1

I've tried to manually point Mason at the newer version, but can't seem to get it working and can't find (google/chatgpt) any concrete info on how to achieve this.  I don't suppose you can either guide me, or release a new Tag?

Thank you Dan, as always ❤️

Issue #2802, opened 2024-12-18
--is-class outputs traits & --is-trait option missing


phpwutz:
Hi everyone,

this is kind of a combo issue. I was just trying to get all classes (and *only* classes) in a given namespace using phpactor:

```
phpactor index:search --fqn-begins mynamespace\\model --is-class
```

and the output looks something like this:

```
class # mynamespace\model\Role (class)
class # mynamespace\model\_traits\DefaultChangedateSyncable (trait)
```

Oh wait but that second one is a trait, not a class!

Seems to me it would be better if `--is-class` really only outputs classes. Additionally, a `--is-trait` option would support getting those pesky traits. Maybe even better if a "types" parameter is supported so we can specify multiple types (defaulting to "all") when querying the index

dantleech:
it's working as expected, it would be more properly labeled as `--class-like`. Each of those results can be an interface, trait, enum or class.

dantleech:
If you want to make a pull request:

- Rename `--is-class` to `--is-class-like`
- Add `--is-class`, `-is-interface`, `--is-trait` and `--is-enum`

See the command:

https://github.com/phpactor/phpactor/blob/72dc7bb63b468ed78d30081b14bea795436e9b6e/lib/Indexer/Extension/Command/IndexSearchCommand.php#L76-L78

It's just a case of adding `Criteria::isClassConcrete`, `Criteria::isClassEnum`, etc.

phpwutz:
i prepared a PR and shamelessly reused your comment in the commit message

https://github.com/phpactor/phpactor/pull/2826

PR #2826, opened 2025-01-17
extend options for index:search

phpwutz:
- Rename --is-class to --is-class-like
- Add --is-class, -is-interface, --is-trait and --is-enum


phpwutz:
this should fix #2802

PR #2835, opened 2025-02-03
Psalm config

GDXbsv:


GDXbsv:
Do not know how to fix it actually

```
Type for "language_server_psalm.config" expected to be "string", got "NULL"  
```

lib/Extension/LanguageServerPsalm/LanguageServerPsalmExtension.php
```diff
@@ -85,13 +89,15 @@ public function configure(Resolver $schema): void
         ]);
         $schema->setTypes([
             self::PARAM_PSALM_BIN => 'string',
+            self::PARAM_PSALM_CONFIG => 'string',
```
dantleech, line 92: maybe remove this, I don't think we support union types - i.e. can't do `null|string`

dantleech, line 92: I think I'd remove this and keep the default as `null` to be consistent with other configurations where `null` => not declared.

dantleech:
it's fine. thanks!

PR #2836, opened 2025-02-05
Add doc for zed code editor

sethstha:
This PR adds documentation for Zed code editor. It supports phpactor by default as it's PHP LSP

### DIFF
--- a/doc/lsp/zed.rst
+++ b/doc/lsp/zed.rst
@@ -3,4 +3,4 @@
 Zed
 ===
 
-Phpactor is default LSP for Zed. Install Zed's `PHP extension <https://github.com/zed-industries/zed/tree/main/extensions/php/>`_.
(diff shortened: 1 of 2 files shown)


doc/lsp/zed.rst
```diff
@@ -0,0 +1,6 @@
+.. _lsp_client_zed:
+
+Zed
+================
```
dantleech, line 4: just 3 `===`

doc/lsp/zed.rst
```diff
@@ -0,0 +1,6 @@
...
+
+phpactor is default LSP for zed. Install zed's `Php extension <https://github.com/zed-industries/zed/tree/main/extensions/php/>`_.
```
dantleech, line 6: `Phpactor is the default LSP for Zed. Install Zed's` then PHP should be all caps: `PHP extension`

dantleech:
thanks. I additionally edited it to say `the default code editor`

PR #2840, opened 2025-02-08
Adding a way to prioritize code actions

### LINKED ISSUE #2820: Ordering the phpactor quickfix suggestions by relevance
When I invoke the quick fix action on an unused import then this is the order the quick fix actions are suggested.

image

While the "Remove unused import" option is there it's not the default option. I think we should prioritize fix options that depend on the current token over generic ones that always pop up.

mamazu:
The values that I set for the code actions are kind of random, maybe we could come up with a better system for that. But it does what it's supposed to do.

image

Now overriding methods is the last option

Closes #2820


dantleech:
ta

PR #2846, opened 2025-02-19
Symfony extension

Bleksak:
This is a WIP PR that adds Symfony completions, code actions, hover and diagnostics. Opening this as a draft because I'm not sure if anyone will be interested in using this.

Currently, it supports: FormType completions (including options) and template path completions

Also I would like to ask two questions:
1. Is there a way to hijack the Indexer, or add my own? I would like to index twig files
2. It seems that phpactor cannot parse Twig files as of now, would it be alright adding a twig parser to this project? (Maybe it won't be necessary, not sure yet)


dantleech:
Hi, thanks for this PR.

> Is there a way to hijack the Indexer, or add my own? I would like to index twig files... (and add a parser)

not really. although you could use the indexer Records to index symbols the current indexer is not aimed at parsing anything that is not PHP and I don't have the capcity to refactor it or review large changes.

---

in general I'd be happy to add the Symfony features if they are accompanied by tests and are provided in bite-size units that I can easily review, test and merge. However it should be noted that I do not have a great deal of capacity.

PR #2860, opened 2025-03-21
fix: Windows Phar fixes

zobo:


zobo:
Here are some Windows Phar related fixes.
There are a bunch of other _uri to path back to uri_ cases in the code, and I think they should be fixed...

lib/WorseReferenceFinder/WorseReflectionDefinitionLocator.php
```diff
@@ -122,7 +122,7 @@ private function gotoFunction(NodeContext $nodeContext): TypeLocations
             throw new CouldNotLocateDefinition($e->getMessage(), 0, $e);
         }
 
-        $path = $function->sourceCode()->uri()?->path();
```
zobo, line 125: On line 136 `$path` is transformed back to URI `TextDocumentUri::fromString($path)`. Why not just pass `$function->sourceCode()->uri()` directly to `new Location(` ?

lib/TextDocument/TextDocumentUri.php
```diff
@@ -82,7 +85,10 @@ public static function fromString(?string $uri): self
             ));
         }
 
-        $path = Path::makeAbsolute(substr($path, 1), '/');
+        if (str_starts_with($path, '/')) {
+            $path = substr($path, 1);
+        }
```
dantleech, line 90: would `ltrim($path, '/')` be better?

lib/WorseReferenceFinder/WorseReflectionDefinitionLocator.php
dantleech: it can be NULL, but :
```diff
diff --git a/lib/WorseReferenceFinder/WorseReflectionDefinitionLocator.php b/lib/WorseReferenceFinder/WorseReflectionDefinitionLocator.php
index 9869cc00..8fc6e24d 100644
--- a/lib/WorseReferenceFinder/WorseReflectionDefinitionLocator.php
+++ b/lib/WorseReferenceFinder/WorseReflectionDefinitionLocator.php
@@ -122,9 +122,9 @@ class WorseReflectionDefinitionLocator implements DefinitionLocator
             throw new CouldNotLocateDefinition($e->getMessage(), 0, $e);
         }

-        $path = $function->sourceCode()->uri()?->path();
+        $uri = $function->sourceCode()->uri();

-        if (null === $path) {
+        if (null === $uri) {
             throw new CouldNotLocateDefinition(sprintf(
                 'The source code for function "%s" has no path associated with it.',
                 $function->name()
@@ -132,10 +132,7 @@ class WorseReflectionDefinitionLocator implements DefinitionLocator
         }

         return new TypeLocations([
-            new TypeLocation(TypeFactory::unknown(), new Location(
-                TextDocumentUri::fromString($path),
-                $function->position()
-            ))
+            new TypeLocation(TypeFactory::unknown(), new Location($uri, $function->position()))
         ]);
     }
```

is better indeed

lib/TextDocument/TextDocumentUri.php
zobo: I think not, because there can be more than one `/` in front... We do have tests tho, so I'll try :)

The problem is that PHP is using URIs in a non-standard way. The correct form is:
`schema:// (SERVER) (/path)`. When there is no server, this looks like `schema://(/path)`. Or for relative path `schema://(path)`.
In phar case we have `phar://(/path)` on Unix but `phar://c:/path` on Windows...

What a mess...

lib/WorseReferenceFinder/WorseReflectionDefinitionLocator.php
dantleech: although the previous version will strip the schema etc which may or may not be intended ... here's hoping the tests know more than I do :laughing:

lib/TextDocument/Tests/Unit/TextDocumentUriTest.php
```diff
@@ -23,6 +23,14 @@ public function testFromPhar(): void
         $this->assertEquals('phar:///home/daniel/www/phpactor/phpactor/vendor/phpstan/phpstan/phpstan.phar/resources/functionMap.php', (string) $uri);
     }
 
+    public function testFromPharWindown(): void
```
mamazu, line 26: Typo in the function name?

dantleech:
Good to merge?

zobo:
Will check on this today.

zobo:
I found a few more I can change, but honestly, let's start with this minimum and do it in some other PR... Good to merge.

PR #2862, opened 2025-03-28
Tighening types

mamazu:
Small type improvements and regenerating the baseline.


lib/ReferenceFinder/Search/NameSearchResult.php
```diff
@@ -17,14 +17,13 @@ private function __construct(
     /**
      * @param string|FullyQualifiedName $name
      * @param string|NameSearchResultType $type
-     * @param string|TextDocumentUri $uri
      */
-    public static function create($type, $name, $uri = null): self
+    public static function create($type, $name, ?TextDocumentUri $uri = null): self
```
mamazu, line 21: It's never used with a string so let's remove the option.

PR #2878, opened 2025-04-24
Adding infix `instanceof` expressions for conditionals

### LINKED ISSUE #2877: Keyword completer should complete `instanceof` in conditions
I think this only makes sense if the parent is an if or a while statement. Any other cases I have missed?

mamazu:
This should help with writing if and while statements based on class instances. But technically this could pop up every time you have an expression. However, I can also imagine that this might be quite noisy.

Closes #2877


dantleech:
yes, I'd rather not introduce more noise..

mamazu:
So according to the tests it should work but I don't get any new completions with neovim. Maybe it's my setup that's broken?

lib/Completion/Bridge/TolerantParser/CompletionContext.php
```diff
@@ -367,6 +369,21 @@ public static function promotedPropertyVisibility(Node $node): bool
...
     }
 
+    public static function conditionInfix(Node $node): bool
+    {
+        // If the current node is not a variable we're at the beginning of the condition like "if (<>"
+        if ($node->getText() === '') {
+            return false;
+        }
+
+        $parent = $node->getParent();
```
przepompownia, line 379: It does not check (or test) the case with existing prefix (`ins` for example). Then we need to query the grandpa of `$node` :grin:

lib/Completion/Bridge/TolerantParser/CompletionContext.php
```diff
@@ -367,6 +370,22 @@ public static function promotedPropertyVisibility(Node $node): bool
...
+    public static function conditionInfix(Node $node): bool
+    {
+        // If the current node is not a variable we're at the beginning of the condition like "if (<>"
+        if ($node->getText() === '') {
+            return false;
+        }
+
+        $parent = $node->parent;
+        if (!$parent instanceof ExpressionStatement) {
+            return false;
```
przepompownia, line 382: Why not like previously:
```suggestion
        if ($parent instanceof IfStatementNode || $parent instanceof WhileStatement) {
            return true;
```

mamazu:
The test is trimming away whitespace at the end.

lib/Completion/Bridge/TolerantParser/CompletionContext.php
mamazu: I mean technically it could occur in any expression. But this was an error in the tests. I've fixed it now.

przepompownia:
...but now we have no test datasets for empty prefix.

mamazu:
True, good catch. Added.

mamazu:
Well it would be nice but I not necessary I think.

PR #2882, opened 2025-05-03
Per document cache

dantleech:
- Use an actively invalidated per-document cache for AST parsing and type resolving.  This will improve performance significantly in certain projects.
- Allow the `highlight` functionality to be toggled


przepompownia:
I have locally merged this branch into my experimental branch rather to catch if something can fail than to observe the impact on performance. 

@dantleech do you consider to add any benchmark here?

Issue #2901, opened 2025-06-16
PHPStan baseline not considered


igorbga:
I have installed VS Code PHPActor extension for an already existing project. I've also installed php

I also have PHP Rector recently installed and as it brings PHPStan as a dependency PHPActor's extension has detected it an created a corresponding phpstan.dist.neon file.

I have configured it like this:

```neon
parameters:
    level: max
    paths:
        - bin/
        - config/
        - public/
        - src/
        - tests/
    excludePaths:
        - public/js
includes:
	- phpstan-baseline.neon
```

And of course I've run phpstan with "--generate-baseline" to generate the file "phpstan-baseline.neon".

When I executed "vendor/bin/phpstan" manually everything works as expected, only 6 errors appear. Those are errors tha the baselining process already told me that would not be included in the baseline and that should be corrected manually. So far so good.

But if I open any file with VSCode all the  errors appear. It seems as if the baseline was not being read at all. 

I'm running VSCode with a devcontainer for the project, so just in case it was justa a matter of restarting PHPActor I closed VS and deleted the docker container. But after rebooting it the problem still persists.

dantleech:
Phpactor analyzes files with a temporary name, so I wouldn't expect the baseline tk catch it. This however is a feature for me, the baseline is there to prevent new errors being introduced, Phpactor is showing you the errors that exist.

stefanfisk:
Do not that PHPStan now have [Editor Mode](https://phpstan.org/user-guide/editor-mode) to allow for analysis on temp files!

stefanfisk:
Unless I am misunderstanding this issue, the root cause of the baseline not being respected is that phpactor runs phpstan on temp files so the paths are incorrect. With editor mode you can run phpstan on temp files while making phpstan treat them as if they were the original. There's probably a bit more to it, but my idea was to pass the `--tmp-file` and `--instead-of` CLI options [here](https://github.com/phpactor/phpactor/blob/e5a8e6997478a9d4b8efcb86949822f24a165a52/lib/Extension/LanguageServerPhpstan/Model/PhpstanProcess.php#L30)

stefanfisk:
And adjust the logic [here](https://github.com/phpactor/phpactor/blob/e5a8e6997478a9d4b8efcb86949822f24a165a52/lib/Extension/LanguageServerPhpstan/Model/Linter/PhpstanLinter.php#L38) so that both the original and temp paths are passed.

[...]

PR #2903, opened 2025-06-18
Update phpstan.rst to warn about baseline not being applied

igorbga:
This improvement tries to clarify situations like the one reported in issue #2901


dantleech:
thankyou!

Issue #2910, opened 2025-07-08
LSP, freeze/performance/IO


drzraf:
phpactor language server always froze emacs, even on small PHP codebase, and this has long been a no-go for me.
Happens that it's still happening nowadays and I wonder...

A quick look at the syscalls while phpactors eats 100% of the CPU and emacs freezes display some patterns.
One of them is this:
``` 
     0.000094 access("/home/user/ae/project/src/Service/Myclass.php", F_OK) = 0
     0.000083 access("/home/user/ae/project/src/Service/Myclass.php", F_OK) = 0
     0.000074 access("/home/user/ae/project/src/Service/Myclass.php", R_OK) = 0
     0.000072 openat(AT_FDCWD, "/home/user/ae/project/src/Service/Myclass.php", O_RDONLY) = 6
     0.000278 read(6, "<?php\n\nnamespace App\\Service;\n\nu"..., 9503) = 1311
     0.000061 read(6, "", 8192)         = 0
     0.000041 read(6, "", 8192)         = 0
     0.005249 access("/home/user/.cache/phpactor/index/project-20aaaf/class_f/e/fe1638107b5391ee09ee72ef531230da.cache", F_OK) = -1 ENOENT (No such file or directory)
     0.000247 access("/home/user/.cache/phpactor/index/project-20aaaf/class_f/e/fe1638107b5391ee09ee72ef531230da.cache", F_OK) = -1 ENOENT (No such file or directory)
     0.000225 access("/home/user/.cache/phpactor/index/project-20aaaf/function_f/e/fe1638107b5391ee09ee72ef531230da.cache", F_OK) = -1 ENOENT (No such file or directory)
     0.000065 access("/home/user/.cache/phpactor/index/project-20aaaf/function_f/e/fe1638107b5391ee09ee72ef531230da.cache", F_OK) = -1 ENOENT (No such file or directory)
     0.000154 access("/home/user/.cache/phpactor/index/project-20aaaf/constant_f/e/fe1638107b5391ee09ee72ef531230da.cache", F_OK) = -1 ENOENT (No such file or directory)
     0.000057 access("/home/user/.cache/phpactor/index/project-20aaaf/constant_f/e/fe1638107b5391ee09ee72ef531230da.cache", F_OK) = -1 ENOENT (No such file or directory)
     0.000124 access("/home/user/.cache/phpactor/worse-reflection/be832483423a45d40e7a0ee767c65d5d.map", F_OK) = 0
[...]

PR #2911, opened 2025-07-10
gh-2910: Memoize calls to the load the stubs

dantleech:


drzraf:
Do you mind activating artifacts for the `Compile phar` job so that we could test phpactor "nightly" builds ?

PR #2924, opened 2025-09-06
Ask permission to load `.phpactor.json` in project root

dantleech:


lib/Rename/Adapter/ReferenceFinder/AbstractReferenceRenamer.php
```diff
@@ -52,7 +53,11 @@ protected function doRename(TextDocument $textDocument, ByteOffset $offset, Byte
                 continue;
             }
 
-            yield $this->renameEdit($reference->location(), $range, $originalName, $newName);
+            try {
+                yield $this->renameEdit($reference->location(), $range, $originalName, $newName);
+            } catch (TextDocumentNotFound) {
+                continue;
```
dantleech, line 59: not related, but equally I'm going to keep it.

lib/Phpactor.php
```diff
@@ -94,19 +97,42 @@ public static function boot(InputInterface $input, OutputInterface $output, stri
...
+        if (!is_string($projectRoot)) {
+            throw new RuntimeException(sprintf(
+                'Unexpected type for project root, expected string got: %s',
+                get_debug_type($projectRoot)
+            ));
+        }
+
         $commandName = $input->getFirstArgument();
 
+        $trustPath = (new Xdg())->getHomeConfigDir() . '/phpactor/trust.json';
```
Xerkus, line 109: I believe this should live in XDG DATA not CONFIG directory, ie in `~/.local/share`

Xerkus:
Worked great.

May be for convenience have newly created config files trusted by default? As long as there was none before.

dantleech:
> May be for convenience have newly created config files trusted by default? As long as there was none before.

I think there's still a risk there that you `git pull` on a compromised repo and ...

Xerkus:
There was an interaction where that makes sense - where phpactor detects psalm and other tooling in composer and asks if they should be enabled.

May be that flow should be updated to ask if location should be trusted and if it is recorded as not trusted no request to enable tools should be made afterwards.

dantleech:
> May be that flow should be updated to ask if location should be trusted and if it is recorded as not trusted no request to enable tools should be made afterwards.

Yeah ... I've changed it so that it modifies the file regardless but the confirmation text indicates that the config file is not trusted and won't be used.

PR #2927, opened 2025-09-16
Prioritize static patterns over dynamic ones

zonuexe:
As added in the test case, it is inconvenient that you cannot exclude static files directly under the project root.


mamazu:
Yes, that's a very good point, I think the other way around this already works (including stuff in excluded directories). But yes static paths should be prioritized.

PR #2936, opened 2025-10-04
Using editormode of phpstan

mamazu:
What I've done:
* Using phpstan editor mode
* Upgrading phpstan to at least version 2.12

## Considerations
Phpactor is configured to use the project's phpstan if this does not meet the minimum requirements for the editor feature it might break. (I don't know if phpstan is forward compatible). Should we make this configurable or should we check the version of the installed phpstan to see if the option is available.

### DIFF
--- a/lib/Extension/LanguageServerPhpstan/Model/PhpstanProcess.php
+++ b/lib/Extension/LanguageServerPhpstan/Model/PhpstanProcess.php
@@ -51,7 +51,7 @@ class PhpstanProcess
             '--no-progress',
             '--error-format=json',
             '--tmp-file='.$tempFile,
-            '--instead-of="'.$filename.'"',
+            '--instead-of='.$filename,
             $filename
         ];
 


dantleech:
In the long term checking to see which version of a tool in is a prerequiste to any version-dependent feature. For now I'd be happy with an use_editor_mode option

dantleech:
I think this PR is missing the updated PHPStan process class? would it be possible to test it?

lib/Extension/LanguageServerPhpstan/LanguageServerPhpstanExtension.php
```diff
@@ -43,9 +45,16 @@ function (Container $container) {
         $container->register(
             Linter::class,
             function (Container $container) {
+                if ($container->parameter(self::PARAM_EDITOR_MODE)->value()
```
dantleech, line 48: this can be `->bool()` instead of `->value()` ?

dantleech:
>  I also added some quotes around the file path 

If I understand correctly quotes are not required as the process is not being invoked in a shell

mamazu:
Fair enough. Removed the quotes, maybe Symfony has a better error handling.

dantleech:
btw @mamazu (not sure how else to contact you) if you like stickers send me an email at my git address and I'll send you some a couple phpactor ones :slightly_smiling_face:

PR #2955, opened 2025-11-08
Make PHPStan severity configurable

rafalglowacz:
Hi! I'd like to use PHPStan integration, but Phpactor returns its diagnostics as errors, which makes some files look like they are full of high-priority problems. It'd be nice to be able to fix all PHPStan issues, of course, but it's not always an option, so I had an idea to make it configurable in Phpactor. Let me know what you think 🙂

### DIFF
--- a/doc/lsp/vim.rst
+++ b/doc/lsp/vim.rst
@@ -12,38 +12,34 @@ Client Guides
     .. tab:: Neovim LSP
 
         Prerequisites:
-            - Neovim 0.5.0 or higher.
-            - `nvim-lspconfig <https://github.com/neovim/nvim-lspconfig>`_ package (this contains common LSP configurations)
+            - Neovim 0.11.0 or higher.
             - The ``phpactor`` binary is :ref:`installed<installation>` and executable in your path
 
-        For example: include it in your ``.vimrc`` with Plug:
+        For example: include it in your ``init.lua``:
 
         ::
 
-            Plug 'neovim/nvim-lspconfig'
+            vim.lsp.enable('phpactor')
 
-        Then enable it in your ``.vimrc``. (note that `on_attach` is a
-        `callback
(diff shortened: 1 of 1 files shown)


lib/Extension/LanguageServerPhpstan/Model/DiagnosticsParser.php
```diff
@@ -14,7 +14,7 @@ class DiagnosticsParser
     /**
      * @return array<Diagnostic>
      */
-    public function parse(string $jsonString): array
+    public function parse(string $jsonString, int $severity): array
```
dantleech: needs the docblock type: `@param DiagnosticSeverity::* $severity`

dantleech:
I think I'm ok with this, not sure if @ondrejmirtes has any plans to introduce severity levels to the phpstan "errors" - but I guess then this could be refactored to only override the error level if explcitly configured.

just one comment

lib/Extension/LanguageServerPhpstan/Model/DiagnosticsParser.php
rafalglowacz: Strange, I was able to reproduce the PHPStan warning caused by the lack of this docblock, then I added it, the analysis started to pass locally, but in the pipelines it still fails. I'll have another look later/tomorrow.

rafalglowacz: Ok, I fixed the issues yesterday, all pipelines are green now.

rafalglowacz:
I'll create an additional PR, thanks for pointing that out.

przepompownia:
At least I needed to run `make configreference` to see the new option quicker than reviewing the whole change :wink:

dantleech:
yes, no objection to that being done as part of a PR of course, but it will happen anyway so I don't mind :slightly_smiling_face:

PR #2956, opened 2025-11-10
Fix type inference for container->get() with string service IDs

DaumantasUrb:
- Use TypeFactory::class() instead of fromString() for proper instance types
- Return null instead of object|null when service is not found
- Allow custom type resolvers to fully override inferred types

This improves type inference accuracy for Symfony DI container get() method by ensuring services return proper reflected class types instead of generic class-string types.

### DIFF
--- a/lib/Extension/Symfony/Adapter/Symfony/XmlSymfonyContainerInspector.php
+++ b/lib/Extension/Symfony/Adapter/Symfony/XmlSymfonyContainerInspector.php
@@ -127,7 +127,7 @@ class XmlSymfonyContainerInspector implements SymfonyContainerInspector
         if (empty($id) || empty($class)) {
             return null;
         }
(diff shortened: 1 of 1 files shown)


lib/Extension/Symfony/Adapter/Symfony/XmlSymfonyContainerInspector.php
```diff
@@ -127,9 +127,10 @@ private function serviceFromEl(DOMNode $serviceEl): ?SymfonyContainerService
         if (empty($id) || empty($class)) {
             return null;
         }
+        // Use TypeFactory::class() to create an instance type
```
dantleech: redundant comment ...

DaumantasUrb:
ill make a cleaner pr

PR #2958, opened 2025-11-11
cs-fixer: allow to choose the correct way of ignoring PHP version

### LINKED ISSUE #2957: php-cs-fixer: prepare to `PHP_CS_FIXER_IGNORE_ENV` deprecation
```
Setting PHP_CS_FIXER_IGNORE_ENV environment variable is deprecated and will be removed in 4.0, use unsupportedPhpVersionAllowed config instead.
```

For `3.89.2` we [need to use](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer?tab=readme-ov-file#supported-php-versions) `--allow-unsupported-php-version=yes` instead. 

I'm not sure if we need to check at the extension side which cs-fixer version is used to give the correct way. 

Edit: initially I checked it on dev-master, but `3.89.2` is a stable version that needs some change.

przepompownia:
Fix #2957

### DIFF
--- a/lib/VersionResolver/Tests/AggregateSemVerResolverTest.php
+++ b/lib/VersionResolver/Tests/AggregateSemVerResolverTest.php
@@ -2,12 +2,9 @@
 
 namespace Phpactor\VersionResolver\Tests;
 
-use Amp\Success;
-use Generator;
 use PHPUnit\Framework\TestCase;
 use Phpactor\VersionResolver\AggregateSemVerResolver;
-use Phpactor\VersionResolver\SemVersion;
-use Phpactor\VersionResolver\SemVersionResolver;
+use Phpactor\VersionResolver\ArbitrarySemVerResolver;
 use Prophecy\PhpUnit\ProphecyTrait;
 
 use function Amp\Promise\wait;
@@ -21,7 +18,10 @@ class AggregateSemVerResolverTest extends TestCase
      */
     public function testResolve(?string $expected, ?string ...$componentVersions): void
     {
-        $resolver = new AggregateSemVerResolver(...[...$this->mockResolvers(...$componentVersions)]);
+        $resolver = new AggregateSemVerResolver(...array_map(
+            fn (?string $version) => new ArbitrarySemVerResolver($version),
+            $componentVersions,
+        ));
 
         $actual = wait($resolver->resolve());
 
@@ -43,20 +43,5 @@ class AggregateSemVerResolverTest extends TestCase
         yield 'not null first' => ['1', '1', null];
         yield 'null first' => ['1', null, '1'];
         yield 'null only' => [null, null, null];
-    }
-
-    /**
-     * @return Generator<SemVersionResolver>
-     */
-    private function mockResolvers(?string ...$componentVersions): Generator
-    {
-        foreach ($componentVersions as $version) {
-            $resolver = $this->prophesize(SemVersionResolver::class);
-            $resolver
-                ->resolve()
-                ->willReturn(new Success((null === $version) ? null : new SemVersion($version)));
-
-            yield $resolver->reveal();
-        }
     }
 }


lib/Extension/LanguageServerPhpCsFixer/Model/PhpCsFixerProcess.php
```diff
@@ -109,12 +119,28 @@ public function run(string ...$args): Promise
...
+                            $data,
+                        ),
                     );
                 });
 
             return $process;
         });
     }
+
+    public function ignorePhpVersion(): static
```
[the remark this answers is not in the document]
przepompownia: Currently `run()` get variadic number of args and passing non-csfixer-arg argument would require change its signature and rather complicate it, whereas we need only modify the command for `fix()`. Leaving `run()` untouched seems easier. 

One thing I noticed here is that the approach somewhere between is to call ignore only from inside `fix()` - we then avoid unneeded calls from `load()` (e.g. for `describe` subcommand).

lib/Extension/LanguageServerPhpCsFixer/VersionResolver/PhpCsFixerVersionResolver.php
```diff
@@ -0,0 +1,34 @@
...
+{
+    public function __construct(
+        private string $binPath,
+        private LoggerInterface $logger,
+    ) {
+    }
+
+    public function resolve(): ?SemVersion
+    {
+        $versionQuery = wait((new PhpCsFixerProcess($this->binPath, $this->logger))->run('--version'));
```
przepompownia: In free time I need to investigate how such internal loops break `PhpCsFixerDiagnosticsProviderTest` and how we can fix it or at least work it around.

lib/Extension/LanguageServerPhpCsFixer/LanguageServerPhpCsFixerExtension.php
```diff
@@ -11,23 +11,42 @@
...
     public function load(ContainerBuilder $container): void
     {
+        $container->register(self::SERVICE_VERSION_RESOLVER, function (Container $container) {
+            $pathResolver = $container->expect(FilePathResolverExtension::SERVICE_FILE_PATH_RESOLVER, PathResolver::class);
+
+            $path = $pathResolver->resolve($container->parameter(self::PARAM_PHP_CS_FIXER_BIN)->string());
+
+            return new CachedSemVerResolver(
+                new CompoundSemVerResolver(
+                    new ArbitrarySemVerResolver($container->parameter(self::PARAM_PHP_CS_FIXER_VERSION)->value()),
```
dantleech: can use `->stringOrNull` here instead of `value`

lib/VersionResolver/CompoundSemVerResolver.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+namespace Phpactor\VersionResolver;
+
+use Amp\Promise;
+use function Amp\call;
+
+class CompoundSemVerResolver implements SemVersionResolver
```
dantleech: `Aggregate` is a term used most frequently for this type of thing in Phactor I think.

lib/Extension/LanguageServerPhpCsFixer/LanguageServerPhpCsFixerExtension.php
przepompownia: Maybe we could add a validator for `array<string,string>` too somewhere separately. For now, I just ignored the environment dictionary like in `PhpCodeSnifferExtension`.

lib/Extension/LanguageServerPhpCsFixer/Model/PhpCsFixerProcess.php
```diff
@@ -35,12 +43,18 @@ public function __construct(
     public function fix(string $content, array $options = []): Promise
     {
         return call(function () use ($content, $options) {
+            $this->ignorePhpVersion(yield $this->versionResolver?->resolve());
```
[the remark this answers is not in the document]
przepompownia: To keep the state untouched and still resolve the version here we need to change the signature of `run` because currently it uses that state (`$this->env`). I'm not sure if this is what you meant, but take a look.

lib/Extension/LanguageServerPhpCsFixer/Model/PhpCsFixerProcess.php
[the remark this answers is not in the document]
dantleech: i meant as an alteranative to having a method call to ignore the PHP version.

lib/Extension/LanguageServerPhpCsFixer/Tests/VersionResolver/PhpCsFixerVersionResolverTest.php
```diff
@@ -0,0 +1,30 @@
...
+        $path = realpath(__DIR__ . '/../../../../../vendor/bin/php-cs-fixer');
+
+        self::assertNotFalse($path);
+
+        $resolver = new PhpCsFixerVersionResolver($path, new NullLogger());
+
+        $process = call(function () use ($resolver) {
+            $version = yield $resolver->resolve();
+
+            self::assertSame('3.89.1', $version->__toString());
```
dantleech: probably just asserting that it's a valid semantic version would be less of a headache, maybe just match `{^[0-9]+.[0-9]+.[0-9+].*$` ?

lib/VersionResolver/Tests/AggregateSemVerResolverTest.php
```diff
@@ -0,0 +1,62 @@
...
+class AggregateSemVerResolverTest extends TestCase
+{
+    use ProphecyTrait;
+
+    /**
+     * @dataProvider provideResolverData
+     */
+    public function testResolve(?string $expected, ?string ...$componentVersions): void
+    {
+        $resolver = new AggregateSemVerResolver(...[...$this->mockResolvers(...$componentVersions)]);
```
dantleech: i'd avoid mocking here (and almost everywhere):

```
new AgrgegateSemVerResolver(
   new ArbitrarySemVerResolver(null),
   new ArbitrarySemVerResolver('1.2.3'),
);
```

doc/reference/configuration.rst
```diff
@@ -2157,17 +2170,30 @@ Path to the php-cs-fixer executable
...
 ``language_server_php_cs_fixer.env``
 """"""""""""""""""""""""""""""""""""
 
 
-Environment for PHP CS Fixer (e.g. to set PHP_CS_FIXER_IGNORE_ENV)
+Environment for PHP CS Fixer
 
 
-**Default**: ``{"XDEBUG_MODE":"off","PHP_CS_FIXER_IGNORE_ENV":true}``
+**Default**: ``{"XDEBUG_MODE":"off"}``
```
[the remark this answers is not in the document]
przepompownia: Let me leave this to your choice.

PR #2973, opened 2025-12-02
Adding support for PHP 8.5's pipe operator

dantleech:
Thanks to Phan and @rlerdorf the parser already supports the pipe operator.

TODO:

- [x] Make Phpactor understand that the arguments are callables.


lib/WorseReflection/Core/Inference/Resolver/BinaryExpressionResolver.php
```diff
@@ -157,6 +158,7 @@ private function walkBinaryExpression(
             TokenKind::SlashToken => TypeUtil::toNumber($left)->divide(TypeUtil::toNumber($right)),
             TokenKind::PercentToken => TypeUtil::toNumber($left)->modulo(TypeUtil::toNumber($right)),
             TokenKind::AsteriskAsteriskToken => TypeUtil::toNumber($left)->exp(TypeUtil::toNumber($right)),
+            TokenKind::PipeToken => $right,
```
dantleech: we need to "invoke" the callable to determine it's type.

PR #2980, opened 2025-12-06
POC: Ast Merging and Scoped Cache Invalidation

dantleech:
Experimenting with merging ASTs to preserve the object IDs of the nodes which might help with far more efficient caching techniques.

The idea would be that the revised AST (e.g. afer user has made a text edit) is merged with the previous one. This means that unchanged nodes would not be replaced and so we are then able to cache by the nodes by their object ID over multiple requests - this is something simulates the behavior of an incremental parser (and the parser is absolutely not the bottleneck currently).

The other part of this would be scoped cache invalidation, so - given that the user changes some text within a scoped area (e.g. a method block) we would only invalidate the nodes within that block and not those in the rest of the document (as we mostly do currently). This would have a dramatic impact on performance in large files.

---

- The merging parser seems to work (or at least I haven't succeeded in breaking it in isiolatino)
- But the LS is becoming corrupted possibly similar to the ssues with #2882

### DIFF
--- a/lib/TolerantAstDiff/Tests/Unit/AstDiffTest.php
+++ b/lib/TolerantAstDiff/Tests/Unit/AstDiffTest.php
@@ -13,23 +13,34 @@ use Phpactor\WorseReflection\Core\Util\NodeUtil;
 final class AstDiffTest extends TestCase
 {
     #[DataProvider('provideDiffTree')]
-    public function testDiffTree(string $source1, string $source2, ?\Closure $assertion = null): void
+    public function testDiffTree(array $sources, ?\Closure $assertion = null): void
     {
         $parser = new Parser();
-        $ast1 = $parser->parseSourceFile($source1);
-        $ast2 = $parser->parseSourceFile($source2);
-
         $diff = (new AstDiff());
-        $diff->merge($ast1, $ast2);
+        $oldAst = null;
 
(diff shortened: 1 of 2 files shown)


lib/TolerantAstDiff/AstDiff.php
```diff
@@ -0,0 +1,219 @@
...
+        }
+
+        $lastPosition = $node1->getFullStartPosition();
+        foreach ($node1->getChildNames() as $childName) {
+            $member1 = $node1->$childName;
+            $member2 = $node2->$childName;
+
+            if (is_array($member1)) {
+                //foreach ($member1 as $member) {
+                //    $lastPosition = $member->getFullStartPosition();
```
dantleech: figure out how why if this is necessary.

lib/TolerantAstDiff/AstDiff.php
```diff
@@ -0,0 +1,283 @@
...
+                            $node1Child->getFullStartPosition(),
+                            $node1Child->getFullWidth(),
+                            $node2Child->getFullText(),
+                        ));
+                        continue;
+                    }
+
+                    if ($node2Child instanceof Node) {
+                        // recurse on the listed node:
+                        $this->doMerge($node1Child, $node2Child);
```
dantleech: but node1...

PR #2996, opened 2025-12-21
Fix PHP 8.5 issues

przepompownia:
Problem: some phpunit tests fail on PHP 8.5

Solution:
- upgrade Psalm version
- increase test Psalm process timeout to 15 s
- fix new deprecations
- add 8.5 to CI matrix


dantleech:
I'm OK with bumping to 8.2

> probably I need your help here

I'm not sure what the issue is - that's a test file? but the tests are passing. It seems the only issue is PHPStan?

dantleech:
thanks

przepompownia:
Thank you too. I believe it's better to have tests for 8.5 in CI with a single deprecation (mentioned above) than to have no that tests.

PR #3000, opened 2025-12-24
Enforce passing textdocument

dantleech:
Refactor to always pass a TextDocument to the AST provider.

Note that this highlights the fact that we parse the document multiple times during completion due to the "need" to truncate it.

### DIFF
--- a/lib/Completion/Bridge/TolerantParser/ChainTolerantCompletor.php
+++ b/lib/Completion/Bridge/TolerantParser/ChainTolerantCompletor.php
@@ -2,6 +2,7 @@
 
 namespace Phpactor\Completion\Bridge\TolerantParser;
 
+use Phpactor\TextDocument\TextDocumentBuilder;
 use Phpactor\WorseReflection\Bridge\TolerantParser\AstProvider\TolerantAstProvider;
 use Generator;
 use Microsoft\PhpParser\Node;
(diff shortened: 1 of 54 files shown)


lib/Completion/Bridge/TolerantParser/ChainTolerantCompletor.php
```diff
@@ -31,15 +32,18 @@ public function complete(TextDocument $source, ByteOffset $byteOffset): Generato
         $truncatedSource = $this->truncateSource((string) $source, $byteOffset->toInt());
         $truncatedLength = strlen($truncatedSource);
 
+        // TODO :this is potentially very inefficient
         $node = $this->parser
-            ->get(substr_replace((string) $source, ' ', $truncatedLength, 0))
+            ->get(TextDocumentBuilder::create(substr_replace((string) $source, ' ', $truncatedLength, 0))->build())
```
dantleech: how can this be avoided?

lib/Completion/Bridge/TolerantParser/WorseReflection/DoctrineAnnotationCompletor.php
```diff
@@ -34,7 +35,9 @@ public function __construct(
     public function complete(TextDocument $source, ByteOffset $byteOffset): Generator
     {
         $truncatedSource = $this->truncateSource((string) $source, $byteOffset->toInt());
-        $sourceNodeFile = $this->parser->get((string) $source);
+
+        // TODO: this is potentially very inefficient
+        $sourceNodeFile = $this->parser->get(TextDocumentBuilder::create($source)->build());
```
dantleech: how else can this be acheved? perhaps at the parser level?

PR #3003, opened 2025-12-26
Make properties `readonly`

mamazu:
Replaces #2988

## Why?

* Preparation for `readonly` classes
* More clear on the intend (services should only contain reaonly properties)
* Makes finding unused properties easier (phpstan points them out)
* Replaces getters in DTO object (another PR)

# What?
Running readonly properties rule: https://getrector.com/rule-detail/read-only-property-rector


dantleech:
I'm not really a fan of this;

- it adds lots of noise
- it creates an unrealistic precedent (I won't be adding `readonly` to my promoted properties)
- it doesn't really clarify "intent" it just indicates if they are currently written to or not.

While I _would_ use `readonoly class` in certain situations as a design choice, and some classes would fall into the category retrospectively, I don't think it's worth adding readonly to everythng (and even then I'd use `readonly` class and probably not declare things at the property level).

PR #3004, opened 2025-12-27
Use container type expectations

przepompownia:


przepompownia:
It's one of fixes from #2996, but may be applied separately.

PR #3011, opened 2025-12-30
Incremental parser

dantleech:
This PR attempts to introduce a level of incremental parsing for the Tolerant PHP Parser.

Currently:

- The user types `$`, `a`, `b` and `c`.
- IDE sends 4 `completion` requests to Phpactor
- Phpactor will re-prase the document for each request

With this change two strategies are employed to avoid reparsing:

## In Token Changes

- The user types `$`, `a`, `b` and `c`
- Phactor sees that the `c` edit is within a `VariableName` token and that
  when the edit is applied to the previous token we go from `$ab` to `$abc`
  and `$abc` is _also_ a variable name token so we can just update it.

## Compound Node Edits

- User changes a statement within a compound node (i.e. within `{` `}` braces)
- Phactor sees this is a change to a statement
- Extracts the text from where the edited statement starts
- Transposes the text edit to the extracted text
- Applies the text edit
- Grafts the newly parsed AST into the old tree
- Reindexes the tokens.

## Affect

In certain situations this can result in a 2x or more speed increase, especially in large, complex, files because:

- Parsing time is 10x faster if an incremental update can be found.
- The unchanged node instances remain the same, meaning the cache can kick-in for type infrormation.

Things become more difficult if more context is required (e.g. `public function foo` is meaningless without the surrounding class construct) but in those cases they could, I suppose, be artificially supported.

---

TODO:

- [x] Document cache no longer cleared on update - document cache is used for diagnostics, so need to use separate caches.
- [x] WorkspaceListner
- [x] Diagnostics are no longer cleared on update (they were cleared via. the document cache clearing)
- [ ] Document cache not purged on goto definition (... ? e.g. after generaitng a missing method and jumping to it).
- [x] Diagnostics are not updated n realtime (event is not captured - text is not available)


lib/Extension/LanguageServer/Listener/IncrementalUpdateListener.php
```diff
@@ -39,13 +40,6 @@ public function getListenersForEvent(object $event): Generator
             };
             return;
         }
-
-        if ($event instanceof TextDocumentIncrementallyUpdated) {
-            yield function (TextDocumentIncrementallyUpdated $updated): void {
-                $this->applyEdits($updated->identifier(), $updated->events());
-            };
-            return;
-        }
```
dantleech: this is currently handled by the AstUpdateListener -- and the question is if the AST listener should pro-actively update the AST (there's a possiblity it does so needlessly) or if we could buffer the edits for the AST - which would risk flooding it when it is requested - (in such a case we could have a buffer limit that when exceeded would trigger a single re-parse)...

another option may be to just make the incremental parser opt-in until it's sufficiently good enough that we can pro-actively parse.

lib/WorseReflection/Bridge/TolerantParser/AstProvider/IncrementalAstUpdater.php
```diff
@@ -0,0 +1,144 @@
...
+        ), false);
+    }
+
+    private function applyToken(Node $node, TextEdit $edit): bool
+    {
+        $token = self::tokenAtRange($node, $edit->range());
+
+        // the text edit is NOT contained in a token
+        if (null === $token) {
+            return false;
```
dantleech: we should return the _reason_ in these cases for debugging/logging.

lib/WorseReflection/Bridge/TolerantParser/AstProvider/IncrementalAstUpdater.php
```diff
@@ -0,0 +1,144 @@
...
+
+        // if the new token STARTS later than `<?php\n` then it has some
+        // leading-whitespace/doc-comment and it will corrupt things.
+        if ($newToken->start > 6) {
+            return false;
+        }
+
+        // TODO: if the edit caused the token kind to change then
+        //       we can probably update the kind?
+        if ($newToken->kind !== $token->kind) {
```
dantleech: can we?

lib/Extension/LanguageServer/LanguageServerExtension.php
```diff
@@ -655,7 +652,11 @@ private function taggedServices(Container $container, string $tag, ?string $fqn
      */
     private function resolveListeners(Container $container): array
     {
-        return array_filter(array_keys($container->getServiceIdsForTag(self::TAG_LISTENER_PROVIDER)), function (string $service) use ($container) {
+        $serviceIds = $container->getServiceIdsForTag(self::TAG_LISTENER_PROVIDER);
+        uasort($serviceIds, function (array $a, array $b) {
+            return ($a['priority'] ?? 0) <=> ($b['priority'] ?? 0);
```
dantleech: i have a suspicion that these are already sorted somewhere

lib/WorseReflection/Bridge/TolerantParser/AstProvider/IncrementalAstUpdater.php
```diff
@@ -0,0 +1,258 @@
...
+        // a single token corresponding to the original then we can update the
+        // original. If we have more than one token then we cannot.
+        if (count($newTokens) !== 1) {
+            return sprintf('more than one token found (%d, orig: %s, new: %s))', count($newTokens), $originalTokenText, $editedTokenText);
+        }
+
+        $newToken = $newTokens[0];
+
+        // WHY?
+        if (strlen($editedTokenText) !== $newToken->length) {
```
dantleech: why indeed

lib/WorseReflection/Bridge/TolerantParser/AstProvider/IncrementalAstUpdater.php
```diff
@@ -0,0 +1,258 @@
...
+use Microsoft\PhpParser\TokenStreamProviderInterface;
+use Phpactor\TextDocument\ByteOffsetRange;
+use Phpactor\TextDocument\TextDocumentBuilder;
+use Phpactor\TextDocument\TextDocumentUri;
+use Phpactor\TextDocument\TextEdit;
+use Phpactor\TextDocument\TextEdits;
+use Phpactor\WorseReflection\Core\AstProvider;
+use Throwable;
+
+class IncrementalAstUpdater
```
dantleech: it would be beneficial to add a benchmark

dantleech:
```
[WR     ][INFO][135748.796360] PARS 0.0964 incremental update with "token" strategy
[WR     ][INFO][135753.272761] PARS 0.6905 incremental update with "token" strategy
[WR     ][INFO][135760.249353] PARS 0.0948 incremental update with "token" strategy
```

It's not clear how the token strategy can take half a second in some cases (35k lines of code)

przepompownia:
Eventually I found time to merge this PR into my working branch to at least catch possible crashes.

PR #3031, opened 2026-03-10
fix(BinaryExpressionResolver): null coalesce on undefined variable

przepompownia:
Fix: #3030

In practice: 

- it fixes hover info on `$baz` from the test case
- diagnostic provider still complains about undefined variable (apparently it's a separate issue).

### DIFF
--- a/lib/WorseReflection/Core/Inference/Resolver/BinaryExpressionResolver.php
+++ b/lib/WorseReflection/Core/Inference/Resolver/BinaryExpressionResolver.php
@@ -284,6 +284,10 @@ class BinaryExpressionResolver implements Resolver
 
     private function nullCoalesce(Type $left, Type $right): Type
     {
+        if ($left instanceof MissingType) {
+            return $right;
+        }
+
         if ($left->isNullable()) {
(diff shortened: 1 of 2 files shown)


przepompownia:
Image

Image

while on master:
Image

dantleech:
in the screen shot it complains that `$undef` is undefined - is that not correct?

mamazu:
This makes sense, php (when prod mode is on) just prefills undefined variables with null and the expression will evaluate to an empty array. I don't think that getting rid of the diagnostic is a good idea though.

przepompownia:
In this case, the code is correct. The complaint is more about potentially bad practice – perhaps it should be shown as a warning (to fix somewhere else).

https://github.com/php/php-src/blob/c658d3c7876ed0269c6b9aef78ff37afa2496145/tests/lang/operators/coalesce.phpt#L27

dantleech:
happy for the diagnostic to be changed to a warning

przepompownia:
@dantleech I guess you test benchmark here and we probably did the rebase at almost the same time :)

PR #3037, opened 2026-04-12
Index optimiser

dantleech:
This PR introduces an index optimiser:

- Remove references records that were defined in now-non-existing files.
- Remove references to non-existing files.

> [!NOTE]
> This is mostly because I'm fed up of seeing entries in the index for non-existing files.
> In theory they should be removed from the index by the file watcher. But the file-watcher service is the LSP client now by default, and Neovim (for example) [doesn't watch files](https://github.com/neovim/neovim/pull/22405) on Linux at least.

In anycase, having a way to periodically clean up is a missing feature and I've observed this operation saving 5-10% on index size on some long-standing projects.

TODO:

- [x] Provide an `index:optimise` command.
- [x] Provide an LSP service that runs periodically.
- [ ] Consider skipping when there's no activity.
- [ ] Add test for the IndexerHandler or split out the Optmiser to a new class.

### DIFF
--- a/lib/Extension/LanguageServerIndexer/Handler/IndexerHandler.php
+++ b/lib/Extension/LanguageServerIndexer/Handler/IndexerHandler.php
@@ -28,7 +28,8 @@ use function Amp\delay;
 
 class IndexerHandler implements Handler, ServiceProvider
 {
-    const SERVICE_INDEXER = 'indexer';
+    public const SERVICE_INDEXER = 'indexer';
+    public const SERVICE_OPTIMISER = 'optimiser';
 
     public function __construct(
         private Indexer $indexer,
@@ -38,6 +39,7 @@ class IndexerHandler implements Handler, ServiceProvider
         private EventDispatcherInterface $eventDispatcher,
         private ProgressNotifier $progressNotifier,
         private ?int $reindexTimeout = null,
+        private int $optimiserTimeout = 3600,
     ) {
     }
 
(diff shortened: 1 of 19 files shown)


lib/Indexer/IndexAgentBuilder.php
```diff
@@ -218,7 +219,12 @@ private function buildIndex(): Index
             $this->logger
         );
 
-        return new SerializedIndex($repository);
+        return new SerializedIndex(
+            $repository,
+            // TODO: inject this as it should probably use the chain locator that includes
+            // the workspace locator
+            new FilesystemTextDocumentLocator()
```
dantleech: TODO

lib/Extension/LanguageServerIndexer/Handler/IndexerHandler.php
```diff
@@ -48,6 +50,7 @@ public function methods(): array
     {
         return [
             'phpactor/indexer/reindex' => 'reindex',
+            'phpactor/indexer/optimise' => 'optimise',
```
przepompownia, line 53: Shouldn't it be `optimizer`? If so, I see that the command starts periodic optimization. In this case we cannot run it once immediately, which could be useful sometimes (especially when the options provided by the LSP client are different from those in phpactor.json).

dantleech: Optimise is the British and [therefore correct](https://aloneonahill.com/blog/if-php-were-british/) spelling.

But no the optimiser is time intensive and it would conflict with the indexer on startup.

przepompownia: I meant a reference to the name of the new method - not spelling correctness :)

dantleech: fixed

PR #3038, opened 2026-04-16
Cleanup after removing PHP 8.1 support

przepompownia:


przepompownia:
The changelog needs to update too. If I'm not mistaken, some parts of CI require updating on the Github side.

dantleech:
Updated the merge rules, thanks!

PR #3048, opened 2026-05-22
Gh 3042: No stacking code actions / code-action concurrency

dantleech:
This PR:

- Will provide an option to resolve code-actions in a separate process and that will be enabled by default.
- Ensures only one code action is running at a time (subsequent code actions will kill the previous code-action process).

TODO:

 - [x] Evaluate performance: **1.6x slower** - as no cache - and caching would require syncing text documents between processes.
 - [ ] Add test for the outsourced code-action provider.
 - [x] Allow the behavior to be disabled (so that code-actions run in-process again)
- [x] Concurrency doesn't work (needs changed upstream of overriding here)

### DIFF
--- a/lib/Extension/LanguageServer/CodeAction/TolerantCodeActionProvider.php
+++ b/lib/Extension/LanguageServer/CodeAction/TolerantCodeActionProvider.php
@@ -15,7 +15,7 @@ final class TolerantCodeActionProvider implements CodeActionProvider
 {
     public function __construct(
         private CodeActionProvider $provider,
-        private ClientApi $client
+        private ?ClientApi $client
     ) {
     }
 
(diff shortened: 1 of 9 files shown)


lib/Extension/LanguageServer/CodeAction/TolerantCodeActionProvider.php
```diff
@@ -25,12 +25,14 @@ public function provideActionsFor(TextDocumentItem $textDocument, Range $range,
...
                 return yield $this->provider->provideActionsFor($textDocument, $range, $cancel);
             } catch (Throwable $error) {
-                $this->client->window()->showMessage()->error(sprintf(
-                    'Provider %s (%s) failed: %s',
-                    $this->provider::class,
-                    $this->provider->describe(),
-                    $error->getMessage(),
-                ));
+                if (null !== $this->client) {
+                    $this->client->window()->showMessage()->error(sprintf(
```
dantleech, line 30: this will mean that code action errors fail silently ... 

either:

- log the error if the `ClientApi` isn't available (which it will not be by default)
- just throw? these providers should not fail.

dantleech:
Merging this for now. The better option for both this and code actions and perhaps other things (completion f.e.) would be to have worker processes that are kept in sync with each other and that can run arbitrary LSP operations.

Issue #3053, opened 2026-07-05
Language server crashes with "Call to undefined function posix_kill()" when ext-posix is missing (undeclared dependency)


fain182:
## Summary

The language server crashes with `PHP Fatal error: Uncaught Error: Call to undefined function posix_kill()` on systems where the `posix` extension is not installed. `ext-posix` is not declared in `composer.json` (only `ext-mbstring` and `ext-tokenizer` are), so nothing catches this at install time — and since the phar distribution bypasses composer's platform checks entirely, the missing extension only surfaces as a runtime fatal error in the middle of an LSP session.

This is a follow-up to #1989 and #2034, which were closed without a fix; the problem is still present in `2026.06.25.0`.

## Environment

- Fedora 44, PHP 8.5.7 (`php-cli` package; on Fedora the `posix` extension ships in the separate `php-process` package, which is not pulled in by `php-cli`)
- phpactor `2026.06.25.0` phar, launched by the Zed editor's PHP extension
- `php -m` shows `pcntl` present, `posix` absent

## What happens

The server starts fine and responds to `initialize`. On the first `textDocument/codeAction` request, `OutsourcedCodeActionProvider` spawns a child process and later calls `Process->kill()`, which reaches `amphp/process`'s `Runner::tryPosixKill()` → `posix_kill()` → fatal error, and the whole server dies:

```
PHP Fatal error:  Uncaught Error: Call to undefined function posix_kill() in phar://.../phpactor.phar/vendor/amphp/process/lib/Internal/Posix/Runner.php:223
Stack trace:
#0 .../vendor/amphp/process/lib/Internal/Posix/Runner.php(197): Amp\Process\Internal\Posix\Runner::tryPosixKill()
...
#4 .../vendor/amphp/process/lib/Process.php(141): Amp\Process\Internal\Posix\Runner->kill()
#5 .../lib/Extension/LanguageServer/CodeAction/OutsourcedCodeActionProvider.php(63): Amp\Process\Process->kill()
...
#23 .../vendor/phpactor/language-server/lib/Core/Server/LanguageServer.php(114): Amp\Loop::run()
```

## Why this is painful to diagnose

From the editor's point of view there is no error at all: since editors request code actions automatically right after a file is opened, the server dies within about a second of starting, and every subsequent feature (go to definition, find references, hover) just fails with "server shut down". The fatal error is only visible if you capture the server's stderr. It took wrapping the phar in a logging shim to find the actual cause.

## Suggested fix

[...]

PR #3054, opened 2026-07-05
Require posix extension and fail early when required extensions are missing

fain182:
Related to #3053

The language server crashed mid-session with `Call to undefined function posix_kill()` on systems without the posix extension (e.g. Fedora, where it ships separately in the `php-process` package). `ext-posix` was not declared in `composer.json`, and PHAR distributions bypass Composer's platform checks entirely, so the missing extension only surfaced as a fatal error when amphp/process tried to kill a subprocess during an LSP session.

This PR implements both fixes suggested in the issue:

- Declare `ext-posix` in `composer.json` `require` (and update `composer.lock`), so Composer-based installations fail early at install time.
- Check required extensions (`mbstring`, `posix`, `tokenizer`) at startup in `bin/phpactor`, exiting with a descriptive error message instead of crashing mid-session — this covers PHAR distributions, with a hint pointing Fedora users to the `php-process` package.

Example output when an extension is missing:

```
Phpactor requires the following PHP extension(s) which are not loaded: posix. Install or enable them and try again (e.g. on Fedora the posix extension is provided by the php-process package).
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+## Unreleased
+
+- Require the posix extension and fail early with a descriptive error when a
+  required extension is missing (instead of crashing mid-session) @fain182 #3053
+
 ## 2026.06.25.0
 
 - Mago diagnostics and lint extension @marjovanlier #3052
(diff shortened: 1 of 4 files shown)


dantleech (approved):
thanks

PR #3056, opened 2026-07-21
Code action to add #[Override]

AJenbo:
Fixes https://github.com/phpactor/phpactor/issues/2966

(Note I still don't know how to quit Vim so I haven't tried it all out manually, hope the tests makes up for it).

It's a bit late here so let me know if you want me to change anything that I missed.

PR #3057, opened 2026-07-21
Adding override attribute to override code action

mamazu:
When generating method overrides with phpactor it should also include the Override attribute for those methods.

refs #3056


templates/code/Method.php.twig
```diff
@@ -6,4 +6,7 @@
 {% endfor %}
  */
 {% endif %}
+{% if prototype.override %}
+#[\Override]
+{% endif %}
```
mamazu, line 11: I don't know what the best way is to scope this only to modern php projects.

lib/CodeTransform/Adapter/WorseReflection/Refactor/WorseOverrideMethod.php
```diff
@@ -58,6 +58,7 @@ private function getMethodPrototype(ReflectionClass $class, ReflectionMethod $me
         );
 
         $methodBuilder = $builder->class($method->declaringClass()->name()->short())->method($method->name());
+        $methodBuilder->override();
```
dantleech: you would need to probably take a similar approach as: https://github.com/phpactor/phpactor/pull/3056/changes#diff-64fe3016a6b976b1ad21f7e0babe3331ef32089c520339a013a07c9eaa1d28cbR74

i.e. in

lib/CodeTransform/Tests/Adapter/WorseReflection/Refactor/WorseOverrideMethodTest.php
```diff
@@ -61,7 +64,7 @@ private function overrideMethod(string $source, string $className, string $metho
     {
         $reflector = $this->reflectorForWorkspace($source);
         $factory = new WorseBuilderFactory($reflector);
-        $override = new WorseOverrideMethod($reflector, $factory, $this->updater());
+        $override = new WorseOverrideMethod($reflector, $factory, $this->updater(), '8.5');
```
mamazu, line 67: Do we also want to test both versions like run the override Method Test for php 8.0 (as an example for a version that should not have the override feature) and for 8.3 again with the Override or is just testing the "newer" version enough?

lib/CodeBuilder/Domain/Builder/MethodBuilder.php
```diff
@@ -28,6 +28,8 @@ class MethodBuilder extends AbstractBuilder implements NamedBuilder
 
     protected bool $abstract = false;
 
+    protected bool $override = false;
```
dantleech, line 31: as a code "builder" this would be better implemented as a list of "attribute" prototypes otherwise the precedent here means having a property for every type of attribute.

mamazu: What kind of datastructure would you suggest? Another bit field like with the modifiers? An Array of Attribute Builders (might be overblown)?

dantleech (approved):

dantleech:
thanks, i'll see if I can add the attribute builder.

PR #3061, opened 2026-08-11
Fix worse:analyse not finding functions

AJenbo:
It kept failing on most things I ran it on, this seems to get it to a working state.


dantleech:
great thanks, that command is primarily for debugging and is in no way supposed to compete with a static analyzer :laughing: but happy if it reports less false positives.

PR #3063, opened 2026-08-14
Fix a few false positives from the analyzer

AJenbo:
The issues where found by running https://zonuexe.github.io/php-typing-conformance/ on the latest master branch. It was actually already the LSP (including qodana) with the least false positives with just 4 failing cases. If you could make a release after merging this I would like to submit a PR to the project so that it can be automatically checked going forward.


dantleech (approved):

dantleech:
thanks tagged: https://github.com/phpactor/phpactor/releases/tag/2026.06.23.0

PR #3064, opened 2026-08-25
gh-2902 - Handling the same variable being declared twice in a stub

### LINKED ISSUE #2902: parse_str second argument showing as undefined variable
When using parse_string the second argument is a variable that is set to the result. This variable doesn't need to be defined beforehand in PHP but Phpactor is expecting it to be.

`parse_str('a=1&b=2', $result); // Shows undefined variable "$result"`

Phpactor handles a similar case correctly. The third argument of preg_match, $matches, is defined in a similar way and doesn't show an error.

`preg_match('/a/', 'a', $matches); // No error`

GregMayes:
Fixes https://github.com/phpactor/phpactor/issues/2902

When a function or method stub declares the same variable twice, it will now correctly stick to the index of the first declaration of the parameter. The last declared parameter will still "win" in terms of ending up in the return `ReflectionParameterCollection` due to the  index names being the same, which is the current functionality.

This only really affects a few functions, with `parse_str` being one of the best examples. Here is the PHPStorm stub showcasing `$result` being declared twice:
```php
function parse_str(
    string $string,
    #[PhpStormStubsElementAvailable(from: '5.3', to: '7.4')] &$result = [],
    #[PhpStormStubsElementAvailable(from: '8.0')] &$result
): void {}
 ```

Was also able to get rid of an entry in PHPStan baseline


dantleech:
thanks!

PR #3065, opened 2026-09-14
Parallel indexing

AJenbo:
I was a bit sad to see that Phpactor is the slowest to index a project, while not being the most CPU hungry, this addresses that by splitting indexing up between sub processes and merging the results in to a single index. Most of the dependencies for this was already there (amphp/process, symfony/process) so only fidry/cpu-core-counter was added as a new dependency.

This result in roughly 6x faster indexing cutting my test project from 3m30s to 35s. This does increase peak memory during indexing but not in an unexpected way.

The PR is a bit on the larger side so I could split it up in a refactor and then add the parallel indexer on top if preferred, but I would prefer not to...


lib/Indexer/Adapter/Parallel/Worker.php
```diff
@@ -0,0 +1,149 @@
...
+        return $this->process->isRunning();
+    }
+
+    /**
+     * @param list<string> $paths
+     */
+    public function assign(array $paths): void
+    {
+        $this->chunk = $paths;
+        $this->input->write(Protocol::encode(
```
dantleech, line 67: why use a  custom protocol instead of f.e. json + newline? how much more efficient is this?

lib/Indexer/Adapter/Php/FileSearchIndex.php
```diff
@@ -114,11 +114,22 @@ private function open(): void
...
-            return [$parts[0], $parts[1], $parts[2] ?? null, $parts[3] ?? null];
-        }, explode("\n", (string)file_get_contents($this->path))));
+            if (!isset($parts[1])) {
+                continue;
+            }
+
+            // key the subjects exactly as `write()` does, otherwise a record
+            // that is written again after the index has been read back gets a
+            // second entry rather than replacing the first
+            $subjects[$parts[0] . $parts[1]] = [$parts[0], $parts[1], $parts[2] ?? null, $parts[3] ?? null];
```
dantleech, line 129: 👍

lib/Indexer/Extension/Command/IndexBuildCommand.php
dantleech, line 138: not necessarily for this PR but would be nice to show some information about the paraellelism in the command output before or after indexing (e.g. how many workers were utilised)

lib/Phpactor.php
```diff
@@ -145,6 +145,8 @@ public static function boot(InputInterface $input, OutputInterface $output, stri
 
         if ($phpactorBin) {
             $config[LanguageServerExtension::PARAM_PHPACTOR_BIN] = $phpactorBin;
+            // Workers do not shre the CWD with the main process so we must resolve relative paths.
+            $config[IndexerExtension::PARAM_WORKER_BIN] = realpath($phpactorBin) ?: $phpactorBin;
```
dantleech, line 149: not sure exactly why this is necessary? Phactor is already spawning workers for other tasks and the CWD is passed as `--working-dir`?

https://github.com/phpactor/phpactor/blob/d2f84a35ef44796b66ceb59540863e1d953eae4c/lib/Extension/LanguageServer/LanguageServerExtension.php#L676-L679

dantleech (commented):
very nice. a couple of comments / questions

lib/Indexer/Extension/Command/IndexBuildCommand.php
AJenbo: Done

lib/Indexer/Adapter/Parallel/Worker.php
AJenbo: It's not really for performance, `unserializing()` of objects is simpler and we don't have to worry about thing like UTF encoding and it matches the disk format so felt like less of a gamble.

lib/Phpactor.php
AJenbo: We need it so that workers use the same version and runtime as the manager process. `--working-dir` doesn't help here since it sets where the worker resolves the project, but not where the executable is resolved.

dantleech: I thought PHACTOR_BIN takes care of that problem, DMd you on discord.

dantleech: Will merge as it is and refactor to:

- Move the `PHPACTOR_BIN` setting to `CoreExtension`
- Make it absolute.
- Change this to use that.

dantleech:
thanks @AJenbo

lib/Phpactor.php
AJenbo: Cool, I'll give it a spin and see if I can make it error and then propose a general fix and steps to reproduce _if_ I have any issues after that.

PR #3067, opened 2026-09-17
Add LSP Call Hierarchy support

klobastov:
## Summary

Implements the LSP Call Hierarchy capability in phpactor's language server,
so clients (e.g. Qwen Code) can use `textDocument/prepareCallHierarchy`,
`textDocument/callHierarchy/incomingCalls` and
`textDocument/callHierarchy/outgoingCalls`.

## Changes

- New `LanguageServerCallHierarchy` extension registering a
  `CallHierarchyHandler` as an LSP method handler.
- `CallHierarchyHandler` registers the `callHierarchyProvider` capability and
  serves the three methods using `Reflector`, `ReferenceFinder`,
  `AstProvider` and `DefinitionLocator`.
- `OutgoingCallsWalker` collects method/function calls inside the enclosing
  function body.
- Unit tests for the handler.
- Extension registered in `lib/Phpactor.php`.

## Dependency

This PR depends on **phpactor/language-server-protocol#30** (adds
`CallHierarchyParams` and the three request interfaces). The
`phpactor/language-server-protocol` dependency must be bumped to the tagged
release containing those classes before this PR can pass `composer integrate`
(phpstan/phpunit).


lib/Extension/LanguageServerCallHierarchy/Tests/Unit/Handler/CallHierarchyHandlerTest.php
```diff
@@ -0,0 +1,292 @@
...
+            ->shouldBeCalled();
+
+        $tester = $this->createTester();
+        $response = $tester->requestAndWait(CallHierarchyPrepareRequest::METHOD, [
+            'textDocument' => ProtocolFactory::textDocumentIdentifier(self::EXAMPLE_URI),
+            'position' => ProtocolFactory::position(0, 0),
+        ]);
+
+        self::assertNotNull($response);
+        $this->assertNull($response->result);
```
dantleech: phpunit assertions should be called with `self::`

lib/Extension/LanguageServerCallHierarchy/Inference/OutgoingCallsWalker.php
```diff
@@ -0,0 +1,84 @@
...
+    private function collectCall(CallExpression $node): void
+    {
+        $callable = $node->callableExpression;
+        $isMethod = $callable instanceof MemberAccessExpression;
+        $name = NodeUtil::nameFromTokenOrNode($node, $callable);
+        if ($name === '' || !$this->inRange($node)) {
+            return;
+        }
+
+        $this->calls[] = [
```
dantleech: it would be better to use an objec rather than an array. e.g. `Call`

lib/Extension/LanguageServerCallHierarchy/Tests/Unit/Handler/CallHierarchyHandlerTest.php
```diff
@@ -0,0 +1,292 @@
...
+
+    private FrameResolver $frameResolver;
+
+    protected function setUp(): void
+    {
+        $this->reflector = $this->prophesize(Reflector::class);
+        $this->finder = $this->prophesize(ReferenceFinder::class);
+        $this->astProvider = $this->prophesize(AstProvider::class);
+        $this->locator = $this->prophesize(DefinitionLocator::class);
+        $this->frameResolver = new FrameResolver(
```
dantleech: I'd prefer ths was tested using the `LanguageServerTester` and that mocks are avoided. You can see an example here: https://github.com/phpactor/phpactor/blob/8dc44fd24a1e407ec0bf44a5f2cafb6f7ab5dc07/lib/Extension/LanguageServerReferenceFinder/Tests/Unit/Handler/GotoImplementationHandlerTest.php#L53-L57

lib/WorseReferenceFinder/MethodCallReferenceFinder.php
```diff
@@ -0,0 +1,115 @@
...
+
+/**
+ * Finds references to a method or function by searching all open documents
+ * in the workspace for matching call expressions.
+ *
+ * Unlike TolerantVariableReferenceFinder (which only finds variable
+ * references within a single document), this finder searches across all
+ * open documents to find method call sites.
+ */
+class MethodCallReferenceFinder implements ReferenceFinder
```
dantleech, line 29: there seems to be no test for this.

phpstan-baseline.neon
```diff
@@ -7962,6 +7962,24 @@ parameters:
 			count: 1
 			path: lib/Extension/LanguageServerBridge/Tests/Converter/TextEditConverterTest.php
 
+		-
+			message: '#^Parameter \#1 \$workspace of class Phpactor\\Extension\\LanguageServerCallHierarchy\\Handler\\CallHierarchyHandler constructor expects Phpactor\\LanguageServer\\Core\\Workspace\\Workspace, mixed given\.$#'
```
dantleech: entries should not be added to the baseline. These can be avoided by using `$container->expect(LanguageServerExternsion::SERVICE_WORKSPACE, Workspace::class)` for example.

lib/Extension/CompletionWorse/Tests/Unit/CompletionWorseExtensionTest.php
```diff
@@ -32,12 +31,12 @@ public function testBuild(): void
         $completor = $container
             ->expect(CompletionExtension::SERVICE_REGISTRY, TypedCompletorRegistry::class)
             ->completorForType('php');
-        assert($completor instanceof Completor);
```
klobastov, line 35: Image

lib/Extension/LanguageServerCallHierarchy/Handler/CallHierarchyHandler.php
```diff
@@ -0,0 +1,268 @@
...
+    public function incomingCalls(CallHierarchyIncomingCallsParams $params): Promise
+    {
+        $item = $params->item;
+        $lspDoc = $this->workspace->get($item->uri);
+        $doc = TextDocumentConverter::fromLspTextItem($lspDoc);
+        $offset = PositionConverter::positionToByteOffset($item->selectionRange->start, $lspDoc->text);
+
+        $promise = call(function () use ($doc, $offset) {
+            $incomingCalls = [];
+            foreach ($this->referenceFinder->findReferences($doc, $offset) as $potentialLocation) {
```
dantleech: this seems to only account for local references (due to the reference finders used in the extension) - it doesn't show incoming calls from the entire workspace. is this by design / is it how other implementations work?

lib/Extension/LanguageServerCallHierarchy/Handler/CallHierarchyHandler.php
```diff
@@ -0,0 +1,268 @@
...
+        );
+    }
+
+    private function buildCallerItem(Location $location): CallHierarchyItem
+    {
+        $uri = $location->uri()->__toString();
+        try {
+            $lspDoc = $this->workspace->get($uri);
+        } catch (UnknownDocument) {
+            $range = RangeConverter::toLspRange($location->range(), '');
```
dantleech: is there a reason to catch this exception? converting a range with an empty string will resul in Range(0,0,0,0) unconditionally.

lib/Extension/LanguageServerCallHierarchy/Handler/CallHierarchyHandler.php
```diff
@@ -0,0 +1,268 @@
...
+        $doc = TextDocumentConverter::fromLspTextItem($this->workspace->get($uri));
+
+        try {
+            $locations = $this->definitionLocator->locateDefinition($doc, ByteOffset::fromInt($offset));
+            $location = $locations->first()->location();
+            $targetUri = $location->uri()->__toString();
+            try {
+                $targetText = $this->workspace->get($targetUri)->text;
+            } catch (UnknownDocument) {
+                $range = RangeConverter::toLspRange(ByteOffsetRange::fromInts($offset, $offset), $text);
```
dantleech: similarly this will result in the wrong range which wouldn't be useful. Unless there is a reason I would let the exception bubble up (alternatively returning `null` and skipping)

lib/Extension/LanguageServerCallHierarchy/Handler/CallHierarchyHandler.php
```diff
@@ -0,0 +1,268 @@
...
+                return new CallHierarchyItem($name, $kind === 'method' ? SymbolKind::METHOD : SymbolKind::FUNCTION, $uri, $range, $range);
+            }
+            $range = RangeConverter::toLspRange($location->range(), $targetText);
+            return new CallHierarchyItem($name, $kind === 'method' ? SymbolKind::METHOD : SymbolKind::FUNCTION, $targetUri, $range, $range);
+        } catch (CouldNotLocateDefinition) {
+            $range = RangeConverter::toLspRange(ByteOffsetRange::fromInts($offset, $offset), $text);
+            return new CallHierarchyItem($name, $kind === 'method' ? SymbolKind::METHOD : SymbolKind::FUNCTION, $uri, $range, $range);
```
dantleech: ditto

lib/WorseReferenceFinder/MethodCallReferenceFinder.php
```diff
@@ -0,0 +1,115 @@
...
+use Phpactor\ReferenceFinder\PotentialLocation;
+use Phpactor\ReferenceFinder\ReferenceFinder;
+use Phpactor\TextDocument\ByteOffset;
+use Phpactor\TextDocument\Location;
+use Phpactor\TextDocument\TextDocument;
+use Phpactor\WorseReflection\Core\AstProvider;
+
+/**
+ * Finds references to a method or function by searching all open documents
+ * in the workspace for matching call expressions.
```
dantleech: is there a _specific_ reason to only search opened documents vs. the entire workspace? is that what other implementations do?

Phpactor reference finding can be very slow accross the entire workspace so it's not necessarily a bad thing to do this but not sure it's what would be expected.

lib/Extension/LanguageServerCallHierarchy/Handler/CallHierarchyHandler.php
```diff
@@ -0,0 +1,268 @@
...
+        $offset = PositionConverter::positionToByteOffset($item->selectionRange->start, $lspDoc->text);
+
+        return call(function () use ($item, $lspDoc, $doc, $offset) {
+            $ast = $this->astProvider->get($doc);
+            $bodyRange = $this->enclosingFunctionBodyRange($ast, $offset->toInt());
+            if (null === $bodyRange) {
+                return [];
+            }
+
+            $walker = new OutgoingCallsWalker($bodyRange);
```
dantleech: using the `Reflector` to walk the AST will perform static analysis which can be very slow. It seems this just needs to collect method calls without resolving the types they are being called on - in which case no analysis of the AST is required.

You can simply call:

```
foreach ($node->getDescendantNodes() as $descendantNode) {
    if (!$descendantNode instance of MemberAcessNode) {
        continue;
    }
    
    // ...
}
```

dantleech:
looks generally good, some questions and suggestions.

we have a separate issue that the previosuly tagged version of the LSP protocol causes issues - so Phpactor was pinned to the one before that. I'll look at upgrading the protocol in `master`

dantleech:
ok, this PR can now be rebased on develop (the protocol is already updated there).

